PR #8, opened 2019-09-10
Change FileName link field to url

bfolie:
Per the docs, the "link" field is now "url".


tests/serialization/test_process_spec.py
```diff
@@ -51,7 +51,7 @@ def valid_data():
             'description': 'a long description',
             'tags': []
         },
-        file_links=[{'type': 'file_link', 'filename': 'Cake spec', 'link': 'specs/cake.txt'}],
+        file_links=[{'type': 'file_link', 'filename': 'Cake spec', 'url': 'specs/cake.txt'}],
```
kroenlein: Do we care that this is not a URL?

bfolie: I believe that this will be an actual URL. I'm going to add some comments to the relevant Jira issue.

bfolie: Oh I see you are referring to this specific text, and not the concept of the url field in general.

It doesn't matter for the test, but would make things more clear. I'll change that in the PR on file upload.

PR #10, opened 2019-09-10
Deleted module type

latture:
This PR removes `module_type` from predictors, processors and capabilities. Corresponding changes to the backend will have to be made before merging this PR.


jtharris:
@latture are we losing information with this?  In the Orion API modules need have a type - I'm surprised this works.  Do you know what is setting the module type for the API?

latture:
@jtharris I don't know what is setting the type. It's also unclear why `module_type` was not required for other module types, e.g. scorers and workflows.

PR #15, opened 2019-09-11
Feature/file upload

bfolie:
Adds the ability to upload a file to S3 through the python client.

Upload is broken up into three pieces: a request to the database that returns temporary credentials, the upload itself using the S3 client, and a message to the database saying that the upload is complete.

Tests make heavy use of mock objects to represent the S3 client and the Citrine Session.

### DIFF
--- a/src/citrine/resources/file_link.py
+++ b/src/citrine/resources/file_link.py
@@ -17,11 +17,10 @@ class _Uploader(object):
 
     def __init__(self):
         self.bucket = ''
-        self.key = ''
+        self.object_key = ''
         self.upload_id = ''
         self.region_name = ''
         self.aws_access_key_id = ''
-        self.aws_secret_access_token = ''
         self.aws_secret_access_key = ''
         self.aws_session_token = ''
         self.object_key = ''
@@ -110,6 +109,7 @@ class FileCollection(Collection[FileLink]):
         extension = os.path.splitext(file_path)[1]
         mime_type = mimetypes.types_map[extension]
         file_size = os.stat(file_path).st_size
+        assert isinstance(file_size, int)
         upload_json = {
             'files': [
                 {
--- a/src/citrine/resources/file_link.py
+++ b/src/citrine/resources/file_link.py
@@ -94,7 +94,7 @@ class FileCollection(Collection[FileLink]):
             object_key = upload_response['uploads'][0]['s3_key']
             upload_id = upload_response['uploads'][0]['upload_id']
         except KeyError:
-            raise RuntimeError("Upload response is missing some fields: "
+            raise RuntimeError("Upload initiation response is missing some fields: "
                                "{}".format(upload_response))
 
         s3_client = boto3_client('s3',
@@ -110,10 +110,16 @@ class FileCollection(Collection[FileLink]):
                                    "exception: {}".format(file_path, e))
             s3_version = upload_s3_response['VersionId']
             path = self._get_path() + "/uploads/{}/complete".format(upload_id)
-            self.session.put_resource(path=path, json={'s3_version': s3_version})
+            complete_response = self.session.put_resource(path=path, json={'s3_version': s3_version})
 
+        try:
+            file_id = complete_response['file_info']['file_id']
+            version = complete_response['file_info']['version']
+        except KeyError:
+            raise RuntimeError("Upload completion response is missing some "
+                               "fields: {}".format(complete_response))
         file_link_dict = {
             'filename': dest_name,
-            'url': self._get_path(object_key)
+            'url': self._get_path(file_id) + '/versions/{}'.format(version)
         }
         return FileLink.build(file_link_dict)
--- a/src/citrine/resources/file_link.py
+++ b/src/citrine/resources/file_link.py
@@ -12,7 +12,7 @@ from citrine._rest.resource import Resource
 from citrine._session import Session
 
 
-class _Uploader(object):
+class _Uploader:
     """Holds the many parameters that are generated and used during file upload."""
 
     def __init__(self):
@@ -23,7 +23,6 @@ class _Uploader(object):
         self.aws_access_key_id = ''
         self.aws_secret_access_key = ''
         self.aws_session_token = ''
-        self.object_key = ''
         self.s3_version = ''
 
 
--- a/src/citrine/resources/file_link.py
+++ b/src/citrine/resources/file_link.py
@@ -12,6 +12,21 @@ from citrine._rest.resource import Resource
 from citrine._session import Session
 
 
+class _Uploader(object):
+    """Holds the many parameters that are generated and used during file upload."""
+
+    def __init__(self):
+        self.bucket = ''
+        self.key = ''
+        self.upload_id = ''
+        self.region_name = ''
+        self.aws_access_key_id = ''
+        self.aws_secret_access_token = ''
+        self.aws_secret_access_key = ''
+        self.aws_session_token = ''
+        self.object_key = ''
+
+
 class FileLink(Resource['FileLink'], TaurusFileLink):
     """Resource that stores the name and url of an external file."""
 
@@ -41,7 +56,7 @@ class FileCollection(Collection[FileLink]):
         """Build an instance of FileLink."""
         return FileLink.build(data)
 
-    def upload(self, file_path, dest_name=None) -> FileLink:
+    def upload(self, file_path: str, dest_name: str = None) -> FileLink:
         """
         Uploads a file to the dataset.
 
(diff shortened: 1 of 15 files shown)


Pipfile
```diff
@@ -14,4 +14,6 @@ sphinxcontrib-apidoc = "*"
 
 [packages]
 citrine = {editable = true,path = "."}
-taurus = {git = "https://github.com/CitrineInformatics/taurus.git", editable = true, ref="develop"}
\ No newline at end of file
+taurus = {git = "https://github.com/CitrineInformatics/taurus.git", editable = true, ref="develop"}
+boto3 = "*"
+botocore = "*"
```
dannycodes: newline?

src/citrine/resources/file_link.py
```diff
@@ -0,0 +1,111 @@
...
+    """Resource that stores the name and url of an external file."""
+
+    pass
+
+
+class FileCollection(Collection[FileLink]):
+    """Represents the collection of all file links associated with a dataset."""
+
+    _path_template = 'projects/{project_id}/datasets/{dataset_id}/files'
+    _dataset_agnostic_path_template = 'projects/{project_id}/files'
```
dannycodes: No such route exists in ProductApi. Do you expect it to or is this for the future?

src/citrine/resources/file_link.py
```diff
@@ -0,0 +1,111 @@
...
+    def upload(self, file_path, dest_name=None) -> FileLink:
+        """
+        Uploads a file to the dataset.
+
+        Parameters
+        ----------
+        file_path: str
+            The path to the file on the local computer.
+        dest_name: str, optional
+            The name the file will have after being uploaded. If unspecifiied, the local name of
```
dannycodes: unspecified

src/citrine/resources/file_link.py
```diff
@@ -0,0 +1,111 @@
...
+        try:
+            region_name = upload_data['s3_region']
+            aws_access_key_id = upload_data['temporary_credentials']['access_key_id']
+            aws_secret_access_key = upload_data['temporary_credentials']['secret_access_key']
+            aws_session_token = upload_data['temporary_credentials']['session_token']
+            bucket = upload_data['s3_bucket']
+            object_key = upload_data['uploads'][0]['s3_key']
+            upload_id = upload_data['uploads'][0]['upload_id']
+        except KeyError:
+            raise RuntimeError("Amazon S3 response is missing some fields: {}".format(upload_data))
```
dannycodes: This is not Amazon S3 response, it's the "upload" to product api response

src/citrine/resources/file_link.py
```diff
@@ -0,0 +1,111 @@
...
+            except ClientError as e:
+                raise RuntimeError("Upload of file {} failed with the following "
+                                   "exception: {}".format(file_path, e))
+            s3_version = upload_response['VersionId']
+            path = self._get_path() + "/uploads/{}/complete".format(upload_id)
+            self.session.put_resource(path=path, json={'s3_version': s3_version})
+
+        file_link_dict = {
+            'filename': dest_name,
+            'url': self._get_path(object_key)
```
dannycodes: what is this url set as?
is it `api/v1/projects/:id/datasets/:id/files/:id`?

bfolie: There's no `api/v1`. That part is appended by the Session object

bfolie: But otherwise, yes

src/citrine/resources/file_link.py
bfolie: Fixed

src/citrine/resources/file_link.py
bfolie: No. Because so many resources have two access routes, one with a dataset and one without, I hard-coded these two paths in. But this one is not needed here and can stay NotImplemented. I'll remove it.

bfolie:
Build is failing because test coverage has dropped below threshold.

Need to figure out how to write tests for this and the other parts of the client that touch an external service.

src/citrine/resources/file_link.py
```diff
@@ -0,0 +1,209 @@
...
+from botocore.exceptions import ClientError
+
+from taurus.entity.file_link import FileLink as TaurusFileLink
+from citrine._serialization.properties import String
+from citrine._rest.collection import Collection
+from citrine._rest.resource import Resource
+from citrine._session import Session
+
+
+class _Uploader(object):
```
jtharris: You don't need `object` here

src/citrine/resources/file_link.py
```diff
@@ -0,0 +1,209 @@
...
+    """Holds the many parameters that are generated and used during file upload."""
+
+    def __init__(self):
```
jtharris: If there is no behavior here, you could probably use a named tuple.

tests/resources/test_file_link.py
```diff
@@ -0,0 +1,152 @@
...
+    uploader.region_name = 'us-west'
+    uploader.aws_access_key_id = 'sahgkjsgahnei'
+    uploader.aws_secret_access_key = 'kdydahkd78452978'
+    uploader.aws_session_token = 'sdfhuaf74yf783g4ofg7g3o'
+    uploader.object_key = '234787521--abcde'
+    uploader.s3_version = '2'
+    return uploader
+
+
+class MockClient(object):
```
jtharris: We should be explicit about naming here.  https://martinfowler.com/articles/mocksArentStubs.html
I would suggest either `FakeS3Client` or `StubS3Client` - test doubles come in different flavors.

Also, moving to `utils` with the other test doubles might organize things a bit more.

tests/resources/test_file_link.py
```diff
@@ -0,0 +1,152 @@
...
+
+def test_string_representation(valid_data):
+    """Test the string representation."""
+    assert str(FileLink.build(valid_data)) == '<File link \'materials.txt\'>'
+
+
+@pytest.fixture
+def uploader() -> _Uploader:
+    """An _Uploader object with all of its fields filled in."""
+    uploader = _Uploader()
```
jtharris: This could be a factory boy factory.

tests/serialization/test_file_link.py
```diff
@@ -0,0 +1,23 @@
+"""Tests of FileLink serialization and deserialization."""
+import pytest
+
+from citrine.resources.file_link import FileLink
+from tests.utils.factories import FileLinkDataFactory
+
+
+@pytest.fixture
+def valid_data() -> dict:
+    return FileLinkDataFactory(url='www.citrine.io', filename='materials.txt')
```
jtharris: Since this is a one-liner, I would recommend just ditching the fixture altogether and having each test call the factory.  This helps the test case tell the story better.

src/citrine/resources/file_link.py
bfolie: I remembered learning once that it is good practice to always have objects extends `object`. Your comment made me look up why, and it has to do with reasons that have not been valid since Python 2.

src/citrine/resources/file_link.py
bfolie: That would be cleaner, but named tuples aren't mutable, and the fields are not filled in all at the same time.

tests/serialization/test_file_link.py
bfolie: Done

tests/resources/test_file_link.py
bfolie: Done

tests/resources/test_file_link.py
bfolie: Done. Thanks for this and the other suggestions.

tests/utils/factories.py
```diff
@@ -84,3 +91,19 @@ class Meta:
...
     summary = factory.Faker('catch_phrase')
     description = factory.Faker('bs')
+
+
+class _UploaderFactory(factory.Factory):
+    class Meta:
+        model = _Uploader
+
+    @factory.post_generation
+    def assign_values(obj, create, extracted):
```
jtharris: You can just set these values directly like in all of the other factories.  No need for a `post_generation` hook if you are assigning static values like this.

tests/utils/session.py
```diff
@@ -57,7 +53,21 @@ def post_resource(self, path: str, json: dict, *args, **kwargs) -> dict:
...
+    def put_resource(self, path: str, json: dict, *args, **kwargs) -> dict:
+        self.calls.append(FakeCall('PUT', path, json))
+        return self.response
+
     def delete_resource(self, path: str) -> dict:
         self.calls.append(FakeCall('DELETE', path))
         return self.response
 
+
+class FakeS3Client:
```
jtharris: Awesome!  Thanks 👍

tests/utils/factories.py
bfolie: I got an error when I tried to set the values directly, I think because they tried to go through the `__init__` method for `_Uploader`, which doesn't take any arguments. 🤷‍♂

bfolie:
> I still think that a cleaner division of responsibility between requesting a file upload (and returning a client) and actually performing the upload would be better testable and less coupled for maintainability, but that can be a later rev if need be.

I didn't make a method that produces a client because the client didn't seem to me to be central to the upload process. It instead seemed to be a result of using S3, and if we uploaded some other way I wouldn't assume that there would also be a client involved _and_ it would also have a `put_object` method. But I have never written code to upload or download files before, so my view of the subject is myopically limited to this framework.

PR #22, opened 2019-09-13
Document the get_material_history method

bfolie:
Write a proper docstring for the method and include it in the "Getting Started" docs.

### DIFF
--- a/docs/source/getting_started/basic_functionality.rst
+++ b/docs/source/getting_started/basic_functionality.rst
@@ -67,8 +67,18 @@ There are currently three ways to filter, and they only work on data model objec
 :func:`~citrine.resources.data_concepts.DataConceptsCollection.filter_by_tags`,
 :func:`~citrine.resources.data_concepts.DataConceptsCollection.filter_by_attribute_bounds`,
 and :func:`~citrine.resources.data_concepts.DataConceptsCollection.filter_by_name`.
+
 Filtering by tags or attribute bounds can be scoped to a dataset or to a project.
 Filtering by name must be scoped to a dataset.
+
+Get Material History
+^^^^^^^^^^^^^^^^^^^^
+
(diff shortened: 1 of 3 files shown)


docs/source/getting_started/basic_functionality.rst
```diff
@@ -67,9 +67,19 @@ There are currently three ways to filter, and they only work on data model objec
...
+
 Filtering by tags or attribute bounds can be scoped to a dataset or to a project.
 Filtering by name must be scoped to a dataset.
 
+Get Material History
+^^^^^^^^^^^^^^^^^^^^
+
+Starting with a specific root :class:`MaterialRun <citrine.resources.material_run.MaterialRun>`,
+you can retrieve the complete material history--every process, ingredient, material, and measurement
+that that went into the root material and that you have read access to. The method is
```
kroenlein: every process, ingredient and material that went into the root material .., as well as the measurements that were performed on all of those materials

?

src/citrine/resources/material_run.py
```diff
@@ -95,8 +95,30 @@ def get_type(cls) -> Type[MaterialRun]:
...
         return MaterialRun
 
-    def get_history(self, scope, id):
-        """Get the history associated with a material."""
+    def get_history(self, scope, id) -> Type[MaterialRun]:
+        """
+        Get the history associated with a material.
+
+        The history contains every single process, ingredient, measurement, and material that
+        went into creating this particular material run. The returned object is a material run
```
kroenlein: See above

docs/source/getting_started/basic_functionality.rst
bfolie: Good suggestion, thanks

src/citrine/resources/material_run.py
bfolie: Same

PR #30, opened 2019-09-14
Feature/mock test coverage

bfolie:
Bring up the test coverage, mostly by using Mock to mock a Session object. This brings us up to 89% and establishes a pattern that can be used to get coverage on most of the remaining statements.

### DIFF
--- a/tests/resources/test_data_concepts.py
+++ b/tests/resources/test_data_concepts.py
@@ -1,0 +1,32 @@
+import pytest
+import unittest
+import mock
+import requests
+from uuid import uuid4
+
+from citrine._session import Session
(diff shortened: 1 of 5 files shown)


tests/resources/test_data_concepts.py
```diff
@@ -0,0 +1,86 @@
...
+    'tags': [],
+    'notes': None,
+    'conditions': [],
+    'parameters': [],
+    'template': None,
+    'file_links': [],
+    'type': 'measurement_spec'
+}
+
+class SessionTests(unittest.TestCase):
```
maxhutch: Is there a reason we have to use unittest here, rather than sticking with pytest patterns?

bfolie:
Closing in favor of #33

PR #33, opened 2019-09-16
Adding more test coverage

jtharris:
I arbitrarily picked `MaterialRun` as a way to concretely cover one of the resource types.  This does introduce some extra testing machinery:

 * [FactoryBoy](https://factoryboy.readthedocs.io/en/latest/index.html) for generating test models and i/o dicts
 * A `FakeSession` for stubbing and asserting network calls.  I think this is a clearer way of understanding the interactions, rather than mocking, but that is just personal preference (we can discuss here if another technique is desired)

This gets the coverage over 90% - we could do some follow-up PRs to bump it up further.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ install:
 - pip install --no-deps -e .
 script:
 - pytest --flake8 --cov=src/ --cov-report term-missing --cov-report term:skip-covered
-  --cov-fail-under=83 -s -r .
+  --cov-fail-under=91 -s -r .
 - cd docs; make html; cd ..;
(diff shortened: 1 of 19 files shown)


maxhutch:
Another quick question about tooling: is extending `unittest` required for `FactoryBoy`?  Previously, we were using pytest and its fixtures.  I'd like to keep our patterns as homogenous as reasonably possible for future developer ease.

jtharris:
@maxhutch nothing about factory boy requires it, I can see how it would look porting over the common setup methods to use the fixtures

jtharris:
@maxhutch I updated the test style and got everything green with python 3.5

PR #34, opened 2019-09-16
Generalize Generic Response Class

mcgalcode:
The response class was specifically expecting two strings for
  instantiation. There is no guarantee that a generic HTTP response
  will have "status" and "message" keys in the body, so this
  relaxes the requirements for instantiating this class.

  Since it is used to represent any generic HTTP response we might
  get that may or may not include a response body, this changes
  the class to accept status_code and body (as a dictionary).


src/citrine/resources/response.py
```diff
@@ -1,12 +1,25 @@
...
+        self.body = body
+
+    def _get_status_string(self):
+        if self.status_code is not None:
+            return str(self.status_code)
+        else:
+            return "No HTTP status available"
+
+    def _get_body_string(self):
+        if self.body is None:
```
maxhutch:
```suggestion
        if self.body is not None:
```

tests/resources/test_response.py
```diff
@@ -0,0 +1,25 @@
+import pytest
+import re
+
+from citrine.resources.response import Response
+
+
+def test_response_repr():
```
maxhutch: Could you add a brief explanation of what's being tested?

tests/resources/test_response.py
```diff
@@ -0,0 +1,25 @@
...
+
+    resp = Response()
+    no_body_found = re.search("No body available", resp.__repr__())
+    no_status_code_found = re.search("No HTTP status available", resp.__repr__())
+    assert no_body_found
+    assert no_status_code_found
+
+    resp_with_code = Response(status_code=404)
```
maxhutch: Why not split this into a different test?

tests/resources/test_response.py
```diff
@@ -0,0 +1,25 @@
...
+    no_body_found = re.search("No body available", resp_with_code.__repr__())
+    status_code_found = re.search("404", resp_with_code.__repr__())
+    assert no_body_found
+    assert status_code_found
+
+    resp_with_code_and_body = Response(status_code=404, body={"message": "a quick message"})
```
maxhutch: Why not split this into a different test?

PR #36, opened 2019-09-17
Added an enumerated design space

latture:
This PR adds an enumerated design space defined by a list of python dictionaries. Users define a list of descriptors and a list of dictionaries that map from descriptor key to value. Values are allowed to be any type, i.e. `data: Mapping[str, Any]`. The backend casts the value to a string when building the design space.

### DIFF
--- a/tests/serialization/test_design_spaces.py
+++ b/tests/serialization/test_design_spaces.py
@@ -3,13 +3,14 @@ import uuid
 
 import pytest
 
-from citrine.informatics.design_spaces import DesignSpace, ProductDesignSpace
+from citrine.informatics.descriptors import CategoricalDescriptor, RealDescriptor
+from citrine.informatics.design_spaces import DesignSpace, ProductDesignSpace, EnumeratedDesignSpace
 from citrine.informatics.dimensions import ContinuousDimension, EnumeratedDimension
 
 
 @pytest.fixture
-def valid_data():
-    """Produce valid design space data."""
+def valid_product_data():
+    """Produce valid product design space data."""
     return dict(
         module_type='DESIGN_SPACE',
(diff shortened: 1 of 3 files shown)


latture:
Rebuild me!

latture:
Should we add an integration test that ensures the enumerated processor and design space can be used in a workflow?

tests/serialization/test_design_spaces.py
```diff
@@ -3,13 +3,14 @@
...
-from citrine.informatics.design_spaces import DesignSpace, ProductDesignSpace
+from citrine.informatics.descriptors import CategoricalDescriptor, RealDescriptor
+from citrine.informatics.design_spaces import DesignSpace, ProductDesignSpace, EnumeratedDesignSpace
 from citrine.informatics.dimensions import ContinuousDimension, EnumeratedDimension
 
 
 @pytest.fixture
-def valid_data():
-    """Produce valid design space data."""
+def valid_product_data():
```
andrewmillspaugh: Might be a good place to check out factory-boy. See https://github.com/CitrineInformatics/citrine-python/blob/develop/tests/utils/factories.py

latture:
Tests fail due to insufficient test coverage:
```
FAIL Required test coverage of 91% not reached. Total coverage: 90.57%
```

tests/serialization/test_design_spaces.py
latture: That's a good idea. If we don't end up auto-generating these classes we should refactor the orion tests. There's quite a bit of code duplication.

latture:
I added a few tests in #40 that bump up test coverage. The build is passing now.

PR #38, opened 2019-09-17
Add Global Sharing Endpoint

dannycodes:
Consumes the Endpoint created in https://github.com/CitrineInformatics/shire/pull/514

Allows users to share with the global project without knowing its ID (decouple logic)

NOTE: We may want to workshop the name of this method xD


src/citrine/resources/project.py
```diff
@@ -157,6 +157,13 @@ def share(self,
             "resource": {"type": resource_type, "id": resource_id}
         })
 
+    def global_share(self,
```
maxhutch: How about `make_public`

maxhutch: Or toggle public

mcgalcode: (to weigh in, I prefer make_public and make_private, toggle is too implicit for an access control feature IMO)

mcgalcode:
@maxhutch added make private here as well.

src/citrine/resources/project.py
```diff
@@ -157,6 +157,22 @@ def share(self,
             "resource": {"type": resource_type, "id": resource_id}
         })
 
+    def make_public(self,
+                    resource_type: str,
```
maxhutch: Are the `resource_type`s documented?  If we're expecting the user to provide the resource type, then they should know what to provide.

src/citrine/resources/project.py
```diff
@@ -157,6 +157,22 @@ def share(self,
...
+                    resource_id: str) -> Dict[str, str]:
```
maxhutch: Can we return something more instructive than `Dict[str, str]` for a user-facing method?  Or minimally: document what the user should expect if the call is successful vs not successful.

src/citrine/resources/project.py
```diff
@@ -157,6 +157,22 @@ def share(self,
...
+        """Make a resource publicly accessible."""
+        return self.session.post_resource(self._path() + "/make-public", {
+            "resource": {"type": resource_type, "id": resource_id}
+        })
+
+    def make_private(self,
```
maxhutch: Some comments as `make_public`

src/citrine/resources/project.py
bfolie: We should be making full docstrings (description, parameters, returns, and anything else as appropriate) for every method.

mcgalcode:
@maxhutch @bfolie I updated the interface to be much simpler and moved the logic for access control entity representations to the Resource class. I expect that this might be adjusted more in the future though.

Also added docstrings.

docs/source/getting_started/public_access.rst
```diff
@@ -0,0 +1,42 @@
...
+
+Toggling Public Access
+----------------------
+
+A dataset can be made public (or made private again) using the project from which it originates
+using the :func:`~citrine.resources.project.Project.make_public` and :func:`~citrine.resources.project.Project.make_private` methods on the project.
+
+Example
+-------
+
```
bfolie: Add a line here to the effect "Assume you have a "band gaps project" with known uid `band_gaps_project_id` and an associated dataset with known uid `strehlow_cook_dataset_id`".

It has been my experience that unless these things are spelled out, people have looked at the docs and thought they were built-ins.

mcgalcode: This is excellent advice.

PR #41, opened 2019-09-17
PLA-1385 Enable searching by external ID

asantas93:
Just adds `scope` param which defaults to Citrine scope. I think the `Collection` class and its usage could probably do with a bit of refactoring so we have more code reuse, but I got worried about *scope* creep when I started trying to do it.


bfolie:
> I think the `Collection` class and its usage could probably do with a bit of refactoring so we have more code reuse,

Agreed

> but I got worried about _scope_ creep when I started trying to do it.

Also agreed

PR #44, opened 2019-09-18
Bumping test coverage a bit more

jtharris:
Starting to get to diminishing returns, but I cleaned out some unused code and added some more coverage.

### DIFF
--- a/src/citrine/_serialization/properties.py
+++ b/src/citrine/_serialization/properties.py
@@ -60,12 +60,22 @@ class Property(typing.Generic[DeserializedType, SerializedType]):
         """Perform deserialization."""
 
     def deserialize_from_dict(self, data: dict) -> DeserializedType:
-        value = data.get(self.serialization_path, self.default)
+        value = data
+        fields = self.serialization_path.split('.')
+        for field in fields:
+            value = value.get(field, self.default)
         return self.deserialize(value)
 
     def serialize_to_dict(self, data: dict, value: DeserializedType) -> dict:
-        data[self.serialization_path] = self.serialize(value)
-        return data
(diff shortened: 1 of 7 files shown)


src/citrine/_serialization/properties.py
```diff
@@ -60,22 +60,12 @@ def _deserialize(self, value: SerializedType) -> DeserializedType:
         """Perform deserialization."""
 
     def deserialize_from_dict(self, data: dict) -> DeserializedType:
-        value = data
-        fields = self.serialization_path.split('.')
-        for field in fields:
-            value = value.get(field, self.default)
+        value = data.get(self.serialization_path, self.default)
```
andrewmillspaugh: Won't this break a bunch of our serialization paths (e.g. 'config.type')? Or am I misreading this?

jtharris: @andrewmillspaugh good question, but I didn't see anything that actually used that.  I think that if it did, that code as written did not seem like it would work.  Rather than write test cases for situations that don't exist, I opted to simplify it to how it was actually being used.  But if there is a use case that I'm not understanding (likely) please help me out.

andrewmillspaugh: Ah...this is the problem with keeping a bunch of important code on a long running feature branch! See https://github.com/CitrineInformatics/citrine-python/blob/7c4d973e2e1c9c9a97035c72168bf78b365baf01/src/citrine/informatics/predictors.py#L45

jtharris: @andrewmillspaugh whew!  good catch - I'll back that out.

PR #45, opened 2019-09-19
Modify tags & uids docstrings to link to data model docs

bfolie:


bfolie:
Just realized I forgot the `, optional` after each arg. Will add that and re-upload.

bfolie:
> Are they really optional, or are they required but possibly empty?

They are optional in that the user does not need to specify them to create the object (or upload the object, for that matter).

PR #46, opened 2019-09-19
List files in dataset

bfolie:
Write `list()` method for `FileLinkCollection` class. Primary complication arises because the JSON returned from the URL does not map cleanly to a FileLink dictionary. 
Tests are written using a fake session.

### DIFF
--- a/src/citrine/resources/file_link.py
+++ b/src/citrine/resources/file_link.py
@@ -2,7 +2,7 @@
 from uuid import UUID
 import os
 import mimetypes
-from typing import Iterable
+from typing import Iterable, Optional
 from boto3 import client as boto3_client
 from botocore.exceptions import ClientError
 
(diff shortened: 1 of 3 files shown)


src/citrine/resources/file_link.py
```diff
@@ -57,6 +68,77 @@ def build(self, data: dict) -> FileLink:
...
+        dict
+            A dictionary that can be built into a FileLink object.
+
+        """
+        typ = 'file_link'
+        filename = file['filename']
+
+        # The field 'versioned_url' contains some information necessary to construct a file path,
+        # but does not contain project and dataset id. It also contains extraneous information.
+        # We assert that the 'versioned_url' "picks up" where the collection path leaves off
```
kroenlein: Mixed single and double quotes is a little odd

bfolie: The single quotes are a key name, while the double quotes indicate that I'm speaking loosely (Imagine I make quotes with my fingers). I'd love a suggestion for how to make this idea of joining the versioned_url and collection path together. It's not a straight join. The versioned url is `https://a-bunch-of-other-stuff-whatever/files/{file_id}/version/{version}` and the collection path is `projects/{project_id}/datasets/{dataset_id}/files`. I'm snipping the irrelevant bits and pasting them together at "files"

PR #48, opened 2019-09-19
File Download

bfolie:
Add the ability to download a file, given a `FileCollection` and a `FileLink` object.

For a `FileLink` named `file_link` and a `Dataset` named `dataset`, original interface was `file_link.download(path: str, session: Session)`. It has been changed to `dataset.files.download(file_link: FileLink, path: str)`. There was a sleekness to `file_link.download`, but it required the user to specify the Session, which is a concept we keep less-visible. I therefore decided that it fit in more with existing interfaces to go through the collection, which is how they get all other resources.

Also contains documentation on working with files, which was merged in with PR #53


src/citrine/resources/file_link.py
```diff
@@ -289,3 +297,30 @@ def _complete_upload(self, dest_name: str, uploader: _Uploader):
...
+        local_path: str
+            Path to save file on the local computer. If `local_path` is a directory,
+            then the filename of this FileLink object will be appended to the path.
+
+        """
+        directory, filename = os.path.split(local_path)
+        if not filename:
+            filename = file_link.filename
+        if not os.path.isdir(directory):
+            os.mkdir(directory)
```
maxhutch: I think you want `os.mkdirs` in case there are multiple levels of directory given, like `my/really/annoying/path`

src/citrine/resources/file_link.py
```diff
@@ -289,3 +297,30 @@ def _complete_upload(self, dest_name: str, uploader: _Uploader):
...
+            Resource referencing the external file.
+        local_path: str
+            Path to save file on the local computer. If `local_path` is a directory,
+            then the filename of this FileLink object will be appended to the path.
+
+        """
+        directory, filename = os.path.split(local_path)
+        if not filename:
+            filename = file_link.filename
+        if not os.path.isdir(directory):
```
maxhutch: If my `cwd` is `/home/maxhutch/` and `directory=/home/maxhutch/file_test/0/1/2` but there is no `/home/maxhutch/file_test/0/1/2`, what does `isdir` return?

src/citrine/resources/file_link.py
```diff
@@ -289,3 +297,30 @@ def _complete_upload(self, dest_name: str, uploader: _Uploader):
...
+            os.mkdir(directory)
+        local_path = os.path.join(directory, filename)
```
dannycodes: reusing the same variable name in local scope is confusing. Though maybe this is normal in python? Not sure, but feels funky to read.

src/citrine/resources/file_link.py
```diff
@@ -289,3 +297,30 @@ def _complete_upload(self, dest_name: str, uploader: _Uploader):
...
+
+        content_link_path = file_link.url + '/content-link'  # get a pre-signed url
```
dannycodes: what is this magic "content-link" parameter? Can you leave a comment about that?

src/citrine/resources/file_link.py
bfolie: Good catch

src/citrine/resources/file_link.py
bfolie: I will make a comment

src/citrine/resources/file_link.py
bfolie: False.

It looks to me like `isdir` considers both absolute paths and paths from the cwd. So if that directory _did_ exist you could get there either with `directory=/home/maxhutch/file_test/0/1/2` or with `directory=file_test/0/1/2`

src/citrine/resources/file_link.py
bfolie: `local_path` isn't supposed to change in general. I'm just handling the case where the user supplies a directory instead of a full path, and I'm constructing the full path for them. For that reason it felt appropriate to use the same variable name. 🤷‍♂

bfolie: But the file will be downloaded to the absolute path, so I should add checks to make sure that the user provided a legitimate absolute path.

dannycodes: yeah that's no problem. Just looked weird, but sounds like reasonable naming

src/citrine/resources/file_link.py
bfolie: But `open` only uses absolute path, so I should add a check that it's a valid absolute path to a directory.

bfolie: I take that back. Both `open` and `isdir` cue off of starting with a backslash (or the equivalent cue in Windows) to determine if it is a relative or absolute path. So this should work.

PR #50, opened 2019-09-19
Allow filter_by_tags when user supplies a single string, as opposed to a list with one string

bfolie:
`tags = ["a tag"]` should work
`tags = "a tag"` should work
`tags = ["a tag", "another tag"]` should throw a `NotImplementedError`

### DIFF
--- a/src/citrine/resources/data_concepts.py
+++ b/src/citrine/resources/data_concepts.py
@@ -497,6 +497,8 @@ class DataConceptsCollection(Collection[ResourceType]):
             See (insert link) for a discussion of how to match on tags.
 
         """
+        # if type(tags) == str:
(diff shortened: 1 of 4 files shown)


src/citrine/resources/data_concepts.py
```diff
@@ -497,6 +497,8 @@ def filter_by_tags(self, tags: List[str],
             See (insert link) for a discussion of how to match on tags.
 
         """
+        # if type(tags) == str:
```
maxhutch: We need this, though

bfolie: And the tests caught it. Hooray tests!

maxhutch:
Test driven test driven development.

PR #54, opened 2019-09-21
Adding yet more testing for coverage

jtharris:
## Description 
This covers some util functions, properties, and resources.

### PR Type:


tests/_serialization/_data.py
```diff
@@ -15,7 +15,7 @@
 ]
 
 VALID_STRINGS = [
-    (properties.Property, 'hi', "<Property 'hi'>"),
+    (properties.Raw, 'hi', "<Property 'hi'>"),
```
andrewmillspaugh: This works? Raw property probably isn't printing correctly, then.

andrewmillspaugh: Can we either add a __str__ method to the Raw property or shift that into the base class?

jtharris: @andrewmillspaugh I think the issue is that the `Property` version is in the `__str__()` for `Property`, which is an abstract class.  So we can't create one and need to use a subclass that doesn't implement their own.  So if this is not what we want, maybe we should just remove the base method.

tests/resources/test_templates.py
```diff
@@ -37,6 +37,3 @@ def test_template_assignment():
     template = ProcessTemplate("Dry", conditions=[[humidity_template, RealBounds(0.5, 0.65, "")]])
     ProcessSpec("Dry a polymer", template=template, conditions=[
         Condition("Humidity", value=NominalReal(0.6, ""), template=humidity_template)])
-    with pytest.raises(RuntimeWarning):
-        ProcessSpec("Dry a polymer", template=template, conditions=[
-                   Condition("Humidity", value=NominalReal(0.7, ""))])
```
bfolie: I believe this is because we removed much of the client-side validation from Taurus this morning, so this no longer raises a warning.

jtharris:
@andrewmillspaugh fixed the `Raw` str test and added a dummy property to cover the baseline functionality.

PR #60, opened 2019-09-24
Resync develop with release/0.1.0

maxhutch:
In gitflow, you finish a release by merging that branch back into `develop`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ deploy:
     keep_history: true
     local_dir: ./docs/_build/html/
     on:
-      branch: develop
+      branch: master
   - provider: pypi
     user: "CitrineInformatics"
     password: "$PYPI_PASSWORD"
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -18,7 +18,7 @@ deploy:
     keep_history: true
     local_dir: ./docs/_build/html/
     on:
-      branch: develop
+      branch: master
```
sparadiso: This should still be `develop`, yes?

maxhutch: I thought we'd want to deploy docs out of master.  It'd be really cool if we could deploy from different locations to different locations, but I think that would be done in a new feature PR.

requirements.txt
```diff
@@ -1,9 +1,9 @@
-git+https://github.com/CitrineInformatics/taurus.git@develop
+taurus-citrine==0.1.0
```
maxhutch: This will likely resolve https://github.com/CitrineInformatics/citrine-python/issues/65

PR #61, opened 2019-09-24
Fix/report lazy load

sparadiso:
# Citrine Python PR

## Description 
This likely warrants some discussion. Just pushing this as a straw man.

### PR Type:


src/citrine/informatics/predictors.py
```diff
@@ -66,12 +65,12 @@ def __init__(self,
                  description: str,
                  input_descriptors: List[Descriptor],
                  output_descriptor: Descriptor,
-                 session: Optional[Session] = None):
+                 session: Session = Session()):
```
jtharris: This is rarely what you want to do.  This will create a single session when the method is interpreted (not called).  Which means that all calls get the same session, and also that the default is always created even if never used.  Better to default to `None` and explicitly write how to create/use a default session.  This probably isn't a huge deal as written but as the code changes it could bite us.

src/citrine/informatics/predictors.py
```diff
@@ -130,4 +127,4 @@ def __str__(self):
 
     def post_build(self, project_id: UUID, data: dict):
         """Creates the predictor report object."""
-        self.report = ReportResource(project_id, self.session).get(data['id'])
+        self.report = lambda: ReportResource(project_id, self.session).get(data['id'])
```
andrewmillspaugh: This is inconsistent with the rest of the api. Let's fix the underlying problem instead.

src/citrine/informatics/predictors.py
sparadiso: Note: this pattern is repeated in many places. We should clean that out!

src/citrine/informatics/predictors.py
sparadiso: 👍

PR #67, opened 2019-09-24
Resync the release with feature/orion

maxhutch:


requirements.txt
```diff
@@ -1,9 +1,9 @@
-git+https://github.com/CitrineInformatics/taurus.git@develop
+taurus-citrine==0.1.0
```
maxhutch: This should fix your builds

sparadiso:
@maxhutch I already have a PR for this - can we use that?

PR #73, opened 2019-09-27
User Methods & Tests

dannycodes:
PR Fixes up the User Model s.t. one can now do some CRUD
1. List Users works
2. Delete User works
3. Register User works
4. "me" works (get yourself)
5. GET is actually not implemented in product api, so that's been marked as not-implemented error.

### DIFF
--- a/src/citrine/resources/user.py
+++ b/src/citrine/resources/user.py
@@ -1,12 +1,13 @@
 """Resources that represent both individual and collections of users."""
 from typing import Optional
 
+from citrine._rest.collection import Collection
+from citrine._rest.resource import Resource
 from citrine._serialization import properties
 from citrine._session import Session
-from citrine._serialization.serializable import Serializable
 
 
-class User(Serializable['User']):
+class User(Resource['User']):
     """
     A Citrine User.
 
@@ -26,10 +27,10 @@ class User(Serializable['User']):
     """
 
     uid = properties.Optional(properties.UUID, 'id')
-    screen_name = properties.String('screenName')
+    screen_name = properties.String('screen_name')
     position = properties.String('position')
     email = properties.String('email')
-    is_admin = properties.Boolean('isAdmin')
+    is_admin = properties.Boolean('is_admin')
 
     def __init__(self,
                  screen_name: str,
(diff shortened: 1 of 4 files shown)


src/citrine/resources/user.py
```diff
@@ -46,13 +47,53 @@ def __init__(self,
...
+        data: dict
+          A dictionary representing the user.
+
+        Returns
+        -------
+        User
+          The user created from data.
+
+        """
+        user = User.build(data)
```
mcgalcode: Is this different than the default implementation of `build` ?

src/citrine/resources/user.py
```diff
@@ -46,13 +47,53 @@ def __init__(self,
...
+        user.session = self.session
+        return user
+
+    def register(self,
```
mcgalcode: I think that for other resources, we just use the default register implementation as well. Any motivation for this wrapper?

tests/resources/test_user.py
```diff
@@ -16,3 +44,115 @@ def test_user_collection_creation():
     session = FakeSession()
 
     assert session == UserCollection(session).session
```
mcgalcode: We'll need to figure out what the best UX is for instantiating a UserCollection. It could be a peer to projects `citrine.users`. Since this isn't a core feature though, I don't mind keeping it like this for right now.

dannycodes: So rn it _is_ a peer to `citrine.projects`.
That aligns nicely with the structure of the API.

src/citrine/resources/user.py
dannycodes: See `projects` for motivation for this pattern.

src/citrine/resources/user.py
dannycodes: Build is abstractmethod in collection so one has to implement it.

tests/resources/test_user.py
mcgalcode: Oh gotcha. I think I was misled by this line that instantiates it separately.

tests/resources/test_user.py
```diff
@@ -16,3 +44,100 @@ def test_user_collection_creation():
     session = FakeSession()
```
bfolie: Why not use the pytest fixture here?

dannycodes: Oh IDK I didn't write this line xD
Git Blame points at Justin as the author, he might remember!

PR #78, opened 2019-10-02
Support "Project Member" Endpoints in Citrine-Python

dannycodes:
Add support in citrine-python for the "project membership" endpoints. See the Postman Collection for details on the implemented methods:   Workspace: "Citrine-APIs"   Collection: "Product API Spec"   Folders: projects -> users

Seeks to bring citrine-python to parity with the product-api.

This will facilitate the migration of functional tests / sdk e2e tests into
a unified wrapper.

### DIFF
--- a/src/citrine/_session.py
+++ b/src/citrine/_session.py
@@ -129,6 +129,7 @@ class Session(requests.Session):
 
     def checked_get(self, path: str, *args, **kwargs) -> dict:
         """Execute a GET request to a URL and utilize error filtering on the response.
+
         :rtype: object
         """
         return self.checked_request('GET', path, *args, **kwargs)
--- a/src/citrine/_session.py
+++ b/src/citrine/_session.py
@@ -128,8 +128,5 @@ class Session(requests.Session):
         return self.checked_request('DELETE', path)
 
     def checked_get(self, path: str, *args, **kwargs) -> dict:
-        """Execute a GET request to a URL and utilize error filtering on the response.
-
-        :rtype: object
-        """
+        """Execute a GET request to a URL and utilize error filtering on the response."""
         return self.checked_request('GET', path, *args, **kwargs)
--- a/src/citrine/_session.py
+++ b/src/citrine/_session.py
@@ -115,7 +115,7 @@ class Session(requests.Session):
         """DELETE a particular resource as JSON."""
         return self.checked_delete(path).json()
 
-    def checked_post(self, path: str, json: dict, *args, **kwargs) -> None:
+    def checked_post(self, path: str, json: dict, *args, **kwargs) -> dict:
         """Execute a POST request to a URL and utilize error filtering on the response."""
         return self.checked_request('POST', path, *args, json=json, **kwargs)
 
@@ -128,5 +128,7 @@ class Session(requests.Session):
         return self.checked_request('DELETE', path)
 
     def checked_get(self, path: str, *args, **kwargs) -> dict:
-        """Execute a GET request to a URL and utilize error filtering on the response."""
+        """Execute a GET request to a URL and utilize error filtering on the response.
+        :rtype: object
+        """
         return self.checked_request('GET', path, *args, **kwargs)
--- a/src/citrine/resources/project.py
+++ b/src/citrine/resources/project.py
@@ -206,7 +206,7 @@ class Project(Resource['Project']):
 
         Returns
         -------
-        ListMemberResponse
+        List[User]
             Dict with key 'users' containing an array of User objects
             These are the members of the current project
 
--- a/src/citrine/resources/project.py
+++ b/src/citrine/resources/project.py
@@ -1,5 +1,5 @@
 """Resources that represent both individual and collections of projects."""
-from typing import Optional, Dict
+from typing import Optional, Dict, NamedTuple, List
 
 from citrine._session import Session
 from citrine.resources.dataset import DatasetCollection
@@ -21,6 +21,12 @@ from citrine.resources.ingredient_spec import IngredientSpecCollection
 from citrine._rest.collection import Collection
 from citrine._rest.resource import Resource
 from citrine._serialization import properties
+from citrine.resources.user import User
+
+
+class ActionRoleResponse(NamedTuple):
+    role: str
+    actions: List[str]
 
 
 class Project(Resource['Project']):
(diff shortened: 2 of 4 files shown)


bfolie:
I broke out the `flake8` into a separate command and now it's working.

src/citrine/resources/project.py
```diff
@@ -199,6 +200,48 @@ def make_private(self,
...
+        )
+        return True
+
+    def remove_member(self, user_uid: str) -> bool:
+        """
+        Remove a User from a Project.
+
+        Returns
+        -------
+        Returns True of user successfully removed
```
bfolie:
```suggestion
        Returns True if user successfully removed
```

src/citrine/resources/project.py
```diff
@@ -199,6 +200,48 @@ def make_private(self,
...
         return True
 
+    def list_members(self) -> List[User]:
+        """
+        List all of the users in the current project.
+
+        Returns
+        -------
+        ListMemberResponse
+            Dict w/ key 'users' containing an array of User objects
```
bfolie:
```suggestion
            Dict with key 'users' containing an array of User objects
```

src/citrine/resources/project.py
```diff
@@ -199,6 +200,48 @@ def make_private(self,
         })
         return True
 
+    def list_members(self) -> List[User]:
+        """
+        List all of the users in the current project.
+
+        Returns
+        -------
+        ListMemberResponse
```
bfolie:
```suggestion
        List[User]
```

src/citrine/resources/project.py
bfolie: It looks like the method extracts the 'users' field, so this should be "List of users in the current project."

dannycodes:
> I broke out the flake8 into a separate command and now it's working.

I can run flake8 locally w/out problems, but for some reason it dies when I run it through pycharm.. some brief digging leads me to the belief that the IDE doesn't really support in-line 3rd party linters. Have you (brendan) discovered a way? 

IMO not a huge problem though, just have to remember to run flake8 manually on local before pushing

src/citrine/_session.py
```diff
@@ -128,5 +128,8 @@ def checked_delete(self, path: str) -> dict:
         return self.checked_request('DELETE', path)
 
     def checked_get(self, path: str, *args, **kwargs) -> dict:
-        """Execute a GET request to a URL and utilize error filtering on the response."""
+        """Execute a GET request to a URL and utilize error filtering on the response.
+
+        :rtype: object
```
bfolie: Suggest deleting unless it can be made more specific (especially since it contradicts the typed return value of `dict`)

dannycodes: done

bfolie:
> > I broke out the flake8 into a separate command and now it's working.
> 
> I can run flake8 locally w/out problems, but for some reason it dies when I run it through pycharm.. some brief digging leads me to the belief that the IDE doesn't really support in-line 3rd party linters. Have you (brendan) discovered a way?
> 
> IMO not a huge problem though, just have to remember to run flake8 manually on local before pushing

I have never tried. I just run flake8 through the command line.

dannycodes:
> > > I broke out the flake8 into a separate command and now it's working.
> > 
> > 
> > I can run flake8 locally w/out problems, but for some reason it dies when I run it through pycharm.. some brief digging leads me to the belief that the IDE doesn't really support in-line 3rd party linters. Have you (brendan) discovered a way?
> > IMO not a huge problem though, just have to remember to run flake8 manually on local before pushing
> 
> I have never tried. I just run flake8 through the command line.

np np

src/citrine/resources/project.py
```diff
@@ -199,6 +200,48 @@ def make_private(self,
...
+        """
+        return self.session.checked_get(self._path() + "/users")["users"]
+
+    def set_member(self, user_uid, role: str, actions: List[str]) -> bool:
+        """
+        Add a User to a Project with the given 'role' and 'actions'.
+
+        Returns
+        -------
+        Returns True if member successfully added
```
bfolie: In order to get picked up by Sphinx and turned into nice documentation, I believe this has to follow the specific format of
```
Returns
---------
type
    description of return
```

src/citrine/resources/project.py
```diff
@@ -199,6 +200,48 @@ def make_private(self,
...
+        )
+        return True
+
+    def remove_member(self, user_uid: str) -> bool:
+        """
+        Remove a User from a Project.
+
+        Returns
+        -------
+        Returns True if user successfully removed
```
bfolie: Same as above

src/citrine/resources/project.py
```diff
@@ -199,6 +200,48 @@ def make_private(self,
...
         return True
 
+    def list_members(self) -> List[User]:
+        """
+        List all of the users in the current project.
+
+        Returns
+        -------
+        List[User]
+            Dict with key 'users' containing an array of User objects
```
bfolie: I think it got swallowed up by your change, but I don't think this is returning the dict. Instead it just looks like it's returning the array of User objects.

src/citrine/resources/project.py
dannycodes: Is there a way I can run a linter against that?

src/citrine/resources/project.py
dannycodes: 👍

src/citrine/resources/project.py
dannycodes: also 👍

src/citrine/resources/project.py
dannycodes: You are correct

src/citrine/resources/project.py
bfolie: The only way I know of is to build the docs locally and inspect them. AFAIK a linter won't catch it because this isn't breaking any linting rules, it just doesn't follow the specification imposed by the sphinx extension.

bfolie: This makes me think we need some sort of "contributing guidelines" doc. I'll make an issue.

dannycodes: https://pypi.org/project/restructuredtext-lint/ I found this

bfolie: Ooh, that's worth looking into.

dannycodes:
> Thanks for working with my docstring-related-nitpicks

No no, thank you! Keeping the documentation bar high is a very good standard to adhere to. Huge 💯for enforcing high standards on the product.

PR #83, opened 2019-10-09
Adding test coverage for workflow executions

jtharris:
# Citrine Python PR

## Description 
Adding test coverage for workflow executions.

### PR Type:

### DIFF
--- a/src/citrine/resources/workflow_executions.py
+++ b/src/citrine/resources/workflow_executions.py
@@ -29,7 +29,7 @@ class WorkflowExecution(Resource['WorkflowExecution']):
         self.session: Session = session
 
     def __str__(self):
-        return '<WorkflowExecution {!r}>'.format(self.uid)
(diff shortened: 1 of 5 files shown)


src/citrine/resources/workflow_executions.py
```diff
@@ -29,7 +29,7 @@ def __init__(self,
         self.session: Session = session
 
     def __str__(self):
-        return '<WorkflowExecution {!r}>'.format(self.uid)
+        return '<WorkflowExecution {!r}>'.format(str(self.uid))
```
andrewmillspaugh: Couldn't you just remove the !r and get the same behavior?

jtharris: @andrewmillspaugh i think the `!r` adds the quotes, which we could also add back in, but this felt like the path of least resistance.

andrewmillspaugh: Ah, I thought it was "render this thing using the __repr__ method"

PR #87, opened 2019-10-11
Table READ

dannycodes:
Add support for table READ in citrine-python

  Can do a GET for a table w/ ID and Version
  Can do a READ from S3 both locally and on remote


src/citrine/_utils/functions.py
```diff
@@ -124,3 +126,24 @@ def object_to_link_by_uid(json: dict) -> dict:
...
+
+
+def write_file_locally(content, local_path: str):
+    """Take content from remote and ensure path exists."""
+    directory, filename = os.path.split(local_path)
+    if filename == "":
+        raise ValueError("A filename must be provided in the path")
+    if not os.path.isdir(directory):
+        os.makedirs(directory)
+    local_path = os.path.join(directory, filename)
```
bfolie: I think this line is unnecessary, since neither `directory` nor `filename` have been modified, hence `local_path` will not change.

src/citrine/resources/table.py
```diff
@@ -0,0 +1,73 @@
...
+    While data objects can represent complex materials data, the format
+    is NOT conducive to analysis and machine learning. Tables, however,
+    can be used to 'flatten' data objects into useful projections.
+    """
+
+    _response_key = 'table'
+
+    uid = properties.Optional(properties.UUID(), 'id')
+    version = properties.Optional(properties.Integer, 'version')
+    download_url = properties.Optional(properties.String, 'signed_download_url')
```
bfolie: These fields should be added to the class docstring as `Attributes` (see `Project` for an example)

src/citrine/resources/table.py
```diff
@@ -0,0 +1,73 @@
...
+        data_location = self.download_url
+        data_location = rewrite_s3_links_locally(data_location)
+        response = requests.get(data_location)
+        write_file_locally(response.content, local_path)
+
+
+class TableCollection(Collection[Table]):
+    """Represents the collection of all tables assocated with a project."""
+
+    _path_template = 'projects/{project_id}/tables/{table_id}/versions/{version}'
```
bfolie: I'd say that the path for the collection is 'projects/{project_id}/tables', and the rest of the path should be added as needed

src/citrine/resources/table.py
dannycodes: 👍

src/citrine/resources/table.py
```diff
@@ -0,0 +1,73 @@
...
+        self.project_id = project_id
+        self.session: Session = session
+
+    def get(self, uid: Union[UUID, str], version: int) -> Table:
+        """Get a Table's metadata."""
+        path = self._path_template.format(
+            project_id=self.project_id,
+            table_id=uid,
+            version=version,
+        )
```
bfolie: Modifying `_path_template` as above, this would become `self._get_path(uid) + "/versions/{}".format(version)`

src/citrine/resources/table.py
```diff
@@ -0,0 +1,73 @@
...
+    def build(self, data: dict) -> Table:
+        """Build an individual Table from a dictionary."""
+        table = Table.build(data)
+        table.project_id = self.project_id
+        table.session = self.session
+        return table
+
+    def register(self, model: Table) -> Table:
+        """Tables cannot be created at this time."""
+        pass
```
bfolie: Do we want to raise an Exception here and tell the user that they can't create tables?

dannycodes: What kind of exception were you thinking?

src/citrine/resources/table.py
dannycodes: 👍

src/citrine/resources/table.py
dannycodes: 👍

src/citrine/_utils/functions.py
dannycodes: nice catch! Always good to save a line

src/citrine/resources/table.py
bfolie: I'm not an Exceptions master, but I think a `RuntimeException` with a clear error message is the most appropriate (I thought about `NotImplementedException`, but that implies that that the method is supposed to be implemented by an inheriting class, and that is not the case here)

tests/_util/test_functions.py
```diff
@@ -67,3 +69,25 @@ def test_validate_type_set_type():
 def test_object_to_link_by_uid_missing_uids():
     assert {'foo': 'bar'} == object_to_link_by_uid({'foo': 'bar'})
 
+
+def test_rewrite_s3_links_locally():
```
bfolie: Could you add a docstring explaining what this is for? Otherwise it will look bizarre and arbitrary.

src/citrine/resources/table.py
dannycodes: NVM it's a NotImplementedError

tests/_util/test_functions.py
```diff
@@ -67,3 +69,25 @@ def test_validate_type_set_type():
...
+    assert "http://localhost:9572" == rewrite_s3_links_locally("http://localstack:4572")
+
+@patch("os.path.isdir")
+@patch("os.path.join")
+@patch("os.makedirs")
+def test_write_file_locally(mock_makedirs, mock_join, mock_isdir):
+    mock_join.return_value = "/User/is/fake/myfile.pdf"
+    with patch('os.path.split') as mock_split:
```
bfolie: Why is it necessary to mock the `os.path.split` method? It's getting `"/User/is/fake/myfile.pdf"`, which it should be able to split.

src/citrine/resources/table.py
```diff
@@ -0,0 +1,78 @@
...
+    def build(self, data: dict) -> Table:
+        """Build an individual Table from a dictionary."""
+        table = Table.build(data)
+        table.project_id = self.project_id
+        table.session = self.session
+        return table
+
+    def register(self, model: Table) -> Table:
+        """Tables cannot be created at this time."""
+        raise NotImplementedError('Creating Tables is not supported at this time.')
```
bfolie: I think you edited the line as I commented, so my comment got erased. But I believe this should be a `RuntimeError` because `NotImplementedError` it meant for situations in which the method is supposed to be implemented in inheriting classes. But that's not the case here.

bfolie: https://docs.python.org/3.3/library/exceptions.html#RuntimeError
https://docs.python.org/3.3/library/exceptions.html#NotImplementedError

tests/_util/test_functions.py
dannycodes: I documented this in the `functions.py` source code (and updated docs to be more verbose

tests/_util/test_functions.py
dannycodes: Yeah that worked fine: good catch!

src/citrine/resources/table.py
dannycodes: Yeah I took a gander as well. RuntimeError seems fine for now

PR #97, opened 2019-10-18
Citrine v0.3.0 is released!

maxhutch:
This release supports the deletion of data objects and the reading of tables from the data service.  Also included in code is experimental support for project member management, but note that it is highly experimental and will very likely change.

## What's new
 * Support for deleting objects!  Deletion is conditional on not orphaning any objects that link to them.
 * Support for reading tables by downloading them as csv files.

## Coming soon
 * Improved project member management
 * Delete datasets


src/citrine/resources/project.py
```diff
@@ -199,6 +206,49 @@ def make_private(self,
...
+        return self.session.checked_get(self._path() + "/users")["users"]
+
+    def set_member(self, user_uid, role: str, actions: List[str]) -> bool:
+        """
+        Add a User to a Project with the given 'role' and 'actions'.
+
+        Returns
+        -------
+        bool
+            Returns True if member successfully added
```
asantas93: To me this isn't meaningful when an exception is thrown rather `False` being returned when the member is not added successfully. Might as well not have a return value.

src/citrine/resources/data_concepts.py
```diff
@@ -611,3 +613,19 @@ def filter_by_name(self, name: str, exact: bool = False,
...
+        Parameters
+        ----------
+        uid: Union[UUID, str]
+            The ID.
+        scope: str
+            The scope of the uid, defaults to Citrine scope ('id')
+
+        """
+        path = self._get_path() + "/{}/{}".format(scope, uid)
+        self.session.delete_resource(path)
```
bfolie: Did we establish that this line throws an error if the attempted delete fails?

src/citrine/resources/project.py
bfolie: We could also return a `Response` object

asantas93: Sorry, I was looking at this out of context of all the changes already having been approved.

src/citrine/resources/data_concepts.py
djack: yes, and it's tested here https://github.com/CitrineInformatics/nextgen-devkit/blob/master/integration-tests/tests/user_journeys_strike-team-data/test_delete.py

PR #102, opened 2019-10-24
PLA-1539 throw more specific HTTP exceptions

asantas93:
# Citrine Python PR

## Description 
Makes errors thrown from HTTP exceptions more specific, parses api errors to a usable model.

### PR Type:

### DIFF
--- a/src/citrine/resources/api_error.py
+++ b/src/citrine/resources/api_error.py
@@ -1,9 +1,11 @@
+from copy import copy
 from typing import Optional, List
 
 from taurus.entity.dict_serializable import DictSerializable
 
 
 class ValidationError(DictSerializable):
+    """A user-facing error message describing why their request was invalid."""
 
     def __init__(self, failure_message: Optional[str] = None, property: Optional[str] = None,
                  input: Optional[str] = None):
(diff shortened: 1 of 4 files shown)


asantas93:
Review welcome but I need to do some cleanup / address failures in travis

src/citrine/resources/api_error.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def __init__(self, code: int, message: str, validation_errors: List[ValidationError] = None):
+        self.code = code
+        self.message = message
+        self.validation_errors = validation_errors or []
+
+    @classmethod
+    def from_dict(cls, d):
+        d = copy(d)
+        # TODO: deserialize to correct type automatically
```
asantas93: Is anyone aware of a way to do this? My impression is that we rely on the type string to pick the right type (when using the taurus encoder), but I don't think that helps here because `ApiError` has no type string.

asantas93:
Ready for review now

src/citrine/resources/api_error.py
bfolie: Were you hoping to have it done automatically so that no `from_dict` method is necessary for the `ApiError` class? The Taurus decoder really relies on the `typ` field, so I also do not know how you would do this. Is it necessary for `ApiError` and `ValidationError` to implement `DictSerializable`? What's the benefit?

asantas93: That was the hope, yes. I think we would need some schema definition in the models or use some reflection to infer the right type.

Having the extend `DictSerializable` just makes it less tedious to construct them from a dict (fewer manual key checks).

PR #123, opened 2019-11-07
Add modules collection

andrewmillspaugh:
# Citrine Python PR

## Description 
Add a `project.modules` collection

### PR Type:


src/citrine/informatics/modules.py
```diff
@@ -0,0 +1,26 @@
...
+    """
+
+    _response_key = None
+
+    @classmethod
+    def get_type(cls, data) -> Type['Module']:
+        """Return the subtype."""
+        from citrine.informatics.design_spaces import DesignSpace
+        from citrine.informatics.processors import Processor
+        from citrine.informatics.predictors import Predictor
```
kyle-laplante: why import these here instead of at the top?

src/citrine/informatics/modules.py
```diff
@@ -0,0 +1,26 @@
...
+        return {
+            'DESIGN_SPACE': DesignSpace,
+            'PROCESSOR': Processor,
+            'PREDICTOR': Predictor
+        }[data['module_type']].get_type(data)
```
kyle-laplante: can this dictionary be a class-level variable instead of creating it then throwing it away on every call to this method?

i.e.

```
class Module(...):
  MODULE_TYPES = {'DESIGN_SPACE': DesignSpace, ... }

  @classmethod
  def get_type(cls, data):
    module = cls.MODULE_TYPES.get(data['module_type'])
    if module is not None:
      return module.get_type(data)
    raise Exception("{} is not a valid module type".format(data['module_type']))
```

src/citrine/informatics/modules.py
andrewmillspaugh: Otherwise you have a circular import...There is probably a bigger conversation to be had here, but I wanted to do one thing at a time.

src/citrine/informatics/modules.py
andrewmillspaugh: Yeah not a bad idea.

src/citrine/informatics/modules.py
kyle-laplante: ohhh. thats generally a sign that things are not organized properly. Who do the sub-types need to subclass `Module`? Cant `Module` just be alongside `Predictor`, `Processor` and `DesignSpace`?

jtharris: I think the smell here is that a super-type knows about its subtypes directly.  That arrow should be one-way.  If a super-type has to know about all types that inherit from it you are breaking a pretty big design principal of OO.

andrewmillspaugh: Agreed. Playing around with it right now to figure out if there's a way around this.

andrewmillspaugh: I think this is a case of trying to do two things with the same object. For inheritance, obviously the subtype needs to know about the supertype. For deserialization, I don't see how you get around the supertype knowing about the subtype. In fact that pattern is used throughout the citrine-python codebase. It's just usually in the same file, so it doesn't seem as bad.

jtharris: I think the answer is to not have the supertype responsible for deserialization.  This is sort of an anti-pattern that I see a lot in the scala codebase as well, but if we can get it to work then whatevs

andrewmillspaugh: We can certainly get it to work. It works in this example. I don't see how you do deserialization without having the supertype be responsible for it - you're not going to know the subtype ahead of time.

andrewmillspaugh: I have read a bunch of different serialization library's code and do not see anyone that has a solution much different than this. The "best" solution I've seen involves using decorators to add objects to a global registry. Either way, I think this is a different conversation.

src/citrine/informatics/modules.py
andrewmillspaugh: Turns out I cannot do that with the dynamic import.

tests/serialization/test_modules.py
```diff
@@ -0,0 +1,26 @@
...
+
+from citrine.informatics.modules import Module
+from citrine.informatics.design_spaces import ProductDesignSpace
+from citrine.informatics.predictors import SimpleMLPredictor
+from citrine.informatics.processors import GridProcessor
+
+
+def test_polymorphic_design_space_deserialization(valid_product_design_space_data):
+    """Ensure that a deserialized ProductDesignSpace looks sane."""
+    module: Module = Module.build(valid_product_design_space_data)
```
jtharris: I think we can defer this to a later PR, but I think the answer to getting around the circular dependencies is to not have `Module` responsible for building this.  If you have some other `Builder` class that knows how to translate json into a concrete class, then `Module` doesn't need to know about any subclasses.  That was my only point previously.

andrewmillspaugh: Sure, but that's completely changing the pattern that the entire library is based on.

PR #131, opened 2019-11-12
Upgraded Arrow to silence warnings and added a Dockerfile for testing

crleblanc:
# Citrine Python PR

## Description 
This is a bugfix for https://citrine.atlassian.net/browse/PLA-1900.  Using arrow.get() with a date string would raise a warning.  We're parsing ISO8601 strings so this warning does not impact us.  Upgrading to the latest version has silenced this warning.  See the Jira issue for more detail.

I added a dockerfile under the scripts directory called Dockerfile.pytest.  This is a clean way of testing code and dependencies in an isolated environment.

The bugfix was simple and no changes to the code were required so it won't impact test coverage/docstrings/pep-8 compliance or have any downstream consequences as our use of Arrow is limited to string/int/float parsing with arrow.get().

### PR Type:

### DIFF
--- a/scripts/Dockerfile.pytest
+++ b/scripts/Dockerfile.pytest
@@ -1,13 +1,17 @@
-# Developmental Docker container
+# Developmental Docker container, performing the same tasks as in .travis.yml
 FROM python:3.8
 
 WORKDIR /src
 COPY requirements.txt .
-RUN pip install -r requirements.txt
+RUN pip install -U -r requirements.txt
 
 COPY test_requirements.txt .
-RUN pip install -r test_requirements.txt
+RUN pip install -U -r test_requirements.txt
 
 COPY . .
-RUN python setup.py develop
-CMD pytest
+RUN pip install --no-deps -e .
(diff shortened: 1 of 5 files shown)


scripts/Dockerfile.pytest
```diff
@@ -0,0 +1,17 @@
...
+COPY test_requirements.txt .
+RUN pip install -U -r test_requirements.txt
+
+COPY . .
+RUN pip install --no-deps -e .
+CMD pytest --cov=src/ --cov-report term-missing \
+    --cov-report term:skip-covered \
+    --cov-fail-under=100 -s -r . \
+    && flake8 src \
+    && cd docs; make html; cd ..
```
mcgalcode: I think if this container is just used for testing then we can forego the documentation building.

Also, is there a way to run a specific test? I'd recommend updating the README with instructions for invoking this container.

crleblanc: Good suggestions @mcgalcode.  I added an ENTRYPOINT and CMD to make this easy.  I also added some docs with simple examples.

PR #132, opened 2019-11-14
Proposed fix for PLA-1952: Annotating ValueError exceptions

### LINKED ISSUE #121: Include field and object name in field type validation error
Currently, if you try to assign `str` to an `int` field or a `None` to a non-optional field, you get an error message like:
```
ValueError: None is not one of valid types: <class 'str'>!
```
which forces you to navigate the stacktrace in order to figure out where its coming from.  It would be really helpful to include the field name and object type in that value error, e.g.:
```
ValueError: None is not one of valid types: <class 'str'> for MaterialRun.name!
```

crleblanc:
Adding the name of the class and the invalid parameter name that
caused the exception.  Added a new optional keyword argument
called base_class to avoid breaking existing code.  Also added
a test that exercises this new path.

There was a test with a duplicate name which I corrected.

Resolves #121

# Citrine Python PR

## Description 

### PR Type:

### DIFF
--- a/tests/_serialization/test_simple_properties.py
+++ b/tests/_serialization/test_simple_properties.py
@@ -6,6 +6,7 @@ from citrine._serialization.properties import (
     Integer,
     String,
     Float,
+    UUID,
     Datetime,
     LinkByUID,
     LinkOrElse,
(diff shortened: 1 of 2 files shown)


tests/_serialization/test_simple_properties.py
```diff
@@ -50,8 +51,23 @@ def test_invalid_deserialization_type(prop_type, serialized):
...
+
+    prop = prop_type()
+    prop.serialization_path = 'ser_path'
+    with pytest.raises(ValueError) as excinfo:
+        prop.deserialize(serialized, base_class=BaseTest().__class__)
+
+    # Check that the exception includes the calling class name and argument
+    if not isinstance(prop, UUID):
+        assert 'BaseTest:ser_path' in str(excinfo.value)
+
```
mcgalcode: Could we add another test here that tries to deserialize an invalid serialization of an existing module (say, a dataset) and demos that the error is what we expect?

crleblanc: No worries, I just added another test that uses a Dataset object as the base_class.

It's a little awkward to supply the 'base_class' parameter but I don't think there's any way to get the class name from an attribute.  For example, if d=Dataset() and d.xyz = 123, d.xyz is just a plain Python integer and I can't find a way of getting the name of it's class.  This is why I had to include this new (optional) parameter.

PR #134, opened 2019-11-22
Pla 2095/audit info

bfolie:
# Citrine Python PR

## Description 
This PR allows for the serialized data objects coming from the backend to have additional fields that are not in the data model. The specific example here is `audit_info`. These fields are not serialized , nor are they expected during deserialization. Instead they are intercepted at the beginning of the `build()` method and attached to the object later on. No warning is thrown and the deserialized object _does_ have an `audit_info` field (if it was in the backend response).

This change will quiet the warnings that currently appear in `DictSerializable`, as long as the serialized object is only one level deep. If deserializing nested objects the warnings will still get thrown. That's because the assumption "serialized objects returned by the backend correspond to the data model" is baked in pretty firmly, and the code is enthusiastic about using the Taurus serialization and deserialization code as much as possible. I don't see an easy way to change that.
If deserializing a full material history the warnings will also get thrown, but I do think that can be changed. That's because only top-level objects come back, so the code could go through them all and strip out `audit_info`. This would quiet the warnings, but the resulting rehydrated material history would *not* have `audit_info` as a field.

### PR Type:

### DIFF
--- a/src/citrine/resources/audit_info.py
+++ b/src/citrine/resources/audit_info.py
@@ -26,6 +26,11 @@ class AuditInfo:
         self.updated_by = updated_by
         self.updated_at = updated_at
 
+    def __repr__(self):
+        return 'Created by: {}\nCreated at: {}\nUpdated by: {}\nUpdated at: {}'.format(
+            self.created_by, self.created_at, self.updated_by, self.updated_at
+        )
+
     def __str__(self):
         create_str = 'Created by user {} at time {}'.format(
             self.created_by, self._pprint_datetime(self.created_at))
--- a/src/citrine/resources/audit_info.py
+++ b/src/citrine/resources/audit_info.py
@@ -1,0 +1,49 @@
+from typing import Optional
+import arrow
+
+
+class AuditInfo:
+    """
+    Model that holds audit metadata. AuditInfo objects should not be created by the user.
+
+    Parameters
+    ----------
+    created_by: str
+        ID of the user who created the object
+    created_at: int
+        Time, in ms since epoch, at which the object was created
+    updated_by: Optional[str]
+        ID of the user who most recently updated the object
+    updated_at: Optional[int]
+        Time, in ms since epoch, at which the object was most recently updated
+
+    """
+
+    def __init__(self, created_by: str, created_at: int,
+                 updated_by: Optional[str] = None, updated_at: Optional[int] = None):
+        self.created_by = created_by
+        self.created_at = created_at
+        self.updated_by = updated_by
+        self.updated_at = updated_at
+
+    def __repr__(self):
+        return 'Created by: {}\nCreated at: {}\nUpdated by: {}\nUpdated at: {}'.format(
+            self.created_by, self.created_at, self.updated_by, self.updated_at
+        )
+
+    def __str__(self):
+        create_str = 'Created by user {} at time {}'.format(
+            self.created_by, self._pprint_datetime(self.created_at))
+        if self.updated_by is not None or self.updated_at is not None:
+            update_str = '\nUpdated by user {} at time {}'.format(
+                self.updated_by, self._pprint_datetime(self.updated_at))
+        else:
+            update_str = ''
+        return create_str + update_str
+
+    def __eq__(self, other):
+        return self.__repr__() == other.__repr__()
+
+    @staticmethod
+    def _pprint_datetime(create_time: int):
+        return arrow.get(create_time / 1000).datetime
--- a/src/citrine/resources/audit_info.py
+++ b/src/citrine/resources/audit_info.py
@@ -1,8 +1,12 @@
 from typing import Optional
-import arrow
+from uuid import UUID
+from datetime import datetime
+
+from citrine._serialization.serializable import Serializable
+from citrine._serialization import properties
 
 
-class AuditInfo:
+class AuditInfo(Serializable):
     """
     Model that holds audit metadata. AuditInfo objects should not be created by the user.
 
@@ -19,31 +23,32 @@ class AuditInfo:
 
     """
 
-    def __init__(self, created_by: str, created_at: int,
-                 updated_by: Optional[str] = None, updated_at: Optional[int] = None):
+    created_by = properties.UUID('created_by')
+    created_at = properties.Datetime('created_at')
+    updated_by = properties.Optional(properties.UUID, 'updated_by')
+    updated_at = properties.Optional(properties.Datetime, 'updated_at')
+
+    def __init__(self, created_by: UUID, created_at: datetime,
+                 updated_by: Optional[UUID] = None, updated_at: Optional[datetime] = None):
         self.created_by = created_by
         self.created_at = created_at
         self.updated_by = updated_by
         self.updated_at = updated_at
 
     def __repr__(self):
-        return 'Created by: {}\nCreated at: {}\nUpdated by: {}\nUpdated at: {}'.format(
+        return 'Created by: {!r}\nCreated at: {!r}\nUpdated by: {!r}\nUpdated at: {!r}'.format(
             self.created_by, self.created_at, self.updated_by, self.updated_at
         )
 
     def __str__(self):
         create_str = 'Created by user {} at time {}'.format(
-            self.created_by, self._pprint_datetime(self.created_at))
+            self.created_by, self.created_at)
         if self.updated_by is not None or self.updated_at is not None:
             update_str = '\nUpdated by user {} at time {}'.format(
-                self.updated_by, self._pprint_datetime(self.updated_at))
+                self.updated_by, self.updated_at)
         else:
             update_str = ''
         return create_str + update_str
 
     def __eq__(self, other):
         return self.__repr__() == other.__repr__()
-
-    @staticmethod
-    def _pprint_datetime(create_time: int):
-        return arrow.get(create_time / 1000).datetime
--- a/src/citrine/resources/audit_info.py
+++ b/src/citrine/resources/audit_info.py
@@ -26,11 +26,6 @@ class AuditInfo:
         self.updated_by = updated_by
         self.updated_at = updated_at
 
-    def __repr__(self):
-        return 'Created by: {}\nCreated at: {}\nUpdated by: {}\nUpdated at: {}'.format(
-            self.created_by, self.created_at, self.updated_by, self.updated_at
-        )
-
     def __str__(self):
         create_str = 'Created by user {} at time {}'.format(
             self.created_by, self._pprint_datetime(self.created_at))
--- a/src/citrine/resources/data_concepts.py
+++ b/src/citrine/resources/data_concepts.py
@@ -1,6 +1,6 @@
 """Top-level class for all data concepts objects and collections thereof."""
 from uuid import UUID
-from typing import TypeVar, Type, List, Dict, Union, Optional, Any
+from typing import TypeVar, Type, List, Dict, Union, Optional
 from copy import deepcopy
 from abc import abstractmethod
 
@@ -11,7 +11,6 @@ from citrine._serialization.serializable import Serializable
 from citrine._serialization.properties import Property, LinkOrElse, Object
 from citrine._serialization.properties import List as PropertyList
 from citrine._serialization.properties import Optional as PropertyOptional
-from citrine.resources.audit_info import AuditInfo
 from citrine._utils.functions import (
     validate_type, scrub_none,
     replace_objects_with_links, get_object_id)
@@ -57,15 +56,9 @@ class DataConcepts(PolymorphicSerializable['DataConcepts']):
     Only populated if the :func:`get_type` method is invoked.
     """
 
-    def __init__(self, typ: str, audit_info: Optional[Dict[str, Any]] = None):
+    def __init__(self, typ: str):
         self.typ = typ
         self.session = None
-        if audit_info is None:
-            audit_info = dict()
-        self.audit_info = AuditInfo(created_by=audit_info.get('created_by', None),
-                                    created_at=audit_info.get('created_at', None),
-                                    updated_by=audit_info.get('updated_by', None),
-                                    updated_at=audit_info.get('updated_at', None))
 
     @classmethod
     def build(cls, data: dict, session: Session = None):
@@ -113,8 +106,10 @@ class DataConcepts(PolymorphicSerializable['DataConcepts']):
         data_copy_dict.pop(DataConcepts._type_key, None)
         cls._build_child_objects(data_copy_dict, data)
 
-        data_concepts_object = cls(**data_copy_dict, **client_only_dict)
+        data_concepts_object = cls(**data_copy_dict)
         data_concepts_object.session = session
+        for client_key in cls._client_keys:
+            setattr(data_concepts_object, client_key, client_only_dict.get(client_key))
 
         cls._build_discarded_objects(data_concepts_object, data, session)
         return data_concepts_object
--- a/src/citrine/resources/data_concepts.py
+++ b/src/citrine/resources/data_concepts.py
@@ -90,10 +90,12 @@ class DataConcepts(PolymorphicSerializable['DataConcepts']):
         # They will be attached to the deserialized object later.
         client_only_dict = dict()
         for client_key in cls._client_keys:
-            client_only_dict.__setitem__(client_key, data_copy.pop(client_key, None))
+            val = DataConcepts._pop_field(data_copy, client_key)
+            client_only_dict.__setitem__(client_key, val)
 
         # Running through a taurus loads/dumps cycle validates all of the fields and ensures
         # the object is now a dictionary with a well-understood structure
+        # Note that this process modifies data_copy and makes it unsuitable for further use
         data_copy_dict = loads(dumps(data_copy)).as_dict()
 
         # Check the type--it should either correspond to LinkByUID or to this class.
@@ -103,7 +105,7 @@ class DataConcepts(PolymorphicSerializable['DataConcepts']):
 
         # Remove the top-level "type" field and build child objects
         data_copy_dict.pop(DataConcepts._type_key, None)
-        cls._build_child_objects(data_copy_dict, data_copy)
+        cls._build_child_objects(data_copy_dict, data)
 
         data_concepts_object = cls(**data_copy_dict)
         data_concepts_object.session = session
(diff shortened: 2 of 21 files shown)


src/citrine/resources/data_concepts.py
```diff
@@ -40,8 +40,13 @@ class DataConcepts(PolymorphicSerializable['DataConcepts']):
 
     """
 
-    _local_keys = ['type']
-    """list of str: keys that appear in the serialized dictionary but not in the object itself."""
+    _type_key = "type"
+    """str: key used to determine type of serialized object."""
+
+    _client_keys = []
```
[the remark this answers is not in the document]
asantas93: 1 possible interface would be to have use a dict here:
```python
_client_keys = {
  'audit_info': AuditInfo
}

def from_dict(cls, d):
  client_fields = []  
  for k, typ in _client_keys.items():
    vdict = d.pop(k)
    v = typ.from_dict(vdict)
    client_fields.append(v)
  ... # typical from_dict stuff
  for k, v in zip(_client_keys.keys(), client_fields):
    set_attr(obj, k, v)
```
This code is fake and won't run. Lmk if you need help understanding the approach I'm suggesting. Or if there's reason I'm overlooking to stick with something closer to what you've already written.

asantas93:
Re: warnings -- they are there to bug the user into upgrading, or inform the developer that there is a bug. We can adjust log levels if the goal is just to silence warnings.

src/citrine/resources/data_concepts.py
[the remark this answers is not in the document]
bfolie: Are you suggesting something like this go in `DictSerializable.from_dict()`? Or that `DataConcepts` get a `from_dict()` method?
The former doesn't make sense to me because if we're in `DictSerializable` then we in Taurus, below the level of the client.

src/citrine/resources/ingredient_run.py
```diff
@@ -62,6 +62,8 @@ class IngredientRun(DataConcepts, Resource['IngredientRun'], TaurusIngredientRun
 
     _response_key = TaurusIngredientRun.typ  # 'ingredient_run'
 
+    _client_keys = ["audit_info"]
```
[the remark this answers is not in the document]
bfolie: The root is `DataConcepts`, which is extended by `Property`, `Parameter`, and `Condition`, none of which have `audit_info`.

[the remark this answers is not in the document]
asantas93: I believe my suggestion is for it to be put in `BaseObject` in taurus.

src/citrine/resources/data_concepts.py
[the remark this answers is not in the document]
asantas93: I _am_ suggesting that it goes in `DictSerializable`. Why would that be problematic? We will have the same issue in taurus if it interacts with the backend independent of citrine-python.

bfolie:
> I commented on the JIRA ticket -- I think there should be an `AuditInfo` class and read-only access to it like we have for `measurements` on materials. I like your `_client_fields` idea and think it can be used more broadly on `DictSerializable`.

It's true that `measurements` is read-only on `materials`, but it's still a fundamental part of the data model. It's just that we only allow the bi-directional link to be established in one direction. This, by contrast, would be putting an object into the data model _that only has meaning due to interaction with an external database_. My feeling is that Taurus should be self-contained.

asantas93:
> It's true that `measurements` is read-only on `materials`, but it's still a fundamental part of the data model. It's just that we only allow the bi-directional link to be established in one direction. This, by contrast, would be putting an object into the data model _that only has meaning due to interaction with an external database_. My feeling is that Taurus should be self-contained.

That's an interesting point about this data coming from interaction with a database. I wasn't thinking of taurus in terms of it being a library for using our data model outside the context of our backend. https://citrine.atlassian.net/browse/PLA-891 says to add the fields to the data model, but I'm curious about what the people who thought up this feature were imagining would happen here.

src/citrine/resources/ingredient_run.py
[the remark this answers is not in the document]
asantas93: sidenote -- do you know what the point of having attributes extend data concepts is?

bfolie:
> That's an interesting point about this data coming from interaction with a database. I wasn't thinking of taurus in terms of it being a library for using our data model outside the context of our backend. https://citrine.atlassian.net/browse/PLA-891 says to add the fields to the data model, but I'm curious about what the people who thought up this feature were imagining would happen here.

Good point that the ticket does say this becomes part of the data model.

One salient point: in a thread on Slack in the data channel, I proposed that this change should not affect the data model, and Hutch agreed. It's possible that the plan changed, but also possible that he just wasn't thinking about PLA-891, which was made months ago. @maxhutch what's the status of PLA-891, and do we want to add `AuditInfo` to the data model?

asantas93:
@kroenlein also curious what you think of the `audit_info` in-or-not-in taurus question

src/citrine/resources/data_concepts.py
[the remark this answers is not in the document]
bfolie: It's not that it's problematic, it's just that if it's in `DictSerializable` then that suggests it's part of the data model, in which case we don't need to treat it differently.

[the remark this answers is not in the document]
asantas93: Yeah, my point here definitely assumes taurus is meant to see this field in one way or another.

src/citrine/resources/ingredient_run.py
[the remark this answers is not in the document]
bfolie: It's because attributes contain attribute templates, so they need the same complicated `build` logic in order to properly serialize and deserialize those.
Conversely, a `Value` doesn't need to extend `DataConcepts` or have a Citrine-Python implementation, because it's just a simple dictionary without anything complicated in its fields.

src/citrine/resources/data_concepts.py
```diff
@@ -79,56 +85,68 @@ def build(cls, data: dict, session: Session = None):
             An object corresponding to a data concepts resource.
 
         """
+        data_copy = deepcopy(data)
```
asantas93: `deep_copy` is significantly more computationally expensive than a shallow `copy`. Is this required?

src/citrine/resources/data_concepts.py
```diff
@@ -79,56 +85,68 @@ def build(cls, data: dict, session: Session = None):
...
+        # Extract the values that are only for the client.
+        # They will be attached to the deserialized object later.
+        client_only_dict = dict()
+        for client_key in cls._client_keys:
+            val = DataConcepts._pop_field(data_copy, client_key)
+            client_only_dict.__setitem__(client_key, val)
```
asantas93: Is this distinct from
```suggestion
            client_only_dict[client_key] = val
```
?

src/citrine/resources/data_concepts.py
```diff
@@ -79,56 +85,68 @@ def build(cls, data: dict, session: Session = None):
...
+
         # Running through a taurus loads/dumps cycle validates all of the fields and ensures
         # the object is now a dictionary with a well-understood structure
-        data_copy_dict = loads(dumps(deepcopy(data))).as_dict()
+        # Note that this process modifies data_copy and makes it unsuitable for further use
+        data_copy_dict = loads(dumps(data_copy)).as_dict()
```
asantas93: (non-blocking / not-really-actionable) I know this code was already here (and in numerous other places), but in general json (de)serialization isn't totally nontrivial computationally. Needing to have code like this seems like a sign that we should be leveraging taurus's serialization constructs better.

src/citrine/resources/data_concepts.py
```diff
@@ -79,56 +85,68 @@ def build(cls, data: dict, session: Session = None):
...
 
-        cls._remove_local_keys(data_copy_dict)
+        # Remove the top-level "type" field and build child objects
+        data_copy_dict.pop(DataConcepts._type_key, None)
         cls._build_child_objects(data_copy_dict, data)
 
         data_concepts_object = cls(**data_copy_dict)
         data_concepts_object.session = session
+        for client_key in cls._client_keys:
+            data_concepts_object.__setattr__(client_key, client_only_dict.get(client_key))
```
asantas93: FYI you should be able to use `setattr(obj, k ,v)`

src/citrine/resources/data_concepts.py
asantas93: Looking at this a little deeper, I think you can get away with a single iteration over `data` vaguely like
```python
client_dict = {}
taurus_dict = {}
for k, v in data.items():
    if k in cls._client_keys:
        client_dict[k] = v
    else:
        taurus_dict[k] = v
```

asantas93:
Btw, one issue I think you may run into if we do end up putting `audit_info` in `DictSerializable` is that some of our equality checks may break because `__eq__` is defined on `DictSerializable`, and audit info will break equality between object pre and post write.

src/citrine/resources/data_concepts.py
bfolie: Here's the motivation for this code: in order to construct a Taurus object from a dictionary, the pointed-to objects must be full objects, not more dictionaries. In other words, you can build a process run from
```
{
  'typ': 'process_run',
  'name': 'my process run',
  'spec': <Process Spec object>
}
```
But you can't build it from
```
{
  'typ': 'process_run',
  'name': 'my process run',
  'spec': {
    'typ': 'process_spec',
    'name': 'my process spec'
  }
}
```
Running `loads(dumps(foo)).as_dict` reconstructs a complete object, runs all of Taurus's object creation logic (very important because it does validation _and_ sets up soft links) and then collapses the top-level object into a dictionary _without_ affecting the pointed-to objects. It takes what can be an uncertain input (is it a fully-serialized dictionary or a partially-serialized dictionary or a Taurus object?) and turns it into something well-understood (a Taurus object). The goal was to not burden the client with the details of what Taurus does when it's rehydrating objects, so if anything I'd say we're over-reliant on Taurus's serialization constructs, to the point that we're running into problems if the serialized JSON from the backend _doesn't_ match what Taurus expects.

src/citrine/resources/data_concepts.py
bfolie: It's required because `loads(dumps(foo))` mutates `foo`

src/citrine/resources/data_concepts.py
bfolie: This is not to say that the current method cannot be improved upon, just that there are two things tying our hands:
1. Reconstructing objects in the client must run all of Taurus's object creation code. Examples: https://github.com/CitrineInformatics/taurus/blob/develop/taurus/entity/object/measurement_run.py#L74 (soft links) https://github.com/CitrineInformatics/taurus/blob/develop/taurus/entity/object/has_conditions.py#L20 (list type validation)
2. The client de/ser code was written assuming that the serialized JSON returned by the backend would match the underlying data model

asantas93: My point is that `loads(dumps(...))` is in principle just a very expensive identity function. It may be (and seems like) the case that we are using this function with overloaded argument types, which makes this the most concise way to get the properties you pointed out, but JSON serialization and deserialization, and `deep_copy` are all pretty expensive operations. I think if we are more deliberate about how we use our serialization functions, we can get away with much less work here. We may also need to have different `build` methods if sometimes the input is a dict of primitives and other times it contains complex types.

bfolie: But it's not an identity--it modifies `loads(dumps(foo))` modifies `foo` from a dictionary of dictionaries to a top-level object, and `.as_dict()` then turns it into a dictionary of objects.

asantas93: I understand that it is not literally the identity function in this case. With the typical `json.loads|dumps` though, it would be. The logic we are actually interested in here is what you just said, and not (de)serialization. I think we should have (or use the existing?) more specific functions to do this work, rather than rely on the side effects of these functions, which both alter object structure and convert to/from JSON string.

src/citrine/resources/data_concepts.py
```diff
@@ -51,9 +56,10 @@ class DataConcepts(PolymorphicSerializable['DataConcepts']):
     Only populated if the :func:`get_type` method is invoked.
     """
 
-    def __init__(self, typ: str):
+    def __init__(self, typ: str, audit_info: Optional[Dict[str, Any]] = None):
```
bfolie: Right now `audit_info` is just a dictionary. It's up to the user to know that it might have keys `created_by`, etc. Should we create an actual `AuditInfo` class?

asantas93: I think that's the way to go, yeah.

src/citrine/resources/data_concepts.py
```diff
@@ -79,56 +85,68 @@ def build(cls, data: dict, session: Session = None):
             An object corresponding to a data concepts resource.
 
         """
+        data_copy = deepcopy(data)
+        # Extract the values that are only for the client.
+        # They will be attached to the deserialized object later.
+        client_only_dict = dict()
+        for client_key in cls._client_keys:
+            val = DataConcepts._pop_field(data_copy, client_key)
```
asantas93: Why did you have to define `_pop_field`? The type hint on this method suggests (and I would expect) that `data` is a `dict`.

asantas93: Also is there a reason you're using `DataConcepts` instead of `cls` in this method? AFAIK the difference between the two is that `DataConcepts.foo` ignores any overrides to `foo` in a child class.

src/citrine/resources/data_concepts.py
bfolie: I've added the AuditInfo class, and the field `audit_info` now comes in through the constructor. I wasn't sure if this called for a leading underscore or not, but since we want it to be read-only, I think it does.

src/citrine/resources/data_concepts.py
bfolie: In the recursive dance that occurs when building an object with soft links, `build` can get called with `data` as an object. Hence the need for `_pop_field`

bfolie: I used `DataConcepts` because I intended `_pop_field` to be a static method that doesn't get overridden, so figured `DataConcepts` and `cls` both work equally well, but `DataConcepts` tells the reader where to find the method.

src/citrine/resources/data_concepts.py
bfolie: > Maybe you can add a new class extending DataConcepts which has all this logic but is only extended by Data Objects and Templates
That might be a good idea, but at the moment the only thing it would get you is the `_client_keys` field. I don't think that on its own is worth the complexity of adding a new class.

asantas93: Well it also keeps `audit_info` out of other classes which that data doesn't apply to. Doesn't leaving it here mean `Property`s will also have an `audit_info` field?
> I've added the AuditInfo class, and the field audit_info now comes in through the constructor.
This is more functional but I actually preferred your old `setattr` approach since having it in the constructor sort of suggests that the user is allowed to set `audit_info` to whatever value they like.

src/citrine/resources/audit_info.py
```diff
@@ -0,0 +1,49 @@
...
+    created_at: int
+        Time, in ms since epoch, at which the object was created
+    updated_by: Optional[str]
+        ID of the user who most recently updated the object
+    updated_at: Optional[int]
+        Time, in ms since epoch, at which the object was most recently updated
+
+    """
+
+    def __init__(self, created_by: str, created_at: int,
```
asantas93: Can you use rich types here (UUID and dates)? Something similar to https://github.com/CitrineInformatics/citrine-python/blob/develop/src/citrine/resources/dataset.py#L73 maybe?

bfolie: Can we assume that `created_by` and `updated_by` will always be uuid4? If so, then yes, I think this is a good idea

asantas93: Yeah they will always be those types

src/citrine/resources/data_concepts.py
bfolie: Those are both good points. The annoying thing about the `setattr` approach was that it did a blind set, so `self.audit_info` became a dictionary instead of an object. I think the right thing to do is to use setters, so this `Storable` object will create an `AuditInfo` object when you pass it  dictionary with `created_by`, `created_at`, etc.

src/citrine/resources/data_concepts.py
```diff
@@ -79,56 +84,68 @@ def build(cls, data: dict, session: Session = None):
...
+    def _pop_field(data, field: str):
         """
-        Get the value of a field from something that might be a dictionary or might be an object.
+        Pop the value of a field from something that might be a dictionary or might be an object.
 
         Parameters
         ----------
         data: dict or object
-            foo
+            From which to pop the value of hte field
```
bfolie:
```suggestion
            From which to pop the value of the field
```

src/citrine/resources/measurement_run.py
```diff
@@ -82,6 +83,7 @@ def __init__(self,
                  file_links: Optional[List[FileLink]] = None,
                  source: Optional[PerformedSource] = None):
         DataConcepts.__init__(self, TaurusMeasurementRun.typ)
+        self._audit_info = None
```
bfolie:
```suggestion
```

src/citrine/resources/audit_info.py
bfolie: Changed to be UUID and Datetime

src/citrine/resources/data_concepts.py
bfolie: I made a `Storable` class that goes in between `DataConcepts` and all of the non-Attributes, mimicking the datasvc structure. This class has a setter for `audit_info` that can accept an `AuditInfo` class or a serialized dict version. `AuditInfo` is now a `Serializable` class and its fields are `Properties`.

src/citrine/resources/audit_info.py
```diff
@@ -0,0 +1,54 @@
...
+        create_str = 'Created by user {} at time {}'.format(
+            self.created_by, self.created_at)
+        if self.updated_by is not None or self.updated_at is not None:
+            update_str = '\nUpdated by user {} at time {}'.format(
+                self.updated_by, self.updated_at)
+        else:
+            update_str = ''
+        return create_str + update_str
+
+    def __eq__(self, other):
```
asantas93: typically you should implement `__hash__` whenever you implement `__eq__`.

src/citrine/resources/process_template.py
```diff
@@ -89,7 +90,7 @@ def __init__(self,
                                        allowed_names=allowed_names)
 
     @classmethod
-    def _build_child_objects(cls, data: dict, session: Session = None):
+    def _build_child_objects(cls, data: dict, data_with_soft_links, session: Session = None):
```
asantas93: Surprised this was working before without this argument. Do you know what happened?

src/citrine/resources/audit_info.py
asantas93: actually I forgot python doesn't automatically make everything hashable -- this doesn't matter unless we want to allow using `AuditInfo` in sets / as dict keys...which doesn't really make sense

asantas93: or doesn't seem likely to be necessary at least

src/citrine/resources/process_template.py
bfolie: The actual implementations of `_build_child_objects` don't use `data_with_soft_links` or `session`. The data with soft links was therefore getting passed as `session`, which went unnoticed because Python doesn't do type checking and the argument wasn't used.

PR #136, opened 2019-12-04
PLA-1615 Add cursor-based request method

asantas93:
# Citrine Python PR

## Description 
At this point (pre- Product API passthrough and full implementation in backend) this PR is mainly just to demonstrate the complexity level of cursor-paging with the current interface. If this seems too annoying to implement for each client, the backend could return simple prev/next links instead of making the client explain how to get the next page (we would lose some flexibility but not a lot).

**EDIT**: The product API passthroughs are live and these methods have been tested against them. Would like to merge for demo.

### PR Type:

### DIFF
--- a/src/citrine/_session.py
+++ b/src/citrine/_session.py
@@ -41,7 +41,7 @@ class Session(requests.Session):
         # Following scheme:[//authority]path[?query][#fragment] (https://en.wikipedia.org/wiki/URL)
         self.headers.update({"Content-Type": "application/json"})
 
-    def versioned_base_url(self, version: str = 'v1'):
+    def _versioned_base_url(self, version: str = 'v1'):
         return '{}://{}/api/{}/'.format(self.scheme, self.authority, version)
 
     def _is_access_token_expired(self):
@@ -50,7 +50,8 @@ class Session(requests.Session):
     def _refresh_access_token(self) -> None:
         """Optionally refresh our access token (if the previous one is about to expire)."""
         data = {'refresh_token': self.refresh_token}
-        response = super().request('POST', self.versioned_base_url() + 'tokens/refresh', json=data)
+        response = super().request(
+            'POST', self._versioned_base_url() + 'tokens/refresh', json=data)
         if response.status_code != 200:
             raise UnauthorizedRefreshToken()
         self.access_token = response.json()['access_token']
@@ -58,11 +59,12 @@ class Session(requests.Session):
             jwt.decode(self.access_token, verify=False)['exp']
         )
 
-    def checked_request(self, method: str, path: str, *args, version: str = 'v1', **kwargs) -> requests.Response:
+    def checked_request(self, method: str, path: str, *args,
+                        version: str = 'v1', **kwargs) -> requests.Response:
         """Check response status code and throw an exception if relevant."""
         if self._is_access_token_expired():
             self._refresh_access_token()
-        uri = self.versioned_base_url(version) + path.lstrip('/')
+        uri = self._versioned_base_url(version) + path.lstrip('/')
         response = super().request(method, uri, *args, **kwargs)
 
         try:
--- a/src/citrine/_session.py
+++ b/src/citrine/_session.py
@@ -131,9 +131,11 @@ class Session(requests.Session):
     @staticmethod
     def cursor_paged_resource(base_method: Callable[..., dict], path: str, *args,
                               forward: bool = True, per_page: int = 100, **kwargs):
-        """Returns a flat generator of results for an API query.
+        """
+        Returns a flat generator of results for an API query.
 
-        Results are fetched in chunks of size `per_page` and loaded lazily."""
+        Results are fetched in chunks of size `per_page` and loaded lazily.
+        """
         params = kwargs.get('params', {})
         params['forward'] = forward
         params['ascending'] = forward
--- a/src/citrine/_session.py
+++ b/src/citrine/_session.py
@@ -1,5 +1,5 @@
 from os import environ
-from typing import Optional, Callable
+from typing import Optional, Callable, Iterator
 from logging import getLogger
 from datetime import datetime, timedelta
 
@@ -39,8 +39,10 @@ class Session(requests.Session):
         self.access_token_expiration: datetime = datetime.utcnow()
 
         # Following scheme:[//authority]path[?query][#fragment] (https://en.wikipedia.org/wiki/URL)
-        self.base_url = '{}://{}/api/v1/'.format(self.scheme, self.authority)
         self.headers.update({"Content-Type": "application/json"})
+
+    def versioned_base_url(self, version: str = 'v1'):
+        return '{}://{}/api/{}/'.format(self.scheme, self.authority, version)
 
     def _is_access_token_expired(self):
         return self.access_token_expiration - EXPIRATION_BUFFER_MILLIS <= datetime.utcnow()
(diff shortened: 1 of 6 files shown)


tests/test_session.py
```diff
@@ -183,3 +183,23 @@ def test_failed_put_with_stacktrace(session: Session):
             session.put_resource('/bad-endpoint', json={})
 
     assert '{"debug_stacktrace": "blew up!"}' == str(e.value)
+
+
+def test_cursor_paged_resource():
+    full_result_set = list(range(26))
+
+    def fake_request(*_, params=None, **__):
```
asantas93: This method mimics the backend's response...if integers were data objects.

src/citrine/_session.py
```diff
@@ -128,6 +128,28 @@ def delete_resource(self, path: str) -> dict:
         """DELETE a particular resource as JSON."""
         return self.checked_delete(path).json()
 
+    @staticmethod
+    def cursor_paged_resource(base_method: Callable[..., dict], path: str, *args,
```
asantas93: `base_method` would be `checked_get` in `filter_by_name`, for example.

asantas93:
```python
params = {'dataset_id': str(self.dataset_id), 'name': name, 'exact': exact}
path = self._get_path(ignore_dataset=True) + "/filter-by-name"
cursor_paged_resource(self._session.checked_get, path, params = params)
```
would be the minimal usage for that method.

src/citrine/_session.py
```diff
@@ -128,6 +128,28 @@ def delete_resource(self, path: str) -> dict:
...
+        params = kwargs.get('params', {})
+        params['forward'] = forward
+        params['ascending'] = forward
+        params['per_page'] = per_page
+        kwargs['params'] = params
+        while True:
+            response_json = base_method(path, *args, **kwargs)
+            for obj in response_json['contents']:
+                yield obj
+            cursor = response_json.get('next')
```
bfolie: What will `cursor` be in general? Is it always an integer?

asantas93: It's actually an arbitrary string which the client doesn't need to understand. Integers are just easy to use in the test. There are a lot of pieces here, which I will link, but you may be better off just reading about general cursor-base paging on the internet. I can find you a good article if you want.

asantas93: I've tried to made some decent docs here. If you're up for reading some scaladocs and source, here are some relevant classes to look at:

General interface for query params:
https://github.com/CitrineInformatics/platform-backend/blob/develop/code/rest/akka-http/src/main/scala/io/citrine/rest/akka/http/modules/PagingSupport.scala
The resulting information about the page to fetch:
https://github.com/CitrineInformatics/platform-backend/blob/develop/code/rest/rest-base/src/main/scala/io/citrine/rest/model/Page.scala
The cursor, itself (not much here):
https://github.com/CitrineInformatics/platform-backend/blob/develop/code/rest/rest-base/src/main/scala/io/citrine/rest/model/Cursor.scala
Example of interpreting a cursor:
https://github.com/CitrineInformatics/platform-backend/blob/develop/code/faraday/dc-dynamo/src/main/scala/io/citrine/faraday/dynamo/DynamoCursor.scala
The results returned:
https://github.com/CitrineInformatics/platform-backend/blob/develop/code/rest/rest-base/src/main/scala/io/citrine/rest/model/PagedResult.scala
Bonus points for reading the code that actually uses the dynamo cursor:
https://github.com/CitrineInformatics/platform-backend/blob/develop/code/faraday/dc-dynamo/src/main/scala/io/citrine/faraday/dynamo/service/DynamoReadDataService.scala#L470

src/citrine/_session.py
```diff
@@ -128,6 +128,28 @@ def delete_resource(self, path: str) -> dict:
...
+    def cursor_paged_resource(base_method: Callable[..., dict], path: str, *args,
+                              forward: bool = True, per_page: int = 100, **kwargs):
+        """
+        Returns a flat generator of results for an API query.
+
+        Results are fetched in chunks of size `per_page` and loaded lazily.
+        """
+        params = kwargs.get('params', {})
+        params['forward'] = forward
+        params['ascending'] = forward
```
bfolie: What's the difference between `forward` and `ascending`?

src/citrine/_session.py
bfolie: An arbitrary string seems fine in this case; it doesn't need to be human-readable.

Thanks for the links. I've started looking at them for my own edification, but I don't think it's necessary for me to review this PR. I was just trying to get a sense of what the client needs to know. I don't think the other method you mentioned (backend returns prev/next links) is conceptually much simpler. Instead of extracting the `cursor` from `response_json` and feeding that into the `params` of the subsequent call, you'd instead have to extract the link to prev/next from (presumably) `response_json` and feed that into the subsequent call somehow. The actual code might be slightly shorter, but I think that the mental energy from deciding "I'd like to implement pagination in this client" to having it functioning would be more or less equal.

I might be misunderstanding how the alternative pagination method would work, but for now I see no reason not to implement it this way.

src/citrine/_session.py
asantas93: https://github.com/CitrineInformatics/platform-backend/blob/develop/code/rest/rest-base/src/main/scala/io/citrine/rest/model/Page.scala talks about this -- the gist is that `ascending` is an absolute sort order for the results, and forward tells the backend whether to return to results "ahead of" or behind the cursor, i.e. for a page size of 2 and a cursor pointing at 5 you would get `3, 4` for `forward = false` and `ascending = true`, `4, 3` for `forward = false` and `ascending = true`, `6, 7` for `forward = true` and `ascending = true`. If you never reverse directions `forward` will always equal `ascending`.

bfolie: Thanks. Sounds like it's fine for the client to always set `params['forward']` and `params['ascending']` to be the same, and that this ability to reverse directions might be more important for a graphical interface.

asantas93: I would say that's accurate, yeah.

tests/resources/test_material_run.py
```diff
@@ -164,6 +165,30 @@ def test_filter_by_name(collection, session):
     assert sample_run['uids'] == runs[0].uids
 
 
+def test_list_by_name(collection, session):
+    all_runs = [
+        MaterialRunDataFactory(name="foo_{}".format(i)) for i in range(20)
+    ]
+    fake_request = make_fake_cursor_request_function(all_runs)
+    # lotta shady stuff going on in this test
```
asantas93: Both of these methods _do_ actually work when run against develop (in AWS) from my computer.

src/citrine/resources/data_concepts.py
```diff
@@ -632,10 +637,81 @@ def filter_by_name(self, name: str, exact: bool = False,
...
+        exact: bool
+            Set to True to change prefix search to exact search (but still case-insensitive).
+            Default is False.
+        forward: bool
+            Set to False to reverse the order of results (i.e. return in descending order).
+        per_page: int
+            Controls how many results are fetched per request. Has no impact on the size of
+            the result set but will have some effect on performance. Lower gets results
+            faster but increases the total number of requests needed to fetch the entire
+            result set, which increases total latency across fetching the entire result set.
```
djack: Since this is a public docstring, less definitive language about performance, and more helpful language may be motivated, here. 

"Controls the number of results fetched with each http request to the backend. Typically, this is set to a sensible default and should not be modified. Consider modifying this value only if you find this method is unacceptably latent."

src/citrine/resources/data_concepts.py
```diff
@@ -632,10 +637,81 @@ def filter_by_name(self, name: str, exact: bool = False,
...
+
+        Parameters
+        ----------
+        forward: bool
+            Set to False to reverse the order of results (i.e. return in descending order).
+        per_page: int
+            Controls how many results are fetched per request. Has no impact on the size of
+            the result set but will have some effect on performance. Lower gets results
+            faster but increases the total number of requests needed to fetch the entire
+            result set, which increases total latency across fetching the entire result set.
```
djack: "Controls the number of results fetched with each http request to the backend. Typically, this is set to a sensible default and should not be modified. Consider modifying this value only if you find this method is unacceptably latent."

PR #138, opened 2019-12-05
Wired in s3 overrides so we can chat to localstack

dleblanc:
This wires in some s3 overrides such that we can target localstack for
file uploads instead of the default Amazon S3 endpoint.

# Citrine Python PR

## Description 

### PR Type:

### DIFF
--- a/src/citrine/resources/file_link.py
+++ b/src/citrine/resources/file_link.py
@@ -4,6 +4,7 @@ import os
 import mimetypes
 from typing import Iterable, Optional
 from boto3 import client as boto3_client
+from boto3.session import Config
 import requests
 from botocore.exceptions import ClientError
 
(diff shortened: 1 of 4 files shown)


src/citrine/resources/file_link.py
```diff
@@ -251,11 +259,21 @@ def _upload_file(file_path: str, uploader: _Uploader):
             The input uploader object with its s3_version field now populated.
 
         """
+
+        additional_s3_opts = {
+            'use_ssl': uploader.s3_use_ssl,
+            'config': Config(s3={'addressing_style': uploader.s3_addressing_style})
+        }
+
+        if uploader.s3_endpoint_url:
```
bfolie: Are you looking for `s3_endpoint_url` to be `True` or a string or simply not None?

dleblanc: Oh good point - this should be:

if ... != None:

dleblanc:
Working on adding some tests to cover this stuff - will remove the WIP tag when it's complete.

PR #145, opened 2019-12-12
Add cursor-paged tag/attribute search methods

asantas93:
# Citrine Python PR

## Description 
Add cursor-paged `list_by_tag` and `list_by_attribute_bounds` methods.

This PR should not be merged before the corresponding backend implementation is merged as well.

### PR Type:

### DIFF
--- a/src/citrine/resources/data_concepts.py
+++ b/src/citrine/resources/data_concepts.py
@@ -384,7 +384,7 @@ class DataConceptsCollection(Collection[ResourceType]):
     dataset_id: UUID
         The uid of the dataset that this collection belongs to. If None then the collection
         ranges over all datasets in the project. Note that this is only allowed for certain
-        actions. For example, you can use :func:`filter_by_tags` to search over all datasets,
+        actions. For example, you can use :func:`list_by_tag` to search over all datasets,
         but when using :func:`register` to upload or update an object, a dataset must be specified.
     session: Session
         The Citrine session used to connect to the database.
@@ -510,6 +510,7 @@ class DataConceptsCollection(Collection[ResourceType]):
         data = self.session.get_resource(path)
         return self.build(data)
 
+    @deprecated(details='please use list_by_tag')
     def filter_by_tags(self, tags: List[str],
                        page: Optional[int] = None, per_page: Optional[int] = None):
         """
@@ -549,6 +550,7 @@ class DataConceptsCollection(Collection[ResourceType]):
             params=params)
         return [self.build(content) for content in response["contents"]]
 
+    @deprecated(details='please use list_by_attribute_bounds')
     def filter_by_attribute_bounds(
             self,
             attribute_bounds: Dict[Union[AttributeTemplate, LinkByUID], BaseBounds],
@@ -583,13 +585,7 @@ class DataConceptsCollection(Collection[ResourceType]):
             and have values within the specified bounds.
 
         """
-        assert isinstance(attribute_bounds, dict) and len(attribute_bounds) == 1
-
-        attribute_bounds_dict = dict()
-        for key, value in attribute_bounds.items():
-            template_id = get_object_id(key)
-            attribute_bounds_dict[template_id] = value.as_dict()
-        body = {'attribute_bounds': attribute_bounds_dict}
+        body = self._get_attribute_bounds_search_body(attribute_bounds)
         params = {}
         if self.dataset_id is not None:
             params['dataset_id'] = str(self.dataset_id)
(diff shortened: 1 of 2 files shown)


asantas93:
For reference, this is the function that actually understands cursors
https://github.com/CitrineInformatics/citrine-python/blob/develop/src/citrine/_session.py#L140-L160

src/citrine/resources/data_concepts.py
```diff
@@ -725,3 +809,19 @@ def delete(self, uid: Union[UUID, str], scope: str = 'id'):
...
         self.session.delete_resource(path)
         return Response(status_code=200)  # delete succeeded
+
+    @staticmethod
+    def _get_attribute_bounds_search_body(attribute_bounds):
+        if not isinstance(attribute_bounds, dict):
+            raise TypeError('attribute_bounds must be a dict mapping template to bounds; '
+                            'got {}'.format(attribute_bounds))
+        if len(attribute_bounds) != 1:
+            raise NotImplementedError('Currently, once searches with exactly one template '
```
bfolie:
```suggestion
            raise NotImplementedError('Currently, only searches with exactly one template '
```

src/citrine/resources/data_concepts.py
```diff
@@ -725,3 +809,19 @@ def delete(self, uid: Union[UUID, str], scope: str = 'id'):
...
+                                      'to bounds mapping is supported; got {}'
```
bfolie:
```suggestion
                                      'to bounds mapping are supported; got {}'
```

src/citrine/resources/data_concepts.py
```diff
@@ -677,6 +673,58 @@ def list_by_name(self, name: str, exact: bool = False,
...
+            A dictionary from attributes to the bounds on that attribute.
+            Currently only real and integer bounds are supported.
+            Each attribute may be represented as an AttributeTemplate (PropertyTemplate,
+            ParameterTemplate, or ConditionTemplate) or as a LinkByUID,
+            but in either case there must be a uid and it must correspond to an
+            AttributeTemplate that exists in the database.
+            Only the uid is passed, so if you would like to update an attribute template you
+            must register that change to the database before you can use it to filter.
+        forward: bool
+            Set to False to reverse the order of results (i.e. return in descending order).
```
bfolie: This method allow the user to specify `forward` or `reverse`, but `list_by_tag` does not

asantas93: That's intentional -- ascending/descending order to the results of this function is actually meaningful, but in `list_by_tag`, which is ordered by UUIDS, I can think of no reason to support it. I think it's best to omit it until people think of a use case. That leaves more flexibility in the backend.

tests/resources/test_material_run.py
```diff
@@ -165,31 +165,38 @@ def test_filter_by_name(collection, session):
...
     setattr(session, 'get_resource', fake_request)
+    setattr(session, 'post_resource', fake_request)
     setattr(session, 'cursor_paged_resource', Session.cursor_paged_resource)
-    # TODO: the results appear equal in the diff but fail equality check
+
     assert len(list(collection.list_by_name('unused', per_page=2))) == len(all_runs)
+    assert len(list(collection.list_all(per_page=2))) == len(all_runs)
+    assert len(list(collection.list_by_tag('unused', per_page=2))) == len(all_runs)
+    assert len(list(collection.list_by_attribute_bounds(
+        {LinkByUIDFactory(): IntegerBounds(1, 5)}, per_page=2))) == len(all_runs)
```
bfolie: Is this just testing that all of these methods (list by name, list all, list by tag, and list by attribute bounds) call get_resource or post_resource? Or is it testing something else?

asantas93: Pretty much. It ensures that they perform a valid invocation of `cursor_paged_resource`, which is not a **total** no-op test to do. I will admit that this is to a significant degree an 'appeasing 100% test coverage' test. The real testing will happen in nextgen-devkit.

asantas93: The test of `cursor_paged_resource` is more thorough: https://github.com/CitrineInformatics/citrine-python/blob/develop/tests/test_session.py#L189

asantas93: Well technically these tests call the same functions, actually. Let me say this: I think `cursor_paged_resource` is decently well tested, and these search functions don't really do anything except call `cursor_paged_resource`.

asantas93: Also gating 100% test coverage is dumb #hottakes

src/citrine/resources/data_concepts.py
bfolie: That makes perfect sense. I actually thought that both methods returned objects in a meaningless order, and was going to advocate for removing the argument in both cases, but I agree it should be used here.

tests/resources/test_material_run.py
bfolie: 😱
I think there's value in at least asserting that `list_by_tag` calls `cursor_paged_resource`. You could also check that it does so with the correct arguments.

asantas93: I actually think asserting that a particular function is called does more harm than good unless you're trying to test something around performance (e.g. request is only made once because it gets cached). I think that kind of testing is brittle and hard to maintain. With simple functions like these, in which you just call another function with arguments derived from user input and instance variables, tests devolve into 'test that this function is written the way I wrote it', which fails every time you change anything.
FWIW though, these tests do implicitly check that `cursor_paged_resource` is called, because the `len` assertions would not pass were the function not calling `cursor_paged_resource`.

PR #147, opened 2019-12-19
Add debug logging to checked_request

asantas93:
# Citrine Python PR

## Description 
Adds debug logging to checked request. Should help us troubleshoot potential client-side serialization issues (and hopefully other problems too).

Also removed `*args` from request passthrough methods. All args to `request` other than `method` and `url` are named and have default values, which makes position arguments a poor choice for specifying them. `kwargs` enables their usage in a more stable and readable way. `request` signature for context:
```
    def request(self, method, url,
            params=None, data=None, headers=None, cookies=None, files=None,
            auth=None, timeout=None, allow_redirects=True, proxies=None,
            hooks=None, stream=None, verify=None, cert=None, json=None):
```

### PR Type:

### DIFF
--- a/src/citrine/_session.py
+++ b/src/citrine/_session.py
@@ -63,18 +63,18 @@ class Session(requests.Session):
             jwt.decode(self.access_token, verify=False)['exp']
         )
 
-    def checked_request(self, method: str, path: str, *args,
+    def checked_request(self, method: str, path: str,
                         version: str = 'v1', **kwargs) -> requests.Response:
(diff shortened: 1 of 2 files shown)


asantas93:
> Good decision to remove `*args`.
> 
> I think it would be helpful to have some documentation as to the different logging levels and how to toggle between them, perhaps in the top-level README. It might be appropriate for a separate PR, but since this PR expands the use of logging, it feels thematically related.

Absolutely agree. I will add that.

PR #149, opened 2019-12-19
Logging standardization and documentation

asantas93:
# Citrine Python PR

## Description 

* Moves all loggers to module level for ease of location for suppression
* Removes unused loggers
* Sets default log level to `INFO` and formats output to include logger module
* Explains logging levels in README (pretty: https://github.com/CitrineInformatics/citrine-python/tree/feature/logging-docs)

### PR Type:

### DIFF
--- a/src/citrine/__init__.py
+++ b/src/citrine/__init__.py
@@ -1,2 +1,5 @@
 # TODO: Add a docstring here
 from citrine.citrine import Citrine  # noqa: F401
+import logging
+
+logging.basicConfig(level=logging.INFO)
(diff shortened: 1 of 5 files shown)


src/citrine/_session.py
```diff
@@ -30,7 +31,6 @@ def __init__(self,
                  host: str = 'citrine.io',
                  port: Optional[str] = None):
         super().__init__()
-        self.logger = getLogger(__name__)
```
asantas93: Instance variables are harder for the user to locate.

src/citrine/__init__.py
```diff
@@ -1,2 +1,5 @@
 # TODO: Add a docstring here
 from citrine.citrine import Citrine  # noqa: F401
+import logging
+
+logging.basicConfig(level=logging.INFO)
```
asantas93: This establishes `INFO` as the default log level. A somewhat perplexing side-effect is the inclusion of the module name in the output, which is actually very nice for locating the logger. We want want to add more deliberate configuration here down the road.

PR #154, opened 2019-12-20
add created_by and create_time to workflows

kyle-laplante:
# Citrine Python PR

## Description 

### PR Type:

### DIFF
--- a/src/citrine/informatics/workflows.py
+++ b/src/citrine/informatics/workflows.py
@@ -67,6 +67,8 @@ class DesignWorkflow(Resource['DesignWorkflow'], Workflow):
         serializable=False
     )
     active = properties.Boolean('active', default=True)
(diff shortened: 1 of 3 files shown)


src/citrine/informatics/workflows.py
```diff
@@ -67,6 +67,8 @@ class DesignWorkflow(Resource['DesignWorkflow'], Workflow):
         serializable=False
     )
     active = properties.Boolean('active', default=True)
+    created_by = properties.UUID('created_by', serializable=False)
```
jtharris: Is this field useful?  I guess we can fetch users by UUID as well?

kyle-laplante: im not too sure what the python user would want it for but we expose it on the API for the frontend (explicit requirement for LH2 scope) so i thought it would be good to just keep the python data up to par with the API content.

PR #158, opened 2020-01-03
List of ingredients gets jumbled during material history building

bfolie:
# Citrine Python PR

## Description 
Fixes a bug in which building a material history (converting from taurus objects to citrine-python objects) jumbled the list of ingredients in a process. The cause was as follows: the taurus process has a list of taurus ingredients, and we iterate over that list, building each ingredient as a citrine-python object and attaching it to the citrine-python version of the process. This action temporarily severs the link between the taurus process and ingredient, and when they are re-connected, the list is permuted (the newly-attached object is added to the end of the list). We were iterating over the list as it was being modified. The fix is to create a copy of the list of ingredients and iterate over that.

This work surfaced another bug, which is that the `build` method creates "dangling links." Given a material, `mat`, `mat.spec.process` points to a process spec, and `mat.process.spec` points to a _different version_ of that process spec, one that is not connected to the material spec. Therefore, `mat.spec.process.output_material` is `None`, when it should equal `mat.spec`. I spent most of the day trying to close that loop and was able to make it work, but could not fix the comparable problem for the ingredient/process run/spec square. 

I believe that the `build` method as it is currently structured cannot solve this problem. It attempts to traverse the tree of Taurus objects, without explicitly keeping track of where it has been. I had hoped that we could leverage the fact that the Taurus objects were already connected correctly to simplify the `build()` logic, but at this point it has become more complicated and is working less well, so I think we should refactor the way material histories are built. One option is to essentially copy Taurus's deserialization code (keep a list of all objects, substitute links for objects one by one), but for citrine-python objects.

I think that the question of how to refactor `build()` is separate from this PR, which is narrowly scoped to fixing PLA-2266.

### PR Type:

### DIFF
--- a/src/citrine/resources/data_concepts.py
+++ b/src/citrine/resources/data_concepts.py
@@ -291,6 +291,11 @@ class DataConcepts(PolymorphicSerializable['DataConcepts']):
         if linked_objects is None:
             return
 
+        # Make `linked_objects` point to the list of linked objects themselves, as opposed to the
+        # attribute of `obj_with_soft_links`. The list in `obj_with_soft_links` will be modified
+        # in the loop below, and we don't want to modify a list while iterating over it.
+        linked_objects = [l for l in linked_objects]
+
         # Cycle through linked objects in obj_with_soft_link and if they are not LinkByUID,
         # build them and then set their `reverse_field` field to dc_obj
         for linked_obj in linked_objects:
--- a/src/citrine/resources/data_concepts.py
+++ b/src/citrine/resources/data_concepts.py
@@ -1,7 +1,7 @@
 """Top-level class for all data concepts objects and collections thereof."""
 from uuid import UUID
 from typing import TypeVar, Type, List, Dict, Union, Optional, Iterator
-from copy import deepcopy
+from copy import copy, deepcopy
 from abc import abstractmethod
 
 from citrine._session import Session
@@ -291,10 +291,10 @@ class DataConcepts(PolymorphicSerializable['DataConcepts']):
         if linked_objects is None:
             return
 
-        # Make `linked_objects` point to the list of linked objects themselves, as opposed to the
+        # Make `linked_objects` point to a copy of the list of linked objects, as opposed to the
         # attribute of `obj_with_soft_links`. The list in `obj_with_soft_links` will be modified
         # in the loop below, and we don't want to modify a list while iterating over it.
-        linked_objects = [l for l in linked_objects]
+        linked_objects = copy(linked_objects)
 
         # Cycle through linked objects in obj_with_soft_link and if they are not LinkByUID,
         # build them and then set their `reverse_field` field to dc_obj
--- a/tests/serialization/test_material_run.py
+++ b/tests/serialization/test_material_run.py
@@ -145,41 +145,16 @@ def test_measurement_material_connection_rehydration():
 
 
 def test_cake():
-    """Test that the cake example from taurus can be built without modification."""
-    def _check_equivalence(obj1, obj2, seen=None):
-        """A convenience method to check if two objects are equivalent.
-
-        obj2 is allowed to be "bigger" than obj1: if they are DictSerializable, obj2 may have more
-        attributes, if they are lists, obj2 may be longer, if they are dicts, obj2 may have more
-        key-value pairs. The important thing is that obj2 have all the attributes of obj1.
-        The intended use is to compare a taurus object (obj1) with a citrine-python object (obj2).
-        The citrine-pythong object can have additional meta-information related to its possible
-        connection with an external database.
-        """
-        if seen is None:
-            seen = set()
-        if obj1.__hash__ is not None:
-            if obj1 in seen:
-                return True
-            else:
-                seen.add(obj1)
-
-        if isinstance(obj1, (list, tuple)):
-            return isinstance(obj2, (list, tuple)) and \
-                   all([_check_equivalence(obj1[i], obj2[i], seen) for i in range(len(obj1))])
-        elif isinstance(obj1, dict):
-            return isinstance(obj2, dict) and \
-                   all([_check_equivalence(obj1.get(key), obj2.get(key), seen)
-                        for key in obj1.keys()])
-        elif isinstance(obj1, DictSerializable):
-            keys = {x.lstrip('_') for x in vars(obj1)}
-            return isinstance(obj2, DictSerializable) and \
-                all([_check_equivalence(getattr(obj1, key), getattr(obj2, key, None), seen)
-                     for key in keys])
-        else:
-            return obj1 == obj2
-
+    """Test that the cake example from taurus can be built without modification.
+    This only tests the fix to a limited problem (not all ingredients being reconstructed) and
+    is not a full test of equivalence, because the reconstruction creates "dangling paths."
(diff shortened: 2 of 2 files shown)


src/citrine/resources/data_concepts.py
```diff
@@ -291,6 +291,11 @@ def _build_list_of_soft_links(dc_obj, obj_with_soft_links, field: str, reverse_f
         if linked_objects is None:
             return
 
+        # Make `linked_objects` point to the list of linked objects themselves, as opposed to the
+        # attribute of `obj_with_soft_links`. The list in `obj_with_soft_links` will be modified
+        # in the loop below, and we don't want to modify a list while iterating over it.
+        linked_objects = [l for l in linked_objects]
```
maxhutch: Is there a reason to copy with this list comprehension instead of python's [copy module](https://docs.python.org/2/library/copy.html) (as used [in taurus](https://github.com/CitrineInformatics/taurus/blob/develop/taurus/util/impl.py#L113))?

Imperssonator:
Thank you @bfolie for tracking this down and for the thorough explanation. Really happy to see this resolved. Also, thanks for explaining the other bug - makes perfect sense. Luckily, as long as data is being uploaded via LinkMaster, there is a 1:1 Run:Spec correspondence, so there should not be a case in the near-term where we need to navigate like `.spec.process` - traversing across a History _via_ Runs and then accessing specs is the more intuitive use-case, IMO. Hopefully this helps inform prioritization.

src/citrine/resources/data_concepts.py
bfolie: No reason not to use `copy` (which I think is equivalent to this list comprehension). I changed it to `copy` for readability.

`deepcopy` does not work, because it is important that `setattr` below act on the actual objects that `obj_with_soft_links` points to, not copies of those objects.

bfolie:
> were the unit test failures stochastic or deterministic? If stochastic, with which rate? Just trying to figure out how thorough the test coverage is.

The problem occurs 100% of the time, but the test as I wrote it failed about 50% of the time. The way that the `build` method works, if you give it a connected square of material run, process run, material spec, and process spec, it will create two copies in memory of the process spec.
Assume that you have the following Taurus objects:
```
PS <---> MS
 /\      /\
 |        |
 |        |
PR <---- MR
```

You wind up with the following citrine-python objects:
```
PS <---> MS
         /\
PS       |
/\       |
 |       |
 |       |
PR <---- MR
```
Whether or not the test to compare the two passed depended on the route it took. From the starting material run, if it traverses up then left, then it will say "this process spec looks fine" but if it traverses left then up it will say "this process spec is missing its `output_material` field, and is not identical to the taurus version."

bfolie:
> At the moment, it's not completely necessary to navigate like `.spec.process` - traversing across a History _via_ Runs and then accessing specs is the more intuitive use-case, IMO.

I agree, but unfortunately traversing via runs and then checking out the spec will result in a bug. Given the example in my above comment, the user would get a process spec object and if they then asked "what's the output material spec of this process spec?" they would get `None` Fortunately they probably won't do that and would probably go straight from the material run to its spec, so I don't think users will feel this bug acutely right away.

maxhutch:
> > were the unit test failures stochastic or deterministic? If stochastic, with which rate? Just trying to figure out how thorough the test coverage is.
> 
> The problem occurs 100% of the time, but the test as I wrote it failed about 50% of the time. 

If we wrap the test in a loop, can we drive that up?  Or do we need to create some randomness in the graph traversal?  I'm just trying to avoid the case where bug is introduced but the tests pass (luckily) and then someone else writes a correct feature and the test starts failing for no obvious reason.

bfolie:
> If we wrap the test in a loop, can we drive that up? Or do we need to create some randomness in the graph traversal? I'm just trying to avoid the case where bug is introduced but the tests pass (luckily) and then someone else writes a correct feature and the test starts failing for no obvious reason.

There is randomness in the graph traversal, which is why it failed half of the time. I removed the test, but you can see it in this intermediate commit: https://github.com/CitrineInformatics/citrine-python/pull/158/commits/770fdb0af2cd20c5444947b4b18a858b546ecf6c#diff-2617a4b3afa881c82a25afd26d2079d2R149

I think that a 50% bug would be caught by our CI process. It's frequent enough that one would recognize the existence of a problem. In theory one could rebuild until they got a passing build, then merge it, but I hope we're wise enough to not do that.

Right now, the build passes because our test coverage is incomplete. It's _not_ passing because we attempt to test for a given behavior, and the test passes stochastically.

maxhutch:
> I think that a 50% bug would be caught by our CI process. It's frequent enough that one would recognize the existence of a problem. In theory one could rebuild until they got a passing build, then merge it, but I hope we're wise enough to not do that.

I don't find this argument convincing, but I'm comfortable deferring to another reviewer.  I'd like to see the false positive rate drop to closer to 1/20, ideally 1/100.

bfolie:
> I don't find this argument convincing, but I'm comfortable deferring to another reviewer. I'd like to see the false positive rate drop to closer to 1/20, ideally 1/100.

To be clear, there are no false positives in this PR. Every test right now passes deterministically. We have a known gap in our testing and an un-tested bug. The way I originally found that bug was through a test that passed stochastically, but I can easily write a test for this bug that would fail 100% of the time. It would be:
```
taurus_cake = make_cake()
cake = MaterialRun.build(taurus_cake)
assert cake.process.spec.output_material is not None
```

maxhutch:
> > I don't find this argument convincing, but I'm comfortable deferring to another reviewer. I'd like to see the false positive rate drop to closer to 1/20, ideally 1/100.
> 
> To be clear, there are no false positives in this PR. Every test right now passes deterministically. We have a known gap in our testing and an un-tested bug, for which there is a Jira ticket. The way I originally found that bug was through a test that passed stochastically, but I can easily write a test for this bug that would fail 100% of the time. It would be:
> 
> ```
> taurus_cake = make_cake()
> cake = MaterialRun.build(taurus_cake)
> assert cake.process.spec.output_material is not None
> ```

That's a different bug, though, right?

I understand that this code passes 100% of the time.  My point is that if we merge it, then I could re-introduce the bug and the test would pass 50% of the time.  If I was clever at covering my tracks, then I could get the bug into `develop` (because the tests happened to pass: false positive) and then the next developer would experience test failures 50% of the time.  If they knew what you now know, then they'd recognize that the test was stochastic and the failure might not have been there's.  But, if they weren't aware of this type of issue (e.g. if they were focused mostly on AI modules) then they might erroneous think they broken something and waste time trying to track down a phantom issue.  This is why I'd claim that non-deterministic failures can be worse (for developers) than missing test coverage: they can cause people to waste a ton of time.

Can we just wrap it in a loop and call it a day?  5 reps gives us ~3% failure rate.  7 reps is < 1%.

bfolie:
> That's a different bug, though, right?

I apologize; this has all been a misunderstanding, induced by me. To clarify, the test introduced in this PR fails 100% of the time without the fix. No stochastic tests are being introduced.

When you asked me if the unit test failures were stochastic or deterministic, you were thinking about the test that is being committed. But I had the other bug on my mind (cake.process.spec.output_material being None), since that's what I spent most of the day working on. I had written a test to check if the two cake histories were 100% identical. That test found the new bug, but only about half of the time. But that information is irrelevant to this PR.

PR #163, opened 2020-01-07
Ara first review

maxhutch:
This review is scoped to a minimal subset of the Ara Definition prototype.  This _should not be merged_.  Rather, when the review is approved, we'll close this PR and make all subsequent contributions to `feature/ara` through separate PRs into that branch.  Then, we'll re-open this PR when Ara is ready for release.

That much being said: this code should be evaluated for quality; this isn't just a design review.  This may be the only detailed review of the code present in this PR.

### DIFF
--- a/src/citrine/ara/columns.py
+++ b/src/citrine/ara/columns.py
@@ -58,7 +58,7 @@ class RealMeanColumn(Serializable['RealMeanColumn'], Column):
     def _attrs(self) -> List[str]:
         return ["data_source", "target_units", "typ"]
 
-    def __init__(self,
+    def __init__(self, *,
                  data_source: str,
                  target_units: Optional[str] = None):
         self.data_source = data_source
@@ -81,6 +81,6 @@ class IdentityColumn(Serializable['IdentityColumn'], Column):
     def _attrs(self) -> List[str]:
         return ["data_source", "typ"]
 
-    def __init__(self,
+    def __init__(self, *,
                  data_source: str):
         self.data_source = data_source
--- a/src/citrine/ara/columns.py
+++ b/src/citrine/ara/columns.py
@@ -1,0 +1,86 @@
+"""Variable definitions for Ara"""
+from typing import Type, Optional, List  # noqa: F401
+from abc import abstractmethod
+
+from taurus.entity.link_by_uid import LinkByUID
+
+from citrine._serialization.serializable import Serializable
+from citrine._serialization.polymorphic_serializable import PolymorphicSerializable
+from citrine._serialization import properties
+
+
+class Column(PolymorphicSerializable['Column']):
+    """A column in the Ara table, defined as some operation on a variable.
+
+    Abstract type that returns the proper type given a serialized dict.
+    """
+
+    @abstractmethod
+    def attrs(self) -> List[str]:
+        pass  # pragma: no cover
+
+    def __eq__(self, other):
+        try:
+            return all([
+                self.__getattribute__(key) == other.__getattribute__(key) for key in self.attrs()
+            ])
+        except AttributeError:
+            return False
+
+    @classmethod
+    def get_type(cls, data) -> Type[Serializable]:
+        """Return the subtype."""
+        types: List[Type[Serializable]] = [
+            RealMeanColumn, IdentityColumn
+        ]
+        return next(x for x in types if x.type == data["type"])
+
+
+class RealMeanColumn(Serializable['RealMeanColumn'], Column):
+    """Column containing the mean of a real-valued variable.
+
+    Parameters
+    ----------
+    data_source: str
+        short_name of the variable to use when populating the column
+    target_units: optional[str]
+        units to convert the real variable into
+
+    """
+
+    data_source = properties.String('data_source')
+    target_units = properties.Optional(properties.String, "target_units")
+    type = properties.String('type', default="real_mean_column", deserializable=False)
+
+    def attrs(self) -> List[str]:
+        return ["data_source", "target_units", "type"]
+
+    def __init__(self,
+                 data_source: str,
+                 target_units: Optional[str] = None):
+        self.data_source = data_source
+        self.target_units = target_units
+
+
+class IdentityColumn(Serializable['IdentityColumn'], Column):
+    """Column containing the value of a string-valued variable.
+
+    Parameters
+    ----------
+    data_source: str
+        short_name of the variable to use when populating the column
+
+    """
+
+    data_source = properties.String('data_source')
+    type = properties.String('type', default="identity_column", deserializable=False)
+
+    def attrs(self) -> List[str]:
+        return ["data_source", "type"]
+
+    def __init__(self,
+                 data_source: str,
+                 target_units: Optional[str] = None):
+        self.data_source = data_source
+        self.target_units = target_units
+
--- a/src/citrine/ara/columns.py
+++ b/src/citrine/ara/columns.py
@@ -45,7 +45,7 @@ class RealMeanColumn(Serializable['RealMeanColumn'], Column):
     Parameters
     ----------
     data_source: str
-        short_name of the variable to use when populating the column
+        name of the variable to use when populating the column
     target_units: optional[str]
         units to convert the real variable into
 
@@ -71,7 +71,7 @@ class IdentityColumn(Serializable['IdentityColumn'], Column):
     Parameters
     ----------
     data_source: str
-        short_name of the variable to use when populating the column
+        name of the variable to use when populating the column
 
     """
 
--- a/src/citrine/ara/columns.py
+++ b/src/citrine/ara/columns.py
@@ -1,8 +1,6 @@
-"""Variable definitions for Ara"""
+"""Column definitions for Ara."""
 from typing import Type, Optional, List  # noqa: F401
 from abc import abstractmethod
-
-from taurus.entity.link_by_uid import LinkByUID
 
 from citrine._serialization.serializable import Serializable
 from citrine._serialization.polymorphic_serializable import PolymorphicSerializable
@@ -16,13 +14,13 @@ class Column(PolymorphicSerializable['Column']):
     """
 
     @abstractmethod
-    def attrs(self) -> List[str]:
+    def _attrs(self) -> List[str]:
         pass  # pragma: no cover
 
     def __eq__(self, other):
         try:
             return all([
-                self.__getattribute__(key) == other.__getattribute__(key) for key in self.attrs()
+                self.__getattribute__(key) == other.__getattribute__(key) for key in self._attrs()
             ])
         except AttributeError:
             return False
@@ -52,7 +50,7 @@ class RealMeanColumn(Serializable['RealMeanColumn'], Column):
     target_units = properties.Optional(properties.String, "target_units")
     type = properties.String('type', default="real_mean_column", deserializable=False)
 
-    def attrs(self) -> List[str]:
+    def _attrs(self) -> List[str]:
         return ["data_source", "target_units", "type"]
 
     def __init__(self,
@@ -75,7 +73,7 @@ class IdentityColumn(Serializable['IdentityColumn'], Column):
     data_source = properties.String('data_source')
     type = properties.String('type', default="identity_column", deserializable=False)
 
-    def attrs(self) -> List[str]:
+    def _attrs(self) -> List[str]:
         return ["data_source", "type"]
 
     def __init__(self,
@@ -83,4 +81,3 @@ class IdentityColumn(Serializable['IdentityColumn'], Column):
                  target_units: Optional[str] = None):
         self.data_source = data_source
         self.target_units = target_units
-
--- a/src/citrine/ara/columns.py
+++ b/src/citrine/ara/columns.py
@@ -82,7 +82,5 @@ class IdentityColumn(Serializable['IdentityColumn'], Column):
         return ["data_source", "typ"]
 
     def __init__(self,
-                 data_source: str,
-                 target_units: Optional[str] = None):
+                 data_source: str):
         self.data_source = data_source
-        self.target_units = target_units
--- a/src/citrine/ara/columns.py
+++ b/src/citrine/ara/columns.py
@@ -28,10 +28,15 @@ class Column(PolymorphicSerializable['Column']):
     @classmethod
     def get_type(cls, data) -> Type[Serializable]:
         """Return the subtype."""
+        if "type" not in data:
+            raise ValueError("Can only get types from dicts with a 'type' key")
         types: List[Type[Serializable]] = [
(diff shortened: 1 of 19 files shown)


tests/resources/test_ara_definition.py
```diff
@@ -0,0 +1,75 @@
...
+    assert retrieved_ara_definition.version == ara_definition["version"]
+
+
+def test_init_ara_definition():
+    ara_definition = AraDefinition(name="foo", description="bar", rows=[], columns=[], variables=[], datasets=[])
+    assert ara_definition.uid is None
+    assert ara_definition.version is None
+
+
+def test_dump_example():
```
maxhutch: @kroenlein this is the test to run if you want to check the serialization.

tests/resources/test_ara_definition.py
```diff
@@ -0,0 +1,75 @@
...
+    )
+    ara_definition = AraDefinition(
+        name="Example Table",
+        description="Illustrative example that's meant to show how Ara Definitions will look serialized",
+        datasets=[uuid4()],
+        rows=[MaterialRunByTemplate(templates=[LinkByUID(scope="templates", id="slices")])],
+        columns=[RealMeanColumn(data_source=density.short_name)],
+        variables=[density]
+    )
+    print(json.dumps(ara_definition.dump(), indent=2))
```
maxhutch: We'll remove this before it gets merged into `develop`

src/citrine/ara/variables.py
```diff
@@ -0,0 +1,98 @@
...
+
+    @abstractmethod
+    def _attrs(self) -> List[str]:
+        pass  # pragma: no cover
+
+    def __eq__(self, other):
+        try:
+            return all([
+                self.__getattribute__(key) == other.__getattribute__(key) for key in self._attrs()
+            ])
```
bfolie: If `other` has matches all of the attributes that `self` has, but has other attributes in addition, then I believe this will erroneously return True.

maxhutch: That's why `type` is included to make sure that the classes being compared are intended to be equivalent.  Note that I lifted this code block from [orion descriptors](https://github.com/CitrineInformatics/citrine-python/blob/develop/src/citrine/informatics/descriptors.py#L46).

src/citrine/ara/variables.py
```diff
@@ -0,0 +1,98 @@
...
+        except AttributeError:
+            return False
+
+    @classmethod
+    def get_type(cls, data) -> Type[Serializable]:
+        """Return the subtype."""
+        types: List[Type[Serializable]] = [
+            RootInfo, AttributeByTemplate
+        ]
+        return next(x for x in types if x.type == data["type"])
```
bfolie: I think this could be more defensive, throwing exceptions if `data` is missing a "type" field or if it doesn't match the `types` list (say, because something new got added in the backend but not in citrine-python).

I looked at the other implementations of `get_type`, and they're all doing basically the same thing, but there are several different styles and levels of error checking. Unrelated to this PR, I think we could require each `PolymorphicSerializable` to have some "types dictionary" and make the `get_type` method more generic. This would also allow for more error checking without having to write tests for that error checking for every single implementation of PolymorphicSerializable.

I can make the issue/ticket and take care of it if you think it's worthwhile.

src/citrine/ara/variables.py
```diff
@@ -0,0 +1,98 @@
...
+        sequence of column headers
+    field: str
+        name of the field to assign the variable to
+
+    """
+
+    short_name = properties.String('short_name')
+    output_name = properties.List(properties.String, 'output_name')
+    field = properties.String('field')
+    type = properties.String('type', default="root_info", deserializable=False)
```
bfolie: For other classes, we've been using `typ` to avoid shadowing the python built-in `type`.

src/citrine/ara/variables.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def _attrs(self) -> List[str]:
+        return ["short_name", "output_name", "field", "type"]
+
+    def __init__(self,
+                 short_name: str,
+                 output_name: List[str],
```
bfolie: It seems strange to use the singular `output_name` if it will generally be a list of length > 1

src/citrine/ara/columns.py
```diff
@@ -0,0 +1,83 @@
...
+    type = properties.String('type', default="identity_column", deserializable=False)
+
+    def _attrs(self) -> List[str]:
+        return ["data_source", "type"]
+
+    def __init__(self,
+                 data_source: str,
+                 target_units: Optional[str] = None):
+        self.data_source = data_source
+        self.target_units = target_units
```
bfolie: Should `IdentityColumn` have `target_units`? It's not in the docstring.

src/citrine/resources/ara_definition.py
```diff
@@ -0,0 +1,78 @@
...
+from citrine._rest.collection import Collection
+from citrine._rest.resource import Resource
+from citrine._serialization import properties
+from citrine._session import Session
+from citrine.ara.columns import Column
+from citrine.ara.rows import Row
+from citrine.ara.variables import Variable
+
+
+class AraDefinition(Resource["AraDefinition"]):
```
bfolie: Will the term "Ara" be user-facing, or will we change this name before release?

src/citrine/ara/columns.py
bfolie: Does every single row gets the same value for an identity column?

tests/resources/test_ara_definition.py
```diff
@@ -0,0 +1,75 @@
...
+    assert ara_definition.uid is None
+    assert ara_definition.version is None
+
+
+def test_dump_example():
+    density = AttributeByTemplate(
+        short_name="density",
+        output_name=["Slice", "Density"],
+        template=LinkByUID(scope="templates", id="density")
+    )
```
bfolie: I find the difference between short_name and output_name (and `field` in the case of RootInfo) confusing. They all seem to be referring to sort-of-but-not-quite-the-same-thing, which is how the variable is referred to. What's the difference?

src/citrine/ara/columns.py
```diff
@@ -0,0 +1,83 @@
...
+        return next(x for x in types if x.type == data["type"])
+
+
+class RealMeanColumn(Serializable['RealMeanColumn'], Column):
+    """Column containing the mean of a real-valued variable.
+
+    Parameters
+    ----------
+    data_source: str
+        short_name of the variable to use when populating the column
```
bfolie: Do all columns have a `data_source` field and do all variables have a `short_name` field? This seems to be implying that there is a uniqueness requirement (all variables in a given Ara definition must have unique short names). Is that true, and what happens if that requirement is not met?

src/citrine/ara/variables.py
maxhutch: I can update it here, but I don't want to broaden the scope of this branch to fixing the pattern everywhere.  :+1: on creating an issue for it.  It probably relates to https://citrine.atlassian.net/browse/PLA-2444

tests/resources/test_ara_definition.py
kroenlein: In looking at the dump, I see `null` in `id`, `version`, and `target_units`.  
* Assuming `id` is the UUID, which will be assigned by the backend, that seems fine.  
* Assuming `version` will be added as we get closer to roll-out, I don't see a clear reason to not put a `0.1` there.
* Was opt-out of defining units intentional?  I.e., does `null` mean use what's in the template?  This could cause table migration if I regenerate a table made 1 year ago whose associated Attribute Template was edited 6 months ago, which feels surprising.  I'd prefer explicitly populating this field from the template (`data_source.template.units`) though I see that this is put together in a way that that template is not available at definition time.

src/citrine/ara/columns.py
```diff
@@ -0,0 +1,83 @@
...
+    target_units: optional[str]
+        units to convert the real variable into
```
kroenlein: Why are `target_units` optional?  What is the default?

src/citrine/ara/columns.py
kroenlein: As per the Slack conversation, `alias` might be more intuitive.  `nickname` operates similarly, but feels less formal.  Would there be value to vetting that there are not collisions in `short_name`s?

src/citrine/ara/variables.py
```diff
@@ -0,0 +1,98 @@
...
+
+from taurus.entity.link_by_uid import LinkByUID
+
+from citrine._serialization.serializable import Serializable
+from citrine._serialization.polymorphic_serializable import PolymorphicSerializable
+from citrine._serialization import properties
+
+
+class Variable(PolymorphicSerializable['Variable']):
+    """A variable that can be assigned values present in material histories.
```
kroenlein: Shouldn't the `output_name` be defined for all `Variable` classes and thus have an abstract method at the parent class level?  Also, this feels more like a taxonomy or hierarchy of labeling, so `output_name` seems to miss on the plurality of terms and their nested nature.  Not that I have a better word at the moment.

src/citrine/ara/variables.py
bfolie: 2444 is also related to deserialization, but I think that 2444 is significantly different in spirit (it focuses on complications around traversing a tree of data concepts objects).

Agree about not broadening the scope, and I don't even think you need to update the code in this PR, since what you wrote adheres to established patterns. I'll make an issue.

jfriend-citrine:
@maxhutch Thanks for your time walking through PR with me this morning. I'm following comments and learning.

src/citrine/ara/columns.py
maxhutch: > Should `IdentityColumn` have `target_units`? It's not in the docstring, and seems out of place for a string.

Nope; I just missed removing it when copying classes around.

maxhutch: > Does every single row gets the same value for an identity column?

No: each row gets the value of the variable in that row.  Variables take on a different value for each row.

src/citrine/resources/ara_definition.py
maxhutch: My expectation is that it will change, but we've released codenames before on a temporary basis.

maxhutch:
> How does the user use an AraDefinition object? Is there a separate route they call to create/get the table? Do they simply include the uid of the AraDefinition object when defining part of the workflow?

They create and register Ara definitions.  Then, they can execute it, which will cause the data platform to read data and produce a Table (which can then be read with the tables route).  The same definition can be re-run at a later time to re-build the Table with new data.

src/citrine/ara/columns.py
maxhutch: They are optional with the default `None`, in which case they are returned with original units.  You'd only expect to use that if you then had the original units as a separate display column.  I can pull that into this minimal example for clarity (note that its already in the backend prototype).

kroenlein: I'd think a default of a non-conversion would be a footgun.  Can we have `None` be an option, which behaves as you say, but default be empty string (or functional equivalent), in which case the behavior is grab the output units from the template?  That would make much more sense as a default to me.

maxhutch: I'd rather have the "by default grab it from the template" behavior be part of the SDK behavior and not part of the API/python binding.  But I do like that being default behavior in the SDK...

src/citrine/ara/columns.py
maxhutch: For the argument name in the column definition?  We had been discussing changing `short_name` in the variable definition.  Do you think `data_source` is confusing?

src/citrine/ara/variables.py
maxhutch: I cannot easily move them into the parent class (see https://github.com/CitrineInformatics/citrine-python/issues/166 ).  Agreed that its the ideal way to do it, but we don't have members like that defined in any of our parent classes.

src/citrine/ara/columns.py
maxhutch: They do.  I made a comment about their interfaces in response to @kroenlein and opened #166.  I've added constructor-time validations in AraDefinition that `short_name` is unique across the variables and each `data_source` is present as the `short_name` of one of the variables.

src/citrine/ara/variables.py
maxhutch: Renamed to `headers`

tests/resources/test_ara_definition.py
maxhutch: Renamed to `name` (the name of the variable to use when referencing it) and `headers` (the things that show up in the header of the table).

src/citrine/ara/columns.py
kroenlein: `data_source` is a good concept and good term.

src/citrine/ara/columns.py
kroenlein: I'm all for handling behavior in the SDK, but would feel better if we swapped it to a required argument if the nominally default behavior would be problematic if implemented as such.

src/citrine/resources/ara_definition.py
```diff
@@ -0,0 +1,101 @@
...
+        # to move into setters.
+        names = [x.name for x in variables]
+        dup_names = self._get_dups(names)
+        if len(dup_names) > 0:
+            raise ValueError("Multiple variables defined these names,"
+                             " which much be unique: {}".format(dup_names))
+        headers = [x.headers for x in variables]
+        dup_headers = self._get_dups(headers)
+        if len(dup_headers) > 0:
+            raise ValueError("Multiple variables defined these output_names,"
```
bfolie:
```suggestion
            raise ValueError("Multiple variables defined these headers,"
```

tests/resources/test_ara_definition.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def test_init_ara_definition():
+    ara_definition = AraDefinition(name="foo", description="bar", rows=[], columns=[], variables=[], datasets=[])
+    assert ara_definition.uid is None
+    assert ara_definition.version is None
+
+
+def test_dup_names():
+    """Make sure that variable short_name and output_name are unique across an ara definition"""
```
bfolie:
```suggestion
    """Make sure that variable name and headers are unique across an ara definition"""
```

src/citrine/ara/columns.py
```diff
@@ -0,0 +1,86 @@
...
+        self.target_units = target_units
+
+
+class IdentityColumn(Serializable['IdentityColumn'], Column):
+    """Column containing the value of a string-valued variable.
+
+    Parameters
+    ----------
+    data_source: str
+        short_name of the variable to use when populating the column
```
kroenlein:
```suggestion
        name of the variable to use when populating the column
```

maxhutch:
Great!  I'm going to protect this branch and close the PR.

PR #164, opened 2020-01-07
Proposed bugfix for PLA-2416: using pagination for listing collections

crleblanc:
# Citrine Python PR

## Description 
This PR resolves https://citrine.atlassian.net/browse/PLA-2416 (failing platform end to end tests).  This is caused by the list method from the Collection class.  It is not using pagination when making calls to the backend API services.

This bug was difficult to reproduce locally.  It requires more than 20 elements in a collection before the tests will fail.  This also explains why the tests started failing in Jenkins without a change to the code to cause this.  Test data persists in the test DB between tests so it took days/weeks for this data to reach more than 20 elements in a collection.

This PR changes the default operation of list() to return all elements of a collection instead of the first page.  If the 'page' option is supplied then it will only return a response for that page.  As a user, I would expect a .list() method to list all elements, not just the first page or elements.

The new list() method has the same signature and return values as before so this should not break any existing code.  Any subclasses of Collection outside citrine-python that overload list() should rename this method to _fetch_page() if they wish to use pagination.  _fetch_page() can return an iteratable; so an iterator, list or tuple are acceptable responses.

All list() methods in the base class and subclasses have been renamed to _fetch_page().  A new list() method has been created on the base class that calls _fetch_page() as many times as required to fetch all elements in the collection.  This includes some logic to avoid infinite loops from API calls that ignore the page/per_page options.

The unit test for listing datasets was modified to include pagination.

### PR Type:

### DIFF
--- a/src/citrine/_rest/collection.py
+++ b/src/citrine/_rest/collection.py
@@ -1,6 +1,7 @@
 from abc import abstractmethod
 from typing import Optional, Union, Generic, TypeVar, Iterable
 from uuid import UUID
+import warnings
 
 from citrine.exceptions import ModuleRegistrationFailedException, NonRetryableException
 from citrine.resources.response import Response
@@ -12,8 +13,6 @@ ResourceType = TypeVar('ResourceType', bound='Resource')
 # Thus, this will never be particularly type safe on its own. The solution is to
 # have subclasses override the create method.
 CreationType = TypeVar('CreationType', bound='Resource')
-
-DEFAULT_PER_PAGE = 20
 
 
 class Collection(Generic[ResourceType]):
@@ -106,7 +105,7 @@ class Collection(Generic[ResourceType]):
 
     def list(self,
              page: Optional[int] = None,
-             per_page: Optional[int] = None) -> Iterable[ResourceType]:
+             per_page: int = 100) -> Iterable[ResourceType]:
         """
         List all visible elements in the collection.
 
@@ -117,11 +116,11 @@ class Collection(Generic[ResourceType]):
         ---------
         page: int, optional
             The "page" of results to list. Default is to read all pages and yield
-            all results.
+            all results.  This option is deprecated.
         per_page: int, optional
-            Max number of results to return per page. Default is 20.  If the page
-            parameter is specified this will limit the number of results in that
-            response.
+            Max number of results to return per page. Default is 100.  This parameter
+            is used when making requests to the backend service.  If the page parameter
+            is specified it limits the maximum number of elements in the response.
 
         Returns
         -------
@@ -129,44 +128,43 @@ class Collection(Generic[ResourceType]):
             Resources in this collection.
 
         """
-        # Do an initial fetch of the first page using supplied args.
-        # If more results expected, proceed with paginated calls.
-        first_page = self._fetch_page(page=page, per_page=per_page)
-        first_page_count = 0
+        if page is not None:
+            warnings.warn("The page parameter is deprecated, default is automatic pagination",
+                          DeprecationWarning)
+
         first_uid = None
-        for idx, element in enumerate(first_page):
-            yield element
-
-            uid = getattr(element, 'uid', None)
-            if idx == 0 and not first_uid and uid:
-                first_uid = uid
-
-            first_page_count += 1
-
-        if page is not None or first_page_count != (per_page or DEFAULT_PER_PAGE):
-            return
-
-        # Read extra pages until nothing is left to read
-        next_page = 2
+        page_idx = page
         while True:
-            subset = self._fetch_page(page=next_page, per_page=per_page)
+            subset = self._fetch_page(page=page_idx, per_page=per_page)
+
             count = 0
             for idx, element in enumerate(subset):
                 # escaping from infinite loops where page/per_page are not
                 # honored and are returning the same results regardless of page:
                 uid = getattr(element, 'uid', None)
                 if first_uid is not None and first_uid == uid:
+                    # TODO: raise an exception once the APIs that ignore pagination are fixed
                     break
 
                 yield element
 
+                if idx == 0:
+                    first_uid = uid
+
                 count += 1
 
-            # Handle the case where we get an unexpected number of results (e.g. last page)
-            if count == 0 or count < first_page_count:
+            # If the page number is specified we exit to disable auto-paginating
+            if page is not None:
                 break
 
-            next_page += 1
+            # Handle the case where we get an unexpected number of results (e.g. last page)
+            if count == 0 or count < per_page:
+                break
+
+            if page_idx is None:
+                page_idx = 2
+            else:
+                page_idx += 1
 
     def update(self, model: CreationType) -> CreationType:
         """Update a particular element of the collection."""
--- a/src/citrine/_rest/collection.py
+++ b/src/citrine/_rest/collection.py
@@ -60,6 +60,7 @@ class Collection(Generic[ResourceType]):
                     per_page: Optional[int] = None) -> Iterable[ResourceType]:
         """
         Fetch visible elements in the collection.  This does not handle pagination.
+
         This method will return the first page of results using the default page/per_page
         behaviour of the backend service.  Specify page/per_page to override these defaults
         which are passed to the backend service.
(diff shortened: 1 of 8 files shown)


crleblanc:
Hmm, I see there's some minor Flake8 issues that are blocking this PR.  I'll fix those up.

bfolie:
Methods have already been added for getting all paged resources, which I think can greatly simplify this PR. See, for example, #145. @asantas93 can weigh in better than I can

crleblanc:
Thanks @bfolie, I wasn't aware of that work by @asantas93.  I had a quick look at cursor_paged_resource() and I have a couple questions:

 - It seems to supply additional params to the API (forward, ascending, etc) that are not used in many of the V1 endpoints.  Does this mean we can't use this for existing endpoints (e.g. listing datasets) without adding support for these params?
 - I see the default is to use V2 of the 'base_method'.  Is the V2 required to support these new params?

While I would like to standardize on one approach to pagination, I wonder if we could use my approach as a short term bugfix and migrate to this other approach in the long term?

bfolie:
> Thanks @bfolie, I wasn't aware of that work by @asantas93. I had a quick look at cursor_paged_resource() and I have a couple questions:
> 
> * It seems to supply additional params to the API (forward, ascending, etc) that are not used in many of the V1 endpoints.  Does this mean we can't use this for existing endpoints (e.g. listing datasets) without adding support for these params?
> * I see the default is to use V2 of the 'base_method'.  Is the V2 required to support these new params?
> 
> While I would like to standardize on one approach to pagination, I wonder if we could use my approach as a short term bugfix and migrate to this other approach in the long term?

I don't think you need to supply any new params. As you can see in `list_by_tag` in the `DataConcepts` collection, it calls `cursor_paged_resource` using `self.session.get_resource` as the 'base_method'. The params "forward" and "ascending" are not included, because they have default values that are filled in. My understanding is that those parameters are important for browsers, but not the python client.

We have other E2E tests that rely on `list`-type methods returning everything (not just the first page) and they use the v2 endpoints, so I think you should be able to use this approach as well.

crleblanc:
That's interesting.  I'll see if I can swap over to the cursor_paged_resource.

src/citrine/_rest/collection.py
```diff
@@ -98,6 +104,70 @@ def list(self,
...
+        next_page = 2
+        while True:
+            subset = self._fetch_page(page=next_page, per_page=per_page)
+            count = 0
+            for idx, element in enumerate(subset):
+                # escaping from infinite loops where page/per_page are not
+                # honored and are returning the same results regardless of page:
+                uid = getattr(element, 'uid', None)
+                if first_uid is not None and first_uid == uid:
+                    break
```
asantas93: Should we throw an exception here? Is this ever valid server-side behavior?

src/citrine/_rest/collection.py
```diff
@@ -13,6 +13,8 @@
 # have subclasses override the create method.
 CreationType = TypeVar('CreationType', bound='Resource')
 
+DEFAULT_PER_PAGE = 20
```
asantas93: Does anyone else think it makes sense to have a smaller default page size for citrine-python? 20ish makes sense for the UI, but I imagine for most citrine-python use cases, we will just end up iterating over the entire result set anyway, meaning we'll need more requests with a small page size.

In the cursor-paging function, I default to `100`, but that wasn't a collective decision, really. cc: @mcgalcode @djack

src/citrine/_rest/collection.py
```diff
@@ -98,6 +104,70 @@ def list(self,
...
+        for idx, element in enumerate(first_page):
+            yield element
+
+            uid = getattr(element, 'uid', None)
+            if idx == 0 and not first_uid and uid:
+                first_uid = uid
+
+            first_page_count += 1
+
+        if page is not None or first_page_count != (per_page or DEFAULT_PER_PAGE):
```
asantas93: Cool way to make this change non-breaking 👍

src/citrine/_rest/collection.py
```diff
@@ -98,6 +104,70 @@ def list(self,
...
+        """
+        List all visible elements in the collection.
+
+        Leaving page and per_page as default values will yield all elements in the
+        collection, paginating over all available pages.
+
+        Parameters
+        ---------
+        page: int, optional
+            The "page" of results to list. Default is to read all pages and yield
```
asantas93: FWIW I would support deprecating the `page` parameter now that you have this default behavior.

asantas93:
Glad to see this stuff is being abstracted away from the user.

> Methods have already been added for getting all paged resources, which I think can greatly simplify this PR. See, for example, #145. @asantas93 can weigh in better than I can

My change was for cursor-based paging (https://github.com/CitrineInformatics/citrine-python/blob/develop/src/citrine/_session.py#L150), which doesn't apply here without some server-side changes. I will say, though, that pulling out this logic to be reusable outside of `list` may be useful in some other places, too. https://github.com/CitrineInformatics/citrine-python/blob/develop/src/citrine/resources/data_concepts.py#L551 will be deprecated, but would still be improved by this. I'm guessing in the future we will have things like listing datasets by name that we would want to page like this, too.

asantas93:
> Thanks @bfolie, I wasn't aware of that work by @asantas93. I had a quick look at cursor_paged_resource() and I have a couple questions:
> 
> * It seems to supply additional params to the API (forward, ascending, etc) that are not used in many of the V1 endpoints.  Does this mean we can't use this for existing endpoints (e.g. listing datasets) without adding support for these params?

I think you would be okay to just not expose those parameters to the user for collections that don't support reverse sorting. I did this for [`list_by_tag`](https://github.com/CitrineInformatics/citrine-python/blob/develop/src/citrine/resources/data_concepts.py#L757) because there's not really a meaningful ordering in the first place. The `forward=true&ascending=true` shouldn't do any harm and will be respected as far as citrine-python knows.

> * I see the default is to use V2 of the 'base_method'.  Is the V2 required to support these new params?

There's no constraint to use `v2` client-side. I'm not sure about how pippin filters the query params though.

> While I would like to standardize on one approach to pagination, I wonder if we could use my approach as a short term bugfix and migrate to this other approach in the long term?

All in favor.

src/citrine/_rest/collection.py
bfolie: Previously the server was always returning the last page, leading to infinite loops. If there were 40 elements and per_page=20, then page 1 returns elements1-20, page 2 returns elements 21-40, and page 3 _also_ returns elements 21-40. It's possible that your work on paging also fixed this behavior for v1 routes, but if not, then some check like this will be necessary.

However, I do not think this logic will solve the above problem. If I understand correctly, this checks to see if the very first uid ever gets returned again. But the problem (which may have changed) wasn't that it looped back to the very beginning of the list, but that it looped back to the beginning of the last page.

src/citrine/_rest/collection.py
bfolie: Agreed.

Unrelated, I suggest that you have `list` pass `DEFAULT_PER_PAGE` if `per_page` is not specified. Otherwise, there is the assumption that two unrelated pieces of logic align (the product API default per_page must equal the default per_page here, but they could be changed independently of each other)

src/citrine/_rest/collection.py
crleblanc: I didn't know about the problem you mentioned above.  I have seen an endpoint that completely ignores the page/per_page params and returns all items on every call so was trying to detect that case.

I think there's a relatively easy way to modify this to support both cases.  Instead of checking against the first UID we could check the first element in the previous list of elements.  I'll implement that now.  For the time being I would like to avoid raising an exception so we can get the end to end tests passing.  Once I fix the misbehaving endpoint I think we should raise an exception if this behavior is detected.

I also came to the conclusion that this cursor based pagination won't work for the v1 page/per_page paginated endpoints.  I think this makes this PR necessary.  Once we fully support cursor based pagination in the API we can swap over.

src/citrine/_rest/collection.py
crleblanc: I don't have a strong opinion on this.  I just selected this to be the same as the API.  I'm happy to use 100.

src/citrine/_rest/collection.py
crleblanc: Good point.  I was leaving this up to the API but it makes sense to honor the per_page parameter.

src/citrine/_rest/collection.py
crleblanc: That's a good point.  I don't want to break any existing clients so raising a deprecation warning is a good idea.

crleblanc:
I've added a new commit to this branch that addresses all of the points that were raised.  Hopefully it's all g2g now.

PR #168, opened 2020-01-09
Added type hints for Taurus objects for DataConcepts.register()

crleblanc:
# Citrine Python PR

## Description 
Adding type hints to the DataConcepts.register method.  This should make it more obvious that this method can accept Taurus objects.

### PR Type:

### DIFF
--- a/src/citrine/resources/data_concepts.py
+++ b/src/citrine/resources/data_concepts.py
@@ -17,6 +17,7 @@ from citrine._utils.functions import (
 from taurus.client.json_encoder import loads, dumps, LinkByUID
 from taurus.entity.dict_serializable import DictSerializable
 from taurus.entity.bounds.base_bounds import BaseBounds
+from taurus.entity.object.base_object import BaseObject
 from taurus.entity.template.attribute_template import AttributeTemplate
 
 from citrine.resources.response import Response
@@ -376,7 +377,7 @@ class DataConcepts(PolymorphicSerializable['DataConcepts']):
 ResourceType = TypeVar('ResourceType', bound='DataConcepts')
 # A Taurus entity that is a subclass of taurus.entity.object.base_object.BaseObject
 # e.g. MaterialRun, MaterialSpec, etc.
-TaurusEntityType = TypeVar('TaurusEntityType', bound='BaseObject')
+TaurusEntityType = TypeVar('TaurusEntityType', bound=BaseObject)
 
 
 class DataConceptsCollection(Collection[ResourceType]):
--- a/src/citrine/resources/data_concepts.py
+++ b/src/citrine/resources/data_concepts.py
@@ -374,6 +374,9 @@ class DataConcepts(PolymorphicSerializable['DataConcepts']):
 
 
 ResourceType = TypeVar('ResourceType', bound='DataConcepts')
+# A Taurus entity that is a subclass of taurus.entity.object.base_object.BaseObject
+# e.g. MaterialRun, MaterialSpec, etc.
+TaurusEntityType = TypeVar('TaurusEntityType', bound='BaseObject')
 
 
 class DataConceptsCollection(Collection[ResourceType]):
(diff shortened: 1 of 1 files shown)


src/citrine/resources/data_concepts.py
```diff
@@ -374,6 +374,9 @@ def as_dict(self) -> dict:
 
 
 ResourceType = TypeVar('ResourceType', bound='DataConcepts')
+# A Taurus entity that is a subclass of taurus.entity.object.base_object.BaseObject
+# e.g. MaterialRun, MaterialSpec, etc.
+TaurusEntityType = TypeVar('TaurusEntityType', bound='BaseObject')
```
maxhutch: How does this generic get assigned to a concrete class?  `ResourceType` is assigned via the collection, but `TaurusEntityType` appears to be unused.

crleblanc: I'm not familiar with the details of Python type hinting and generics.  I'm surprised to see how convoluted it is.  It looks like I used the incorrect syntax.  Looking at https://www.python.org/dev/peps/pep-0484/#generics and https://docs.python.org/3/library/typing.html#classes-functions-and-decorators it looks like this could be written as

```
from taurus.entity.object.base_object import BaseObject
...
TaurusEntityType = TypeVar('TaurusEntityType', BaseObject)
```
This would indicate that any instance of a subclass of BaseObject will be an acceptable TaurusEntityType.  How does that sound?

maxhutch: I understand that as being a type upper bound, but not actually binding the type.  I'd defer to someone who understands python type hints better, though.  cc: @mcgalcode : do we have a python typing expert?

crleblanc: I've looked into this and found that the snippet above works.  TaurusEntityType does not have to be generic.  It's used lower in the file for the argument types of the register method: `Union[ResourceType, TaurusEntityType]`.

Here's a demo file that works fine with `mypy --strict`.  It's doing the same thing as above, specifying that it expects a subclassed instance of Base.  The only difference is we're giving it a class instead of a string when using bound=<>.  This is easy to change in the PR.

```
from typing import TypeVar, Union

class Alien:
    a: int = 999

class Base:
    a: int = 123

class Derived(Base):
    a: int = 456


T = TypeVar('T', bound=Derived)

class UseTypes:

    def check(self, obj: Union[Alien, T]) -> int:
        return obj.a

alien = Alien()
base = Base()
derived = Derived()

u = UseTypes()
print(u.check(alien))       # works
print(u.check(derived))   # works
print(u.check(base))       # fails, as expected
```

maxhutch: My concern is specifically that `project.measurement_runs.register(ProcessSpec(...))` will work because the generic is never narrowed beyond `BaseObject`.  It sounds like there is some debt in our testing of typing that's preventing us from going TDD on this, though**.  Is that right, or could we test this?

** Even when I try to run mypy locally, it cannot find type hints for the types that I'm trying to check, e.g.:
```
from citrine.resources.dataset import Datase
from taurus.entity.object.process_spec import ProcessSpec


def foo(dataset: Dataset):
    dataset.material_runs.register(ProcessSpec("foo"))
```
gives me:
```
foo.py:1: error: Cannot find implementation or library stub for module named 'citrine.resources.dataset'
foo.py:1: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
foo.py:2: error: Cannot find implementation or library stub for module named 'taurus.entity.object.process_spec'
foo.py:5: error: Name 'Dataset' is not defined
```

crleblanc: I'm not sure how these type hints were run in the past.  I think there's a possibility that mypy was never used to check these types.  I don't see any reference to mypy in .travis.yml so this tool is not being run in CI.  I see the same sorts of issues, and many others, when I try to run mypy.

Are you concerned someone will pass an instance of a Taurus BaseObject to register?  Given that these are just type hints and not enforced I don't see that as a major issue.

mcgalcode:
cc: @crleblanc  changed base branch to master as part of the master flow update.

PR #173, opened 2020-01-14
Adding display column definitions for Ara

maxhutch:
This adds all of the ones that are in scope (no composition).  We're
also replacing the "lower confidence bound" and "upper confidence bound"
columns with a quantile column.

### DIFF
--- a/src/citrine/ara/columns.py
+++ b/src/citrine/ara/columns.py
@@ -52,6 +52,7 @@ class MeanColumn(Serializable['MeanColumn'], Column):
         units to convert the real variable into
 
     """
+
     data_source = properties.String('data_source')
     target_units = properties.Optional(properties.String, "target_units")
     typ = properties.String('type', default="real_mean_column", deserializable=False)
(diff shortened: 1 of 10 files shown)


src/citrine/ara/columns.py
```diff
@@ -31,15 +31,17 @@ def get_type(cls, data) -> Type[Serializable]:
         if "type" not in data:
             raise ValueError("Can only get types from dicts with a 'type' key")
         types: List[Type[Serializable]] = [
-            RealMeanColumn, IdentityColumn
+            IdentityColumn,
+            MeanColumn, StdDevColumn, QuantileColumn, OriginalUnitsColumn,
+            MostLikelyCategoryColumn, MostLikelyProbabilityColumn
```
kroenlein: I don't think `MostLikelyProbabilityColumn` literally means what you want it to.  `HighestLikelihood`?

maxhutch: Let's take a rep at those later.  cc: @gregor-robinson

PR #184, opened 2020-01-28
Add Ara variables.

gregor-robinson:
There are still a couple TODOs here requiring attention before merging to feature/ara but I'm pushing this for visibility before the hack session. Even the most nitpicky feedback is eagerly welcomed.

### DIFF
--- a/src/citrine/_serialization/properties.py
+++ b/src/citrine/_serialization/properties.py
@@ -6,6 +6,7 @@ from datetime import datetime
 import uuid
 import arrow
 
+from taurus.enumeration.base_enumeration import BaseEnumeration
 from taurus.entity.link_by_uid import LinkByUID
 from taurus.entity.dict_serializable import DictSerializable
 from citrine._serialization.serializable import Serializable
@@ -377,6 +378,35 @@ class MixedList(Property[list, list]):
             serialized.append(element_type.serialize(element_type.default))
 
         return serialized
+
+
+class Enumeration(Property[BaseEnumeration, BaseEnumeration]):
+
+    def __init__(self,
+                 klass: typing.Type[typing.Any],
+                 serialization_path: typing.Optional[str] = None,
+                 serializable: bool = True,
+                 deserializable: bool = True,
+                 default: typing.Optional[DeserializedType] = None):
+        super().__init__(serialization_path,
+                         serializable,
+                         deserializable,
+                         default)
+        self.klass = klass
+
+    @property
+    def underlying_types(self):
+        return self.klass
+
+    @property
+    def serialized_types(self):
+        return str
+
+    def _deserialize(self, value: str) -> uuid.UUID:
+        return self.klass.get_enum(value)
+
+    def _serialize(self, value: typing.Any) -> str:
+        return self.klass.get_value(value)
 
 
 class Object(Property[typing.Any, dict]):
--- a/src/citrine/_serialization/properties.py
+++ b/src/citrine/_serialization/properties.py
@@ -380,7 +380,7 @@ class MixedList(Property[list, list]):
         return serialized
 
 
-class Enumeration(Property[BaseEnumeration, BaseEnumeration]):
+class Enumeration(Property[BaseEnumeration, str]):
 
     def __init__(self,
                  klass: typing.Type[typing.Any],
--- a/src/citrine/ara/variables.py
+++ b/src/citrine/ara/variables.py
@@ -1,12 +1,20 @@
 """Variable definitions for Ara."""
-from typing import Type, Optional, List  # noqa: F401
 from abc import abstractmethod
+from enum import Enum
+from typing import Type, Optional, List  # noqa: F401
 
 from taurus.entity.link_by_uid import LinkByUID
 
 from citrine._serialization.serializable import Serializable
 from citrine._serialization.polymorphic_serializable import PolymorphicSerializable
 from citrine._serialization import properties
+
+
+class IngredientQuantityDimension:
+    absolute = 1
+    mass = 2
+    volume = 3
+    number = 4
 
 
 class Variable(PolymorphicSerializable['Variable']):
@@ -33,7 +41,10 @@ class Variable(PolymorphicSerializable['Variable']):
         if "type" not in data:
             raise ValueError("Can only get types from dicts with a 'type' key")
         types: List[Type[Serializable]] = [
-            RootInfo, AttributeByTemplate
+            RootInfo, AttributeByTemplate, AttributeByTemplateAfterProcessTemplate,
+            AttributeByTemplateAndObjectTemplate, IngredientIdentifierByProcessTemplateAndName,
+            IngredientLabelByProcessAndName, IngredientQuantityByProcessAndName,
+            RootIdentifier
         ]
         res = next((x for x in types if x.typ == data["type"]), None)
         if res is None:
@@ -52,7 +63,6 @@ class RootInfo(Serializable['RootInfo'], Variable):
         sequence of column headers
     field: str
         name of the field to assign the variable to
-
     """
 
     name = properties.String('name')
@@ -83,7 +93,6 @@ class AttributeByTemplate(Serializable['AttributeByTemplate'], Variable):
         sequence of column headers
     template: LinkByUID
         attribute template that identifies the attribute to assign to the variable
-
     """
 
     name = properties.String('name')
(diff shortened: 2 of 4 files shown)


src/citrine/ara/variables.py
```diff
@@ -9,6 +10,13 @@
 from citrine._serialization import properties
 
 
+class IngredientQuantityDimension:
```
maxhutch: Do these serialize to their string names, or to their numbers?  In `taurus`, we define enumerations with this pattern: https://github.com/CitrineInformatics/taurus/blob/develop/taurus/enumeration/origin.py

src/citrine/ara/variables.py
```diff
@@ -101,3 +110,232 @@ def __init__(self, *,
         self.name = name
         self.headers = headers
         self.template = template
+
+
+class AttributeByTemplateAfterProcessTemplate(Serializable['AttributeByTemplateAfterProcessTemplate'], Variable):
+    """Attribute of an object that is marked by an attribute template and derives from a process
+    marked by a locally unique object template.
+
+    TODO: define "locally-unique"
```
maxhutch: Unique within the scope of the material history that's being converted into the row

src/citrine/ara/variables.py
```diff
@@ -101,3 +110,232 @@ def __init__(self, *,
...
+        self.name = name
+        self.headers = headers
+        self.attributeTemplate = attributeTemplate
+        self.processTemplate = processTemplate
+
+
+class AttributeByTemplateAndObjectTemplate(Serializable['AttributeByTemplateAndObjectTemplate'], Variable):
+    """Attribute marked by an attribute template and an object template.
+
+    TODO: clarify documentation of use case here.
```
maxhutch: An example would be if the same property is measured with two different measurement techniques.  The property would have the same attribute template, but the measurements would have different measurement templates, allowing them to be distinguished.

src/citrine/ara/variables.py
```diff
@@ -101,3 +110,232 @@ def __init__(self, *,
         self.name = name
         self.headers = headers
         self.template = template
+
+
+class AttributeByTemplateAfterProcessTemplate(Serializable['AttributeByTemplateAfterProcessTemplate'], Variable):
+    """Attribute of an object that is marked by an attribute template and derives from a process
+    marked by a locally unique object template.
```
bfolie:
```suggestion
    marked by a locally unique process template.
```
It's also an object template, but might as well be as specific as possible.

src/citrine/ara/variables.py
```diff
@@ -101,3 +110,232 @@ def __init__(self, *,
...
+    name = properties.String('name')
+    headers = properties.List(properties.String, 'headers')
+    attributeTemplate = properties.Object(LinkByUID, 'attributeTemplate')
+    processTemplate = properties.Object(LinkByUID, 'processTemplate')
+    typ = properties.String('type', default="attribute_after_process", deserializable=False)
+
+    def _attrs(self) -> List[str]:
+        return ["name", "headers", "attributeTemplate", "processTemplate", "typ"]
+
+    def __init__(self, *,
```
bfolie: Why are we leaving room for additional arguments? Do we expect the interface to change? @maxhutch

src/citrine/ara/variables.py
```diff
@@ -101,3 +110,232 @@ def __init__(self, *,
...
+    ---------
+    name: str
+        a short human-readable name to use when referencing the variable
+    headers: list[str]
+        sequence of column headers
+    processTemplate: LinkByUID
+        process template associated with this ingredient identifier
+    ingredientName: str
+        name of ingredient
+    scope: str
```
bfolie: What is scope?

src/citrine/ara/variables.py
```diff
@@ -101,3 +110,232 @@ def __init__(self, *,
...
+                 scope: str):
+        self.name = name
+        self.headers = headers
+        self.processTemplate = processTemplate
+        self.ingredientName = ingredientName
+        self.scope = scope
+
+
+class IngredientLabelByProcessAndName(Serializable['IngredientLabelByProcessAndName'], Variable):
+    """Define a boolean variable indicating whether a given label is applied to an in ingredient
```
bfolie:
```suggestion
    """Define a boolean variable indicating whether a given label is applied to an ingredient
```

src/citrine/ara/variables.py
```diff
@@ -101,3 +110,232 @@ def __init__(self, *,
...
+                 ingredientName: str,
+                 scope: str):
+        self.name = name
+        self.headers = headers
+        self.processTemplate = processTemplate
+        self.ingredientName = ingredientName
+        self.scope = scope
+
+
+class IngredientLabelByProcessAndName(Serializable['IngredientLabelByProcessAndName'], Variable):
```
bfolie: This name makes me think that it will return the label of the ingredient with a given name and process, not a boolean indicating whether or not the ingredient has a specified label.

src/citrine/ara/variables.py
```diff
@@ -101,3 +110,232 @@ def __init__(self, *,
...
+                 label: str):
+        self.name = name
+        self.headers = headers
+        self.processTemplate = processTemplate
+        self.ingredientName = ingredientName
+        self.label = label
+
+
+class IngredientQuantityByProcessAndName(Serializable['IngredientQuantityByProcessAndName'], Variable):
+    """Get the dimension of the quantity in an ingredient that is associated with a process
```
bfolie:
```suggestion
    """Get the quantity of an ingredient that is associated with a process
```

src/citrine/ara/variables.py
bfolie: Since you wrote the code this might be a rhetorical question, but it looks to me like they serialize to strings: https://github.com/CitrineInformatics/platform-backend/blob/develop/code/ara/model/src/main/scala/io/citrine/ara/model/variable/IngredientQuantity.java

src/citrine/ara/variables.py
maxhutch: `*` in the middle of the arguments list isn't for position arguments; it is for [keyword-only arguments](https://www.python.org/dev/peps/pep-3102/).

src/citrine/ara/variables.py
maxhutch: Yeah, the backend is expecting them as strings.

src/citrine/ara/variables.py
```diff
@@ -101,3 +110,232 @@ def __init__(self, *,
...
+        name of ingredient
+    quantity: IngredientQuantity
+        dimension of the ingredient quantity (e.g. absolute, number fraction...)
+    """
+
+    name = properties.String('name')
+    headers = properties.List(properties.String, 'headers')
+    processTemplate = properties.Object(LinkByUID, 'processTemplate')
+    ingredientName = properties.String('ingredientName')
+    quantityDimension = properties.Integer('quantityDimension')
```
bfolie: Based on the platform-backend definition, I think that this variable will be serialized with the key "quantity". I do see the logic for changing it to `quantityDimension` for the purposes of this user-facing client. That can be done here. For example, you'll notice that the line below is `typ = properties.String('type')`. That says that it should be serialized as "type" but deserialized as "typ". You can do the same thing for quantity/quantityDimensions, and then create an Enum object and in the `__init__` method apply that Enum object to the value of `quantityDimension` to ensure that corresponds to one of the four expected values.

src/citrine/ara/variables.py
```diff
@@ -101,3 +110,232 @@ def __init__(self, *,
...
+
+    def _attrs(self) -> List[str]:
+        return ["name", "headers", "processTemplate", "ingredientName", "quantityDimension", "typ"]
+
+    def __init__(self, *,
+                 name: str,
+                 headers: List[str],
+                 processTemplate: LinkByUID,
+                 ingredientName: str,
+                 quantityDimension: int):
```
bfolie: I think this will be a string

src/citrine/ara/variables.py
bfolie: That's confusing, since **kwargs uses two asterisks. But regardless, do we expect to add more fields?

maxhutch: Not specifically, but possibly.  I'm doing keyword-only arguments everywhere in Ara...

src/citrine/ara/variables.py
gregor-robinson: Agreed. I had the same confusion. How about `IngredientHasLabelByProcessAndName`? I think that change needs to be coordinated on the backend.

maxhutch: Works for me.  That backend change is easy to make (and I've been expecting to massage the names as we go).

src/citrine/ara/variables.py
gregor-robinson: I went ahead and added an Enumerated property type.

src/citrine/_serialization/properties.py
```diff
@@ -379,6 +380,35 @@ def _serialize(self, value: tuple) -> list:
         return serialized
 
 
+class Enumeration(Property[BaseEnumeration, BaseEnumeration]):
```
bfolie: I think this should be `class Enumeration(Property[BaseEnumeration, String]):`, since the serialized type is a string

gregor-robinson: Good catch; corrected to `str`

bfolie:
Could you add a check that attempting to deserialize from a string that is _not_ in the Enumeration leads to an error?

PR #185, opened 2020-01-29
Fixed key-value order in expression aliases.

latture:
The expression predictor expects a map from alias to descriptor key. This was reversed in citrine-python.


kyle-laplante:
should we update this as well?

latture:
Yes, sorry. I made those changes, but forgot to add it to the commit. I'll open a new PR.

PR #187, opened 2020-01-29
Include alpha designation in the README

maxhutch:


README.md
```diff
@@ -98,3 +98,8 @@ originating module will locate the correct instance, e.g. the log line
 INFO:citrine._session:200 GET /projects/fc568490-224a-4070-807f-1427c4f4dcd8
 ```
 is an example of output from the logger in the previous example.
+
+## Developement Status
+
+Classes and methods may be marked as *alpha* by including `[ALPHA]` at the start of their docstrings.
+They should be not be used, and may change or be removed without notice.
```
jsomara: Perhaps clarify a little bit - 
"These methods are intended for development, testing, and experimentation, are not supported, and may change or be removed without notice"

jsomara: How about this instead:

How about this instead: 
```suggestion
These methods are intended for development, testing, and experimentation, are not supported, and may change or be removed without notice```

PR #188, opened 2020-01-29
[ALPHA] Ara definition

maxhutch:
There isn't presently any backend support, so the paths in the resource will need to be updated yet.  All of this code was independently reviewed and marked as ALPHA.


maxhutch:
> LGTM - very excited to see this in action!

ALPHA

PR #196, opened 2020-02-05
Add support for finding materials runs from templates

dekatzenel:
# Citrine Python PR

## Description 
Add support for finding materials runs from templates

Also add intermediate support for finding:
-materials specs from templates
-materials runs from specs

### PR Type:


tests/resources/test_material_spec.py
```diff
@@ -0,0 +1,46 @@
...
+
+@pytest.fixture
+def collection(session) -> MaterialSpecCollection:
+    return MaterialSpecCollection(
+        project_id=UUID('6b608f78-e341-422c-8076-35adc8828545'),
+        dataset_id=UUID('8da51e93-8b55-4dd3-8489-af8f65d4ad9a'),
+        session=session)
+
+
+def test_get_runs(collection, session):
```
kroenlein: Can you explicitly define (comment) your intent with the test?

tests/resources/test_material_template.py
```diff
@@ -0,0 +1,78 @@
...
+
+@pytest.fixture
+def spec_collection(session) -> MaterialSpecCollection:
+    return MaterialSpecCollection(
+        project_id=UUID('6b608f78-e341-422c-8076-35adc8828545'),
+        dataset_id=UUID('8da51e93-8b55-4dd3-8489-af8f65d4ad9a'),
+        session=session)
+
+
+def test_get_specs(collection, session):
```
kroenlein: Can you explicit define the intent of your test?

tests/resources/test_material_template.py
```diff
@@ -0,0 +1,78 @@
...
+        path="projects/{}/material-templates/{}/{}/material-specs".format(project_id, test_scope, test_id),
+    )
+    assert session.last_call == expected_call
+    assert key in specs
+    output = specs[key]
+    assert 1 == len(output)
+    assert output == [sample_spec]
+
+
+def test_get_runs(collection, session, spec_collection):
```
kroenlein: Can you explicitly define the intent of the test?

src/citrine/resources/material_spec.py
```diff
@@ -87,3 +87,18 @@ class MaterialSpecCollection(DataConceptsCollection[MaterialSpec]):
     def get_type(cls) -> Type[MaterialSpec]:
         """Return the resource type in the collection."""
         return MaterialSpec
+
+    def get_runs(self, scope: str, id: str) -> dict:
```
kroenlein: The primary `get` operations for collections are defined [here](https://github.com/CitrineInformatics/citrine-python/blob/391283f46bc8cead857df4a25c04b8b319a84e7f/src/citrine/resources/data_concepts.py#L519).  What's the thought in making this a MaterialSpecCollection specific method instead of something more general?

src/citrine/resources/material_template.py
```diff
@@ -115,3 +116,41 @@ class MaterialTemplateCollection(DataConceptsCollection[MaterialTemplate]):
...
+        :param scope: The scope used to locate the material template.
+        :param id: The unique id corresponding to `scope`.
+            The lookup will be most efficient if you use the Citrine ID (scope='id')
+            of the material template.
+        :return: A search result of material specs
+        """
+        path = self._get_path(ignore_dataset=True) + "/" + scope + "/" + id + "/material-specs"
+        return self.session.get_resource(path)
+
+    def get_runs(self,
```
kroenlein: Comment above, considering that you've implemented two methods with the name `get_runs`.

dekatzenel:
Broken when the backend endpoint returns more than one page of search results. Working on fix.

dekatzenel:
Paging is broken, but in a way that is probably caused by the backend. Still need to respond to comments

src/citrine/resources/material_spec.py
dekatzenel: The methods that I'm adding methods for are specific to material templates and material specs per PLA-2138, so it didn't seem to make sense to put them on a shared type. Definitely open to discussion and others' opinions

kroenlein: Just soliciting discussion.  Given how tightly scoped this is, it's the right choice.  We should probably put a ticket in to make these more general methods or to detail the full suite of appropriate routes.

kroenlein:
Have you had a chance to do an end-to-end spot check?

dekatzenel:
@kroenlein no because the backend changes haven't been merged yet. Those are currently blocked on some seemingly unrelated test failures I'm seeing locally

dekatzenel:
Juypter tested with backend changes. Works as expected once https://github.com/CitrineInformatics/platform-backend/pull/887 merges

tests/resources/test_material_template.py
```diff
@@ -0,0 +1,82 @@
...
+
+    # Then
+    assert 1 == session.num_calls
+    expected_call = FakeCall(
+        method="GET",
+        path="projects/{}/material-templates/{}/{}/material-specs".format(project_id, test_scope, test_id),
+        params={"forward": True, "ascending": True, "per_page": 20}
+    )
+    assert session.last_call == expected_call
+    assert 1 == len(specs)
```
kroenlein:
```suggestion
```
Why two tests?  How could it be other that 1 long if they are equal?

tests/resources/test_material_template.py
```diff
@@ -0,0 +1,82 @@
...
+    sample_run1_2 = MaterialRunDataFactory(spec=sample_spec1['uids'][test_scope])
+    sample_run2_2 = MaterialRunDataFactory(spec=sample_spec2['uids'][test_scope])
+    session.set_responses({key: [sample_spec1, sample_spec2]}, {key: [sample_run1_1, sample_run1_2]}, {key: [sample_run2_1, sample_run2_2]})
+
+    # When
+    runs = [run for run in collection.get_runs(test_scope, template_id, spec_collection, per_page=1)]
+
+    # Then
+    assert 3 == session.num_calls
+    assert 4 == len(runs)
```
kroenlein:
```suggestion
```
This is a little more meaningful than the len == 1 test, but still redundant.  Is the goal clarity in error conditions?

tests/resources/test_material_spec.py
```diff
@@ -0,0 +1,47 @@
...
+
+    # Then
+    assert 1 == session.num_calls
+    expected_call = FakeCall(
+        method="GET",
+        path="projects/{}/material-specs/{}/{}/material-runs".format(project_id, test_scope, test_id),
+        params={"forward": True, "ascending": True, "per_page": 20}
+    )
+    assert session.last_call == expected_call
+    assert 1 == len(runs)
```
kroenlein:
```suggestion
```
As per other comments.

src/citrine/resources/material_template.py
```diff
@@ -115,3 +116,48 @@ class MaterialTemplateCollection(DataConceptsCollection[MaterialTemplate]):
...
+        return self.session.cursor_paged_resource(self.session.get_resource,
+                                                  path,
+                                                  per_page=per_page,
+                                                  version="v1")
+
+    # Retrieve all material runs associated with a material template
+    def get_runs(self,
+                 template_scope: str,
+                 template_id: str,
+                 spec_collection: MaterialSpecCollection,
```
asantas93: I think it would be preferable to not make our users pass in `dataset.material_specs` / `project.material_specs` here -- it should be possible to just make the collection in the body like we do from the project https://github.com/CitrineInformatics/citrine-python/blob/30c6313329299fd38cf88d48bd0fd9e897d5f2ad/src/citrine/resources/project.py#L183

src/citrine/resources/material_template.py
```diff
@@ -115,3 +116,48 @@ class MaterialTemplateCollection(DataConceptsCollection[MaterialTemplate]):
...
+        path = self._get_path(ignore_dataset=True) + "/" + scope + "/" + id + "/material-specs"
+        return self.session.cursor_paged_resource(self.session.get_resource,
+                                                  path,
+                                                  per_page=per_page,
+                                                  version="v1")
+
+    # Retrieve all material runs associated with a material template
+    def get_runs(self,
+                 template_scope: str,
+                 template_id: str,
```
asantas93: Have you considered flipping which collection you put these methods on? I think it may be more intuitive to structure like
```python
my_project.material_runs.with_spec(...)
my_project.material_runs.with_template(...)
my_project.material_specs.with_template(...)
```
Opinion of @lkubie, @ventura-rivera, @Imperssonator, etc is probably more valuable here though.

lkubie: I really like the` with_...` access patterns as it parallels the other filtering access patterns. e.g.
`my_project.material_runs.filter_by_name("...")` @asantas93

src/citrine/resources/material_spec.py
```diff
@@ -87,3 +87,22 @@ class MaterialSpecCollection(DataConceptsCollection[MaterialSpec]):
     def get_type(cls) -> Type[MaterialSpec]:
         """Return the resource type in the collection."""
         return MaterialSpec
+
+    def get_runs(self, scope: str, id: str, per_page: int = 20) -> Iterator[dict]:
```
ventura-rivera: Not a requirement, but I'd suggest the input parameters be similar to the `get` method in data_concepts:
So `scope="id"` by default.

src/citrine/resources/material_template.py
```diff
@@ -115,3 +116,48 @@ class MaterialTemplateCollection(DataConceptsCollection[MaterialTemplate]):
     def get_type(cls) -> Type[MaterialTemplate]:
         """Return the resource type in the collection."""
         return MaterialTemplate
+
+    def get_specs(self, scope: str, id: str, per_page: int = 20) -> Iterator[dict]:
```
ventura-rivera: Same as above. Not a requirement, but I'd suggest the input parameters be similar to the `get` method in data_concepts:
So `scope="id"` by default.

tests/resources/test_material_template.py
dekatzenel: I think the original was copying from another test, but good call on the redundancy

src/citrine/resources/material_template.py
dekatzenel: The issue with doing it that way is that the endpoints won't match the endpoints on the backend. I can certainly switch the endpoints on the backend, but probably worth discussing

dekatzenel:
Validated by running jupyter notebook against local dev changes

tests/utils/session.py
```diff
@@ -94,6 +95,29 @@ def _get_response(self):
 
         return self.responses[0]
 
+    @staticmethod
+    def cursor_paged_resource(base_method: Callable[..., dict], path: str,
```
gregor-robinson: Nice!

src/citrine/resources/material_run.py
```diff
@@ -166,3 +167,54 @@ def get_history(self, scope, id) -> Type[MaterialRun]:
         model = loads(json.dumps([data['context'], data['root']], cls=TaurusEncoder))
         # Convert taurus objects into citrine-python objects
         return MaterialRun.build(model)
+
+    def with_spec(self,
```
gregor-robinson: There is precedent in citrine-python (e.g. https://github.com/CitrineInformatics/citrine-python/blob/391283f46bc8cead857df4a25c04b8b319a84e7f/src/citrine/resources/data_concepts.py#L540) to use the naming convention `filter_by_*` for methods like this. Thoughts on calling this `filter_by_spec`?

src/citrine/resources/material_run.py
```diff
@@ -166,3 +167,54 @@ def get_history(self, scope, id) -> Type[MaterialRun]:
...
+                                             self.dataset_id,
+                                             self.session)._get_path(ignore_dataset=True)
+        path = path_prefix + "/" + spec_scope + "/" + spec_id + "/material-runs"
+        return self.session.cursor_paged_resource(self.session.get_resource,
+                                                  path,
+                                                  per_page=per_page,
+                                                  version="v1")
+
+    # Retrieve all material runs associated with a material template
+    def with_template(self,
```
gregor-robinson: Thoughts on calling this `filter_by_template`?

src/citrine/resources/material_spec.py
```diff
@@ -87,3 +88,28 @@ class MaterialSpecCollection(DataConceptsCollection[MaterialSpec]):
     def get_type(cls) -> Type[MaterialSpec]:
         """Return the resource type in the collection."""
         return MaterialSpec
+
+    def with_template(self,
```
gregor-robinson: Thoughts on calling this `filter_by_template`?

src/citrine/resources/material_run.py
ventura-rivera: That's a solid point. I forgot about these. I am on the side of keeping this consistent by using filter_by_spec.

src/citrine/resources/material_run.py
ventura-rivera: I am on the side of keeping this consistent by using filter_by_template.

src/citrine/resources/material_spec.py
ventura-rivera: I am on the side of keeping this consistent by using filter_by_template.

tests/utils/session.py
dekatzenel: Thank Aris - I just copied this to the test code

PR #198, opened 2020-02-06
add docs for how to deactivate a module

kyle-laplante:
# Citrine Python PR

Screen Shot 2020-02-06 at 11 39 54 AM

### DIFF
--- a/docs/source/workflows/index.rst
+++ b/docs/source/workflows/index.rst
@@ -20,6 +20,19 @@ The modules are dictate how the platform utilizes research data to generate comp
 -  :doc:`Processors <processors>` define the method used to search the design space.
    The processor and design space are coupled: depending on the design space used, only a subset of processors are applicable.
 
+Activation
+**********
+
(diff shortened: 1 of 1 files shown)


docs/source/workflows/index.rst
```diff
@@ -20,6 +20,19 @@ The modules are dictate how the platform utilizes research data to generate comp
...
+
+Modules come `active` by default when created. If you would like to deactivate a module so it cannot be used again consider this example:
+
+.. code:: python
+
+   predictor = project.predictors.get(predictor_uuid)
+   predictor.active = False
+   project.predictors.update(predictor)
+
+Modules that are not `active` cannot be used in workflows and will not show up in search.
```
jtharris: Right now there isn't a search, right?  Maybe change this to "will not show up when listing modules" or something to that effect.

kyle-laplante: ah good catch. mentally i was thinking exactly that (the `.list()` method) but i always considered it the `search functionality`. will update

PR #202, opened 2020-02-07
Ekim/documentation suggestions

eddotman:
# Citrine Python PR

## Description 
Some quick suggested changes to the documentation:
+ Typo (?) in an example where a `MeanColumn` should be a `StdColumn` -- unless I misunderstood the example.
+ Change "outer product" to "Cartesian product" in explanation of `Product design space`, since I think this is the correct mathematical concept that describes what is happening.
+ Changed NDE to be NDME consistently
+ Changed description of high NDME to indicate high `error` and not necessarily high `uncertainty`.
+ Spell out more detail on the computation of support-weighted F1

### PR Type:

^^^N/A to these -- only changed documentation text.


docs/source/workflows/design_spaces.rst
```diff
@@ -10,11 +10,11 @@ Currently, there are two design spaces:
 Product design space
 --------------------
 
-Materials from a product design space are composed from the outer product of univariate dimensions.
+Materials from a product design space are composed from the Cartesian product of univariate dimensions.
```
bfolie: Why do you expect "Cartesian product" to be more intuitive to readers than "outer product?"

docs/source/workflows/performance_workflows.rst
```diff
@@ -51,20 +51,20 @@ For each test set, the rest of the data are used to train the predictor, and the
 By comparing the model's predictions to the true values, we can compute several performance metrics that provide information about model quality.
 
 - Root-mean squared error (RMSE): the square root of the average of the squared prediction error
-- Non-dimensional model error (NDME): The RMSE, normalized by the RMSE of a trivial model that always predicts the mean value of the data.
+- Non-dimensional model error (NDME): The RMSE, normalized by the RMSE of a trivial model that always predicts the mean value of the data (== standard deviation of the output variable).
```
bfolie:
```suggestion
- Non-dimensional model error (NDME): The RMSE, normalized by the standard deviation of the output variable (which is the RMSE of a trivial model that always predicts the mean value of the data).
```
Your addition made me think that we should just lead with the standard deviation. Feel free to reject if you think this is less legible

docs/source/workflows/performance_workflows.rst
```diff
@@ -82,7 +82,7 @@ For categorical responses, performance metrics include either the area under the
    If AUC=1.0, all samples are classified correctly.
    If AUC=0.5, the model cannot distinguish between the two categories.
    If AUC=0.0, all samples are classified incorrectly.
--  F1 score is calculated from precision and recall of the model, weighted by the number of true positives according to the formula ``2.0 * precision * recall / (precision + recall) * true_positives``.
+-  Support-weighted F1 score is calculated from averaged precision and recall of the model, weighted by the in-class fraction of true positives according to the formula ``2.0 * precision * recall / (precision + recall) * fraction_true_positives`` applied to each class.
```
bfolie:
```suggestion
-  Support-weighted F1 score is calculated from averaged precision and recall of the model, weighted by the in-class fraction of true positives according to the formula ``2.0 * precision * recall / (precision + recall) * fraction_true_positives`` summed over each class.
```

docs/source/workflows/design_spaces.rst
Imperssonator: I see that Cartesian product is more correct, but I had to look that up, and outer product was more immediately intuitive to me.

eddotman: I agree that outer product is more well known as a term. However, I think there is a risk of confusion: for both of these products, I'd expect people to intuit "I should get a matrix." 

But:

- For outer product, the elements of the matrix should be products themselves
- For Cartesian product, the elements of the matrix should be coordinates

Compromise: What if we said something like `outer (Cartesian) product`? Then it's hopefully clear precisely what we mean -- coordinates, and not multiplication -- without relying on the reader's knowledge of the less common product.

bfolie: When in doubt I think we should go for correctness. If the user is not familiar with the term "Cartesian product" then they will Google it and immediately get a box with a definition and a clear picture. If we make up/muddle terms then we risk more confusion in the long run.

Imperssonator: Yes, you guys are right. I realized I'm being hypocritical here having harped so much on "MLI" vs. "LI" - we should be correct first and it's partially our job to teach customers what these concepts mean if they're not intuitive.

docs/source/workflows/performance_workflows.rst
```diff
@@ -51,20 +51,20 @@ For each test set, the rest of the data are used to train the predictor, and the
...
+For numeric responses, performance metrics include root mean squared error (RMSE), non-dimensiona model error (NDME), standard residual and standard confidence.
 
 -  RMSE is a useful and popular statistical metric for model quality.
    Lower RMSE means the model is more accurate.
--  NDE is the ratio between RMSE and standard deviation of the output variable.
-   NDE is a useful non-dimensional model quality metric.
+-  NDME is the ratio between RMSE and standard deviation of the output variable.
+   NDME is a useful non-dimensional model quality metric.
    A value of NDE = 0 is a perfect model. If NDE = 1, then the model is uninformative.
    An acceptable NDE depends on how the model is used.
```
gregor-robinson:
```suggestion
   An acceptable NDME depends on how the model is used.
```

eddotman:
I think all suggestions have been added in now -- lmk if I missed anything!

mcgalcode:
cc: @eddotman changed base branch to master as part of the master flow update.

docs/source/workflows/performance_workflows.rst
```diff
@@ -51,22 +51,22 @@ For each test set, the rest of the data are used to train the predictor, and the
...
-   Generally, NDE > 0.9 indicates a model with very high uncertainty.
-   If 0.9 > NDE > 0.6, this model is typically a good candidate for a design workflow.
-   Lower values of NDE indicate increasingly accurate models.
+-  NDME is the ratio between RMSE and standard deviation of the output variable.
+   NDME is a useful non-dimensional model quality metric.
+   A value of NDME = 0 is a perfect model. If NDME = 1, then the model is uninformative.
+   An acceptable NDME depends on how the model is used.
+   Generally, NDME > 0.9 indicates a model with very high error.
+   If 0.9 > NDME > 0.6, this model is typically a good candidate for a design workflow.
+   Lower values of NDME indicate increasingly accurate models.
```
gregor-robinson: Here would be a nice place to explain what a lower value means as a candidate for a design workflow, continuing the thought from the previous sentence.

eddotman: Changed! lmk what you think in the new update

PR #203, opened 2020-02-07
Reproduction steps for flattening bug on taurus-extended objects

maxhutch:
There was an exception being thrown when flatten was called because of inconsistencies in the logic of `as_dict` and `build`.  This was resolvable in taurus, so we just pulled the new version here.


maxhutch:
I didn't mean to close this.

PR #216, opened 2020-02-13
PLA-2567 Add AraDefinitionCollection.register()

dekatzenel:
Covers both creating a new Ara Definition and updating an existing one

# Citrine Python PR

## Description 
Adds citrine-python method corresponding to the endpoints to create and update an AraDefinition

### PR Type:

### DIFF
--- a/src/citrine/resources/ara_definition.py
+++ b/src/citrine/resources/ara_definition.py
@@ -192,6 +192,29 @@ class AraDefinitionCollection(Collection[AraDefinition]):
             data = data[self._individual_key] if self._individual_key else data
             return self.build(data)
         else:
+            # Implement update as a part of register both because:
+            # 1) The validation requirements are the same for updating and registering an
+            #    AraDefinition
+            # 2) This prevents users from accidentally registering duplicate AraDefinitions
             data = self.session.put_resource(self._get_path(defn.definition_uid), body)
             data = data[self._individual_key] if self._individual_key else data
             return self.build(data)
+
+    def update(self, defn: AraDefinition) -> AraDefinition:
+        """
+        [ALPHA] Update an AraDefinition.
+
+        If the provided AraDefinition does have a uid, update (replace) the AraDefinition at that
+        uid with the provided AraDefinition.
+
+        Raise a ValueError if the provided AraDefinition does not have a definition_uid.
+
+        :param defn: The AraDefinition to updated
+        :return: The updated AraDefinition with updated metadata
+        """
+        if defn.definition_uid is None:
+            raise ValueError("Cannot update Ara Definition without a definition_uid."
+                             " Please either use register() to initially register this"
+                             " Ara Definition or retrieve the registered details before calling"
+                             " update()")
+        return self.register(defn)
--- a/src/citrine/resources/ara_definition.py
+++ b/src/citrine/resources/ara_definition.py
@@ -40,8 +40,9 @@ class AraDefinition(Resource["AraDefinition"]):
     def _get_dups(lst: List) -> List:
         return [x for x in lst if lst.count(x) > 1]
 
-    uid = properties.Optional(properties.UUID(), 'id')
-    version = properties.Optional(properties.Integer, 'version')
+    definition_uid = properties.Optional(properties.UUID(), 'definition_id')
+    version_uid = properties.Optional(properties.UUID(), 'id')
+    version_number = properties.Optional(properties.Integer, 'version_number')
     name = properties.String("name")
     description = properties.String("description")
     datasets = properties.List(properties.UUID, "datasets")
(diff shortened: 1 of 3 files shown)


dekatzenel:
@alexandraj777 for context

alexandraj777:
+1 to use GitHub labels to designate PR type, including things like `blocked` `needs db migration`, and `breaking change`

alexandraj777:
(I know this is probably just a template that's re-used in every PR, but wasn't sure where to comment)

alexandraj777:
Was there a specific area you wanted me to provide feedback on?

dekatzenel:
> Was there a specific area you wanted me to provide feedback on?

@maxhutch can probably explain this better, but my understanding is that we're providing a template for how to handle create/update for versioned items that Orion might want to take advantage of later

src/citrine/resources/ara_definition.py
```diff
@@ -126,20 +129,25 @@ class AraDefinitionCollection(Collection[AraDefinition]):
...
+    _individual_key = 'version'
 
     def __init__(self, project_id: UUID, session: Session):
         self.project_id = project_id
         self.session: Session = session
 
-    def get(self, uid: Union[UUID, str], version: int) -> AraDefinition:
-        """Get an Ara Definition."""
-        path = self._get_path(uid) + "/versions/{}".format(version)
+    def get_with_version(self, definition_uid: Union[UUID, str],
```
maxhutch: @alexandraj777 here is where we get a specific version of a versioned asset

src/citrine/resources/ara_definition.py
```diff
@@ -51,15 +52,17 @@ def _get_dups(lst: List) -> List:
 
     def __init__(self, *, name: str, description: str, datasets: List[UUID],
                  variables: List[Variable], rows: List[Row], columns: List[Column],
-                 uid: Optional[UUID] = None, version: Optional[int] = None):
+                 version_uid: Optional[UUID] = None, version_number: Optional[int] = None,
```
maxhutch: @alexandraj777 here is were we define the identifier/version system for versioned assets (like modules and workflows, in the future)

src/citrine/resources/ara_definition.py
```diff
@@ -161,3 +169,29 @@ def preview(self, defn: AraDefinition, preview_roots: List[LinkByUID]) -> dict:
             "rows": [x.as_dict() for x in preview_roots]
         }
         return self.session.post_resource(path, body)
+
+    def register(self, defn: AraDefinition) -> AraDefinition:
+        """
+        [ALPHA] Register an Ara Definition.
+
+        If the provided AraDefinition does not have a definition_uid, create a new element of the
```
maxhutch: @alexandraj777 here is where we're using the same `register` method for create and for update. This follows a pattern that's used elsewhere in citrine-python.  Update increments the version number and assigns a new version_uid (server-side).

src/citrine/resources/ara_definition.py
```diff
@@ -161,3 +169,29 @@ def preview(self, defn: AraDefinition, preview_roots: List[LinkByUID]) -> dict:
             "rows": [x.as_dict() for x in preview_roots]
         }
         return self.session.post_resource(path, body)
+
+    def register(self, defn: AraDefinition) -> AraDefinition:
```
maxhutch: @dekatzenel could you comment on the decision not to use `update` here?  Is it because of the lack of server-side update validation?

dekatzenel: Note: I don't have strong feelings here and can definitely switch it if requested. The reasoning I was using was based on:

1. Like you said, the validation requirements for an AraDefinition do not change between when it is registered vs updated, which means that there doesn't need to be a substantial UX difference between register & update
2. This prevents users from accidentally registering duplicate AraDefinitions when they modify an existing one, which is presumably a thing we want to do based on the design model (though I could also imagine wanting to facilitate leaving old ones alone)

When you said comment, you meant in the code, right? Commenting here first in case consensus is to switch to using update given my reasoning.

maxhutch: Both here and in code :-P.  I agree that we want to do (2), but it could be by raising a ValueError if the defn passed into `register` has a `definition_id` assigned already.

src/citrine/resources/ara_definition.py
alexandraj777: So, what's the serverside call? Is it the case that clientside create and update both call the same serverside register, or that clientside create calls serverside create, and clientside update calls serverside update, and serverside create and update call the same register? 

I ask because the former couples clientside architecture decisions tightly to serverside architecture decisions, and the latter allows flexibility for the server to change the implementation as needed.

dekatzenel:
Validated in jupyter notebook

PR #222, opened 2020-02-16
Citrine v0.8.0 is released!

maxhutch:
Citrine v0.8.0 is released!  This release includes significant improvements to the reconstruction of material histories and the serialization and deserialization of the client version of taurus objects.

## Improvements
 * Material histories will be reconstructed completely, including the ingredients of processes and measurements of materials (https://github.com/CitrineInformatics/citrine-python/pull/223)
 * [Taurus v0.6.0](https://github.com/CitrineInformatics/taurus/releases/tag/v0.6.0) is now included, providing enhanced flatten and json support (https://github.com/CitrineInformatics/citrine-python/pull/223)
 * Sphinx is now configured to inherit docs from parent classes (https://github.com/CitrineInformatics/citrine-python/pull/197)
 * Added documentation of the unit testing system this package uses (https://github.com/CitrineInformatics/citrine-python/pull/201)

## Fixes
 * Expression predictor documentation had reversed the order of the alias name and the full name (https://github.com/CitrineInformatics/citrine-python/pull/185)
 * The session is no longer added to data platform resources that might be json serialized (https://github.com/CitrineInformatics/citrine-python/pull/207)
 * The workflow execution controller extracts the right fields from the API response (https://github.com/CitrineInformatics/citrine-python/pull/221)

## Changes
 * `citrine.attributes` has been deprecated and will be removed in v0.9.0.  Attributes should be imported from `taurus.entity.attribute` instead (https://github.com/CitrineInformatics/citrine-python/pull/223)
 * Client support for the resources and components of the AI Engine have been marked as `[ALPHA]` to note that their APIs are unstable.  They will be omitted from future release notes until they reach beta.


requirements.txt
```diff
@@ -1,4 +1,4 @@
-taurus-citrine==0.4.0
+taurus-citrine==0.4.1
```
djack: Do we want 0.4.1 or 0.5.0?

maxhutch: We never actually re-applied the update to 0.5.0.

CONTRIBUTING.md
```diff
@@ -18,3 +18,18 @@ A passing build requires the following:
 * All tests pass
 * The linter finds no violations of PEP8 style
 * Every line of code is executed by a test (100% coverage)
+
+## Dependencies
```
djack: :+1:

PR #234, opened 2020-02-26
Adding Ara columns for Composition values

maxhutch:
# Citrine Python PR

## Description 
This PR adds four column definitions for composition-valued variables in Ara definitions.

### PR Type:

### DIFF
--- a/src/citrine/ara/columns.py
+++ b/src/citrine/ara/columns.py
@@ -10,7 +10,7 @@ from citrine._serialization import properties
 
 
 class CompositionSortOrder(BaseEnumeration):
-    """[ALPHA] Order to use when sorting the components in a composition
+    """[ALPHA] Order to use when sorting the components in a composition.
 
     * ALPHABETICAL is alpha-numeric order by the component name
     * QUANTITY is ordered from the largest to smallest quantity, with ties broken alphabetically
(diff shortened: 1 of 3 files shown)


src/citrine/ara/columns.py
```diff
@@ -194,6 +209,111 @@ def __init__(self, *,
...
+
+class NthBiggestComponentNameColumn(Serializable["NthBiggestComponentNameColumn"], Column):
+    """[ALPHA] Name of the Nth biggest component.
+
+    Parameters
+    ----------
+    data_source: str
+        name of the variable to use when populating the column
+    n: int
+        index of the component name to extract, starting with 1 for the biggest
```
bfolie: What happens if `n` is larger than the number of components? For `NthBiggestComponentQuantityColumn` I would expect to get a value of 0, but I do not know what to expect here.

src/citrine/ara/columns.py
```diff
@@ -194,6 +209,111 @@ def __init__(self, *,
...
+    data_source = properties.String('data_source')
+    n = properties.Integer("n")
+    typ = properties.String('type', default="biggest_component_name_column", deserializable=False)
+
+    def _attrs(self) -> List[str]:
+        return ["data_source", "n", "typ"]
+
+    def __init__(self, *,
+                 data_source: str,
+                 n: int):
```
bfolie: Should we set `n=1` as the default? I think it depends on how we expect people to use this column (will they almost always be grabbing the largest amount?)

maxhutch: Given that "N" is in the name, I'd say "no".

src/citrine/ara/columns.py
maxhutch: Its empty.  I'll clarify in docs.

PR #235, opened 2020-02-27
Only dataset-scoped data concepts collections can register

maxhutch:
This fixes an example in the docs that would actually throw an error.

### DIFF
--- a/docs/source/data_entry.rst
+++ b/docs/source/data_entry.rst
@@ -32,12 +32,15 @@ __ https://citrineinformatics.github.io/gemd-docs/specification/unique-identifie
 Registering Data Model Objects
 ---------------------------------
 
-Data model objects are created on the Citrine Platform through the ``register`` method present in each data model object collection.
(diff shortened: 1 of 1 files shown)


docs/source/data_entry.rst
```diff
@@ -32,12 +32,15 @@ __ https://citrineinformatics.github.io/gemd-docs/specification/unique-identifie
...
+Data model objects are created on the Citrine Platform through the ``register`` method present in each data model object collection that comes from a dataset.
 For example:
 
 .. code-block:: python
 
-    project.process_specs.register(ProcessSpec(...))
+    dataset.process_specs.register(ProcessSpec(...))
+
+Note that registration must be performed within the scope of a dataset: the dataset into which the objects are being written.
+The data model object collections that are defined with the project scope are read-only and will throw an error if their register method is called.
```
bfolie:
```suggestion
The data model object collections that are defined with the project scope (such as `project.process_specs`) are read-only and will throw an error if their register method is called.
```
This could be difficult for the reader to parse, so I suggest adding an example.

PR #236, opened 2020-02-27
Pla 3234/add all ingredients

bfolie:
# Citrine Python PR

## Description 
Write method that automatically adds columns for all ingredients in a process given by a specific process template.

### PR Type:

### DIFF
--- a/src/citrine/resources/ara_definition.py
+++ b/src/citrine/resources/ara_definition.py
@@ -150,7 +150,7 @@ class AraDefinition(Resource["AraDefinition"]):
             the scope for which to get ingredient ids (default is Citrine scope, 'id')
 
         """
-        quantity_dimension_display = {
+        dimension_display = {
             IngredientQuantityDimension.ABSOLUTE: "absolute quantity",
             IngredientQuantityDimension.MASS: "mass fraction",
             IngredientQuantityDimension.VOLUME: "volume fraction",
@@ -162,16 +162,16 @@ class AraDefinition(Resource["AraDefinition"]):
         new_columns = []
         for name in process.allowed_names:
             identifier_variable = IngredientIdentifierByProcessTemplateAndName(
-                name="{} of ingredient {} in process {}".format(scope, name, process.name),
-                headers=[process.name, name + ' ' + scope],
+                name='_'.join([process.name, name, str(hash(process_template.id + name + scope))]),
+                headers=[process.name, name, scope],
                 process_template=process_template,
                 ingredient_name=name,
                 scope=scope
             )
             quantity_variable = IngredientQuantityByProcessAndName(
-                name="{} of ingredient {} in process {}".format(
-                    quantity_dimension_display[quantity_dimension], name, process.name),
-                headers=[process.name, name, quantity_dimension_display[quantity_dimension]],
+                name='_'.join([process.name, name, str(hash(
+                    process_template.id + name + dimension_display[quantity_dimension]))]),
+                headers=[process.name, name, dimension_display[quantity_dimension]],
                 process_template=process_template,
                 ingredient_name=name,
                 quantity_dimension=quantity_dimension
--- a/src/citrine/resources/ara_definition.py
+++ b/src/citrine/resources/ara_definition.py
@@ -8,7 +8,6 @@ from citrine._rest.collection import Collection
 from citrine._rest.resource import Resource
 from citrine._serialization import properties
 from citrine._session import Session
-from citrine.resources.project import Project
 from citrine.resources.process_template import ProcessTemplate
 from citrine.ara.columns import Column, MeanColumn, IdentityColumn, OriginalUnitsColumn
 from citrine.ara.rows import Row
@@ -129,7 +128,7 @@ class AraDefinition(Resource["AraDefinition"]):
 
     def add_all_ingredients(self, *,
                             process_template: LinkByUID,
-                            project: Project,
+                            project,
                             quantity_dimension: IngredientQuantityDimension,
                             scope: str = 'id'
                             ):
@@ -170,16 +169,18 @@ class AraDefinition(Resource["AraDefinition"]):
                 scope=scope
             )
             quantity_variable = IngredientQuantityByProcessAndName(
-                name="{} of ingredient {} in process{}".format(
+                name="{} of ingredient {} in process {}".format(
                     quantity_dimension_display[quantity_dimension], name, process.name),
                 headers=[process.name, name, quantity_dimension_display[quantity_dimension]],
                 process_template=process_template,
                 ingredient_name=name,
                 quantity_dimension=quantity_dimension
             )
-            new_variables.extend([identifier_variable, quantity_variable])
 
-            new_columns.append(IdentityColumn(data_source=identifier_variable.name))
+            if identifier_variable.name not in [var.name for var in self.variables]:
+                new_variables.append(identifier_variable)
+                new_columns.append(IdentityColumn(data_source=identifier_variable.name))
+            new_variables.append(quantity_variable)
             new_columns.append(MeanColumn(data_source=quantity_variable.name))
             if quantity_dimension == IngredientQuantityDimension.ABSOLUTE:
                 new_columns.append(OriginalUnitsColumn(data_source=quantity_variable.name))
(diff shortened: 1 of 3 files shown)


src/citrine/resources/ara_definition.py
```diff
@@ -124,6 +126,74 @@ def add_columns(self, *,
...
+            IngredientQuantityDimension.VOLUME: "volume fraction",
+            IngredientQuantityDimension.NUMBER: "number fraction"
+        }
+        process: ProcessTemplate = project.process_templates.get(
+            uid=process_template.id, scope=process_template.scope)
+        new_variables = []
+        new_columns = []
+        for name in process.allowed_names:
+            identifier_variable = IngredientIdentifierByProcessTemplateAndName(
+                name="{} of ingredient {} in process {}".format(scope, name, process.name),
```
bfolie: I'm worried about accidental collisions with other names (either if the user adds them manually or if there is a different process template that happens to have the same name and includes an ingredient of the same name). Should we include a unique ID of some sort here, such as the id of the process template?

src/citrine/resources/ara_definition.py
```diff
@@ -124,6 +126,74 @@ def add_columns(self, *,
...
+                headers=[process.name, name + ' ' + scope],
+                process_template=process_template,
+                ingredient_name=name,
+                scope=scope
+            )
+            quantity_variable = IngredientQuantityByProcessAndName(
+                name="{} of ingredient {} in process {}".format(
+                    quantity_dimension_display[quantity_dimension], name, process.name),
+                headers=[process.name, name, quantity_dimension_display[quantity_dimension]],
```
bfolie: These and the headers above are editorial decisions. Maybe we should let the user provide their own headers? On the other hand they are using a convenience method, so reducing the number of decisions they have to make seems like it could be justified.

src/citrine/resources/ara_definition.py
```diff
@@ -124,6 +126,74 @@ def add_columns(self, *,
             columns=copy(self.columns) + columns
         )
 
+    def add_all_ingredients(self, *,
+                            process_template: LinkByUID,
+                            project,
+                            quantity_dimension: IngredientQuantityDimension,
+                            scope: str = 'id'
+                            ):
```
kroenlein: Are we saying that we only want to support complete lists of ingredients in the scenario that the Process Template has `allowed_names` set?

src/citrine/resources/ara_definition.py
kroenlein: `name` is never user visible.  It would make sense to use some kind of hash function or randomized string.

src/citrine/resources/ara_definition.py
kroenlein: If we're only using the convenience method route, then that sounds like an unnecessary option, at least in MVP.

src/citrine/resources/ara_definition.py
```diff
@@ -124,6 +126,74 @@ def add_columns(self, *,
...
+                quantity_dimension=quantity_dimension
+            )
+
+            if identifier_variable.name not in [var.name for var in self.variables]:
+                new_variables.append(identifier_variable)
+                new_columns.append(IdentityColumn(data_source=identifier_variable.name))
+            new_variables.append(quantity_variable)
+            new_columns.append(MeanColumn(data_source=quantity_variable.name))
+            if quantity_dimension == IngredientQuantityDimension.ABSOLUTE:
+                new_columns.append(OriginalUnitsColumn(data_source=quantity_variable.name))
```
kroenlein: We should consider if we want to create a mechanism for `automatically cast this into SI` rather than create the column.

bfolie: But even if we cast to SI, wouldn't it still be better from a user perspective to see the units next to the quantity?

bfolie:
> I'm not certain that this should be a convenience method as opposed to a variable definition that results in multiple variables.

What other columns might the user want to create from these variables?

src/citrine/resources/ara_definition.py
bfolie: That's my understanding.

The alternative would be to find the process based on this template _for each row_ of the table and aggregate the names of all the ingredients that feed into those processes. But that presents two problems:
1. I don't think the client can get those processes (though we could add an endpoint to do so)
2. What if the user then adds another row that has different ingredient names for the associated process?

We could also find every process in the project that references this process template, but then we could wind up with an obscenely long list of ingredients (and again, I don't think the client can perform that operation).

src/citrine/resources/ara_definition.py
bfolie: I want it to be deterministic because I want to prevent the user from adding the same column twice. I think a hash based on the ingredient and process name will work.

kroenlein: A hash based on the process UUID and the ingredient name would be nearly guaranteed unique.

src/citrine/resources/ara_definition.py
kroenlein: For the convenience method approach, you are right.  I was pushing back on that assumption.

src/citrine/resources/ara_definition.py
kroenlein: The question is if we feel confident that the user won't mix volume and mass in one column.  I'm not certain I feel confident on that, so your approach makes sense.  I'm just generally very against mixing units within one column.

kroenlein:
> > I'm not certain that this should be a convenience method as opposed to a variable definition that results in multiple variables.
> 
> What other columns might the user want to create from these variables?

What i mean is pass to the Ara Definition a Variable that is QuantityOfAllVariables, and have that transform into an ordered list of quantities.  That way, we don't need to know the full set of names in order to generate the necessary table.

src/citrine/resources/ara_definition.py
bfolie: Are you imagining a situation in which the user adds ingredients in an ABSOLUTE basis and some ingredient amounts are given in `mg` and some are in `ml`? And are you suggesting breaking that out into 2 columns, one for the values in units of mass and one for the values in units of volume?

src/citrine/resources/ara_definition.py
bfolie: Done. I also kept in the strings themselves, just in case someone in the future is debugging something on platform-backend and having some human-readable pieces of the variable names makes their lives easier.

bfolie:
> What i mean is pass to the Ara Definition a Variable that is QuantityOfAllVariables, and have that transform into an ordered list of quantities. That way, we don't need to know the full set of names in order to generate the necessary table.

Do you think that users won't be able to populate the `allowed_names` field in many cases? Because to me this (relying on `allowed_names`) feels like an expression of the "strongly encourage/basically require users to be thorough with their templates."

src/citrine/resources/ara_definition.py
kroenlein: That's kinda where my thoughts are.  We can save this for another PR.

maxhutch:
@bfolie : could we add documentation to https://github.com/CitrineInformatics/citrine-python/blob/develop/docs/source/data_extraction.rst in a follow-up PR?

src/citrine/resources/ara_definition.py
Imperssonator: The way that I did this for calcite was basically this:
"The alternative would be to find the process based on this template for each row of the table and aggregate the names of all the ingredients that feed into those processes."

But their complete ingredient set for the lighthouse project is a relatively limited set of ingredients. <20

src/citrine/resources/ara_definition.py
Imperssonator: This is a case where I would _hope_ the user uses the same quantity type for that ingredient across all processes (and this should be the case most of the time), but I would warn the user if that were not the case

bfolie: There are 2 distinct issues.
1. The user reports quantities with different dimensions (e.g. some things in mass fraction and others in volume fraction). I think this is what you are referring to Nils. In this case I think they would get a table with a bunch of NaN values in it, but no warning. I believe a warning would have to be generated in Faraday.
2. The user reports all quantities in the ABSOLUTE dimension but the units are of different types (e.g., mg for some and ml for others). I think that's less of a problem. Sometimes different ingredients are more naturally expressed in different quantities. Our formulations stack and optimizer should be able to handle it  (the "share in" descriptors won't be easily interpretable by a human, but it should still work)

PR #239, opened 2020-03-02
Bug/pla 2982 strip hints

crleblanc:
# Citrine Python PR

## Description 
This PR adds a 'build_py' custom builder, wrapping the existing run method with code that strips all of the type hints that are incompatible with Python 3.5 to enable support for Python 3.5+ with a single wheel package.  It also uses an older print syntax instead of the newer f-string.

I have replaced the strip_hints.sh script with a Python function in setup.py.  The script had some issues, if strip-hints was not installed it would wipe all of your source code and would not report an error.  Strip_hints is available as a Python module so it's easy to use from setup.py which will report an error if it's not installed.

Manual testing shows the type hints are removed in the wheel file and it imports cleanly in Python 3.5 (via Docker).  I've also run "python setup.py test" after pip-installing from requirements.txt and test_requirements.txt and all 5 tests pass.

I've updated the strip-hints dependency.  I can't find a changelog or release for this version, but a higher number must be better, right?  It produces the expected results so I think it's a safe change.

### PR Type:


setup.py
```diff
@@ -1,8 +1,30 @@
...
+
+STRIP_DIRS = ["src", "tests"]
+
+
+def strip_file(file_path):
+    from strip_hints import strip_file_to_string
+
+    if file_path.endswith('.py'):
+        stripped = strip_file_to_string(file_path, to_empty=True, only_assigns_and_defs=True)
+        print(file_path)
```
maxhutch: Is this a debugging print?

setup.py
```diff
@@ -1,8 +1,30 @@
...
+        with open(file_path, 'w') as file_obj:
+            file_obj.write(stripped)
+
+
+def strip_all_hints(dirnames):
+    print('stripping type annotations from: {}'.format(dirnames))
```
maxhutch: Is this a debugging print?

setup.py
```diff
@@ -39,14 +68,15 @@ def run(self):
           "requests>=2.22.0,<3",
           "pyjwt>=1.7.1,<2",
           "arrow>=0.15.4,<0.16",
-          "strip-hints>=0.1.5,<=0.1.7",
+          "strip-hints>=0.1.8",
```
maxhutch: Can we keep an upper bound here?

setup.py
crleblanc: I was implementing what the strip_hints.sh shell script had.  It prints out each file it processes.  I don't have a strong opinion on them so I'm happy to remove them.

setup.py
crleblanc: Same as above, it's doing the same operation as the shell script.

setup.py
crleblanc: Sure, or we could just lock it to 0.1.8.

setup.py
bfolie: The print statements were originally in there just to ensure the actions were being logged in the Travis output. I don't think it hurts to be verbose in this case 🤷‍♂

crleblanc: Usually I'm not a fan of too many print statements but in this case I think it's useful.  There aren't very many files and it's nice to see that it's running as expected.

PR #240, opened 2020-03-03
Make workflows processors optional

andrewmillspaugh:
Title


maxhutch:
Np if this is a follow-up, but can we be sure to update the docstring here: https://github.com/CitrineInformatics/citrine-python/pull/240/files#diff-752910cfb338d1869ea8b76e3fb540eeR52 ?

PR #245, opened 2020-03-09
David citrine test

anduill:
# Citrine Python PR

## Description 
Proposed changes for citrine-python client job submission, and getting job status.
JIRA ticket: https://citrine.atlassian.net/browse/PLA-2568


src/citrine/resources/ara_job.py
```diff
@@ -0,0 +1,179 @@
...
+
+        Parameters
+        ----------
+        project: Project
+            The project on behalf of which the job executes.
+        ara_def: AraDefinition
+            the ara definition describing the new table
+        """
+        job_id = uuid4()
+        url_suffix: str = "/ara-definitions{ara_definition}/versions/{version_number}/build?job_id={job_id}"
```
asantas93:
```suggestion
        url_suffix: str = "/ara-definitions/{ara_definition}/versions/{version_number}/build?job_id={job_id}"
```

src/citrine/resources/ara_job.py
```diff
@@ -0,0 +1,179 @@
...
+        if ara_def.version_number is None:
+            return ApiError(400, "Version number was not present request", None)
```
asantas93:
```suggestion
            return ApiError(400, "Version number was not present", None)
```
Did you mean to have 'request' in there?

src/citrine/resources/ara_job.py
```diff
@@ -0,0 +1,179 @@
...
+    session: Session
+        the remote session for hitting framework endpoints
+    """
+    _path_template = 'projects/{project_id}'
+
+    def __init__(self, session: Session):
+        self.session = session
+
+    def build_ara_table(self, project: Project,
+                        ara_def: AraDefinition) -> Union[ApiError, AraJobSubmissionResponse]:
```
asantas93: We don't typically actually return instances of `ApiError`. They're accessible in the `BadRequest` thrown by `checked_request` if the user want's to catch the exception and act based on the specific issue.

src/citrine/resources/ara_job.py
```diff
@@ -0,0 +1,179 @@
...
+                project_id=project.uid
+            ) + url_suffix.format(
+                ara_definition=ara_def.definition_uid,
+                version_number=ara_def.version_number,
+                job_id=job_id
+            )
+            response: dict = self.session.post_resource(path=path)
+            return AraJobSubmissionResponse.build(response)
+
+    def get_job_status(self, project: Project, job_id: str):
```
asantas93: This is probably fine for now, but we will at least also want this on a more generic job collection. Job status has no ara-specific requirements and can be checked regardless of the endpoint which launched it.

src/citrine/resources/ara_job.py
```diff
@@ -0,0 +1,179 @@
...
+            The job we retrieve the status for
+        """
+        url_suffix: str = "/execution/job-status?job_id={job_id}"
+        if project.uid is None:
+            return ApiError(400, "Project UUID was not present request", None)
+        else:
+            path: str = self._path_template.format(
+                project_id=project.uid
+            ) + url_suffix.format(job_id=job_id)
+            response: dict = self.session.post_resource(path=path)
```
asantas93: I would expect
```suggestion
            response: dict = self.session.get_resource(path=path)
```

tests/ara/test_job_client.py
```diff
@@ -0,0 +1,29 @@
+import pytest
+from citrine.resources.ara_job import TaskNode, JobStatusResponse
+
+
+def task_node_1() -> dict:
+    tn1 = {'id': 'dave_id1', 'task_type': 'dave_type', 'status': 'dave_status',
+                   'dependencies': {'dep1', 'dep2'}}
```
asantas93: I see at least one test failure is a result of you omitting a `failure_reason` of `None` from the dump

src/citrine/_serialization/properties.py
```diff
@@ -325,6 +325,41 @@ def _serialize(self, value: list) -> list:
         return serialized
 
 
+class Set(Property[set, typing.Iterable]):
```
asantas93: This seems cool, but a list would be fine if you end up running into any serde issues trying to support this

tests/ara/test_job_client.py
asantas93: See you fixed that. Looks like the remaining issues are all test coverage / flake8 appeasement

src/citrine/resources/ara_job.py
```diff
@@ -0,0 +1,179 @@
...
+    ----------
+    session: Session
+        the remote session for hitting framework endpoints
+    """
+    _path_template = 'projects/{project_id}'
+
+    def __init__(self, session: Session):
+        self.session = session
+
+    def build_ara_table(self, project: Project,
```
asantas93: I think I would expect this method to be a peer to https://github.com/CitrineInformatics/citrine-python/blob/master/src/citrine/resources/ara_definition.py#L231 rather than living in this new `AraJobFramework`. That will also save you the trouble of figuring out how the user is supposed to actually create an instance of this class, too. @maxhutch were you imagining that's where this method would go? What about the job status?

anduill:
gah, I forgot to test the endpoint calls

src/citrine/resources/ara_job.py
anduill: oops.  my bad.  I forgot to test this method anyway...so I didn't catch this

src/citrine/resources/ara_job.py
anduill: yes, that was typo.  I'll probably just remove these.  They look web-dev'y anyway

src/citrine/resources/ara_job.py
anduill: gotcha.

bfolie:
High level comments:
* Could you please include a description of the PR, preferably with a reference to the associated Jira ticket? It's difficult to review without that context
* There's a "Draft PR" option you can use for this kind of situation (to get a review, but it shouldn't be merged yet)
* As you've probably noticed, PRs are gated on 100% test coverage and on linting. The [.travis.yml file](https://github.com/CitrineInformatics/citrine-python/blob/master/.travis.yml) contains the relevant commands so you can run them locally and check for test coverage/linting

src/citrine/_serialization/properties.py
```diff
@@ -325,6 +325,41 @@ def _serialize(self, value: list) -> list:
...
+                         default)
+        self.element_type = element_type if isinstance(element_type, Property) else element_type()
+
+    @property
+    def underlying_types(self):
+        return set
+
+    @property
+    def serialized_types(self):
+        return typing.Iterable
```
bfolie: I think this has to be a list, since that's how json will store it.

src/citrine/resources/ara_definition.py
```diff
@@ -66,7 +66,7 @@ def __init__(self, *, name: str, description: str, datasets: List[UUID],
         self.version_number = version_number
         self.definition_uid = definition_uid
 
-        # Note that these validations only apply at construction time. The current intended usage
+            # Note that these validations only apply at construction time. The current intended usage
```
bfolie: ?

src/citrine/resources/ara_job.py
```diff
@@ -0,0 +1,179 @@
...
+    def serialized_types(self):
+        return dict
+
+    def _serialize(self, value: TaskNode) -> dict:
+        result = {"id": value.id, "task_type": value.task_type, "status": value.status,
+                  "failure_reason": value.failure_reason, "dependencies": value.dependencies}
+        return result
+
+    def _deserialize(self, value: dict) -> TaskNode:
+        return TaskNode.build(value)
```
bfolie: You shouldn't need to define a new `TaskNode` Property. By virtue of being a `Resource` it should do this automatically through the code in `properties.Object`

src/citrine/resources/ara_job.py
bfolie: On the surface this feels similar to the flow on the Orion side of the client--the user submits requests that take a long time and hence needs to be able to get an updated job status. On the backend the mechanics might be very different, but client-side can we use the same framework for both?

src/citrine/resources/ara_job.py
bfolie: I important consideration in this case is "what user actions might lead to this error and what can the user do about it?" It looks to me like this and the next error will only occur if the user creates an AraDefinition locally and tries to use it without registering it, since the act of registering will add a uid and a version number. We can therefore provide some sort of descriptive error message. It wouldn't be an ApiError--maybe a TypeError?

asantas93: `ValueError` seems reasonable for this

src/citrine/resources/ara_definition.py
anduill: oops.  I'll revert

src/citrine/resources/ara_job.py
anduill: it didn't work, but cause there are two paths for serialization (jobStatus has a list of tasknodes). I had to create the type to make it work

bfolie: You can chain them together to specify a list of TaskNodes. Example: https://github.com/CitrineInformatics/citrine-python/blob/master/src/citrine/ara/rows.py#L54

src/citrine/resources/ara_job.py
anduill: seems to me like we need two types.  One for the definition, and another for a registered definition.  I think the current type conflates these two ideas

anduill: `AraDefinition(name: str, description:str, ...)`

`RegisteredAraDefinition(ara_def: AraDefinition, definition_uid: UUID, version_uid: UUID...)`

anduill: This method would ONLY process Registered definitions.

dekatzenel: While I agree that having that distinction would be useful, it might be worth noting that the UI doesn't distinguish between registered and unregistered across the board, so that would be introducing a new concept/workflow to users when dealing with AraDefinitions

asantas93: @anduill personally I think I would take that a step further and just only let the SDK create an `AraDefinition` from the backend's response (`register` would just look a lot like the constructor), buuuut that's not the pattern we've been using for the most part, and I don't think working that out is In Scope for this PR. For now raising a `ValueError` when the critical attributes are missing should be fine. Or if that bothers you, could just require the id/version as method parameters directly, and let the user sort out tracking them down.

anduill: or, I can just throw an exception

asantas93: is that different from 'raising a `ValueError`'?

src/citrine/resources/ara_job.py
```diff
@@ -0,0 +1,171 @@
...
+
+        Parameters
+        ----------
+        project: Project
+            The project on behalf of which the job executes.
+        ara_def: AraDefinition
+            the ara definition describing the new table
+        """
+        job_id = uuid4()
+        url_suffix: str = "/ara-definitions/{ara_definition}/versions/{version_number}/build?job_id={job_id}"
```
asantas93: there are arguments to `post_resource` for specifying query params, e.g. https://github.com/CitrineInformatics/citrine-python/blob/master/src/citrine/resources/data_concepts.py#L389

src/citrine/_serialization/properties.py
anduill: I don't think it broke anything.  I left it as iterable because it could work with either sets or lists.  Do you anticipate any issues with this more general type?

asantas93: I would have expected you to have to use a `list` in `_serialize` in order to actually go from the output of `dump` to an actual JSON string. If it works though, whatever.

src/citrine/resources/ara_definition.py
```diff
@@ -228,6 +229,52 @@ def build(self, data: dict) -> AraDefinition:
...
+        Parameters
+        ----------
+        project: Project
+            The project on behalf of which the job executes.
+        ara_def: AraDefinition
+            the ara definition describing the new table
+
+        """
+        job_id = uuid4()
+        url_suffix: str = "/{ara_definition}/versions/{version_number}/build?job_id={job_id}"
```
asantas93: typically we use `params=...` in the call to `post_resource` instead of embedding them in the URL directly.

src/citrine/_serialization/properties.py
bfolie: Leaving it general here is fine. Sets are not formally part of json, and though python seems to be smart enough to work with it, I do think that the _serialize code should use a list instead of a set, if nothing else then for reasons of clarity

src/citrine/_serialization/properties.py
```diff
@@ -325,6 +325,41 @@ def _serialize(self, value: list) -> list:
...
+        deserialized = set()
+        for element in value:
+            deserialized.add(self.element_type.deserialize(element))
+        return deserialized
+
+    def _serialize(self, value: set) -> set:
+        serialized = set()
+        for element in value:
+            serialized.add(self.element_type.serialize(element))
+        return serialized
```
bfolie: I suggest changing the output from a set to a list, even if it works this way

PR #248, opened 2020-03-12
Pla 3324/component quantity columns

bfolie:
# Citrine Python PR

## Description 
Add  `normalize` option to `ComponentQuantityColumn` and `NthBiggestComponentQuantityColumn`, using False as the default.

I tested the change by registering an Ara Definition to develop and making sure that there was no error and that the result contained the expected values for the normalize field.

We still need a complete end-to-end test that builds a table using the relevant rows and columns, but I couldn't find anything to build off of (the tests in nextgen-devkit just test registering and updating, and I couldn't find an Ara demo in nextgen-devkit). Creating such an end-to-end test seems like an important task but it's out of scope and should be carded separately.

### PR Type:

### DIFF
--- a/tests/ara/test_columns.py
+++ b/tests/ara/test_columns.py
@@ -16,7 +16,7 @@ from citrine.ara.columns import MeanColumn, IdentityColumn, Column, StdDevColumn
     MostLikelyCategoryColumn(data_source="color"),
     MostLikelyProbabilityColumn(data_source="color"),
     FlatCompositionColumn(data_source="formula", sort_order=CompositionSortOrder.QUANTITY),
-    ComponentQuantityColumn(data_source="formula", component_name="Si"),
+    ComponentQuantityColumn(data_source="formula", component_name="Si", normalize=True),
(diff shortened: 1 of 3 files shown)


tests/ara/test_columns.py
```diff
@@ -16,7 +16,7 @@
     MostLikelyCategoryColumn(data_source="color"),
     MostLikelyProbabilityColumn(data_source="color"),
     FlatCompositionColumn(data_source="formula", sort_order=CompositionSortOrder.QUANTITY),
-    ComponentQuantityColumn(data_source="formula", component_name="Si"),
+    ComponentQuantityColumn(data_source="formula", component_name="Si", normalize=True),
```
kroenlein: This is a pretty spare test (API only).  Are there any consequences we can test here?

bfolie: As far as the client is concerned these objects only get (de)serialized. There's no other logic associated with them, hence nothing to test. What we really need is an end-to-end test in nextgen-devkit.

PR #251, opened 2020-03-12
Adding client support for AttributeInTrunk variable

maxhutch:
# Citrine Python PR

## Description 
Adds client support for the AttributeInTrunk variable definition

### PR Type:

### DIFF
--- a/src/citrine/ara/variables.py
+++ b/src/citrine/ara/variables.py
@@ -53,7 +53,7 @@ class Variable(PolymorphicSerializable['Variable']):
             RootInfo, AttributeByTemplate, AttributeByTemplateAfterProcessTemplate,
             AttributeByTemplateAndObjectTemplate, IngredientIdentifierByProcessTemplateAndName,
             IngredientLabelByProcessAndName, IngredientQuantityByProcessAndName,
-            RootIdentifier, AttributeInTrunk
+            RootIdentifier, AttributeInOutput
         ]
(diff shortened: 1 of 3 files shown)


maxhutch:
@ventura-rivera its before the process in the traversal order, but that is the reverse of chronological order, so I could see that being confusing.  If you provide a "mixing process" template, it will get the attributes from *later in the history* than that template.

I agree that the name is not great.  Do you have any other ideas?

ventura-rivera:
@maxhutch Yes, I see what you are saying. That's really tricky. Another potential name that comes to mind is `AttributeInOutput` or `AttributeInProcessOutput`

maxhutch:
I'll switch to AttributeInOutput for now and we can see how that goes.

src/citrine/ara/variables.py
```diff
@@ -396,3 +398,70 @@ def __init__(self, *,
...
+
+class AttributeInOutput(
+        Serializable['AttributeInOutput'], Variable):
+    """[ALPHA] Attribute marked by an attribute template in the trunk of the history tree.
+
+    The search for an attribute that marks the given attribute template starts at the root
+    of the material history tree and proceeds until any of the given process templates are reached.
+    Those templates block the search from continuing into their ingredients but do not halt the
+    search entirely. This variable definition allows attributes that are present both in output
+    and the inputs of a process cto be distinguished.
```
bfolie:
```suggestion
    and the inputs of a process to be distinguished.
```

PR #252, opened 2020-03-12
Pla 3366/embedded predictors in graph

bfolie:
# Citrine Python PR

## Description 
Allows for embedded predictors in graph.
* GraphPredictor `predictors` field now accepts a list of UUIDs, other predictors, or a mixture of the 2
* Created `Union` Property to allow for this "UUID or predictor" logic
* Renamed `MixedList` Property to `SpecifiedMixedList` to clarify that each entry has a specified type ("MixedList" sounds like its List[Union[...]], which it isn't)
* Update GraphPredictor serde test to include an embedded expression predictor.

Follow-on PR in orion-documentation will reflect these changes

### PR Type:

### DIFF
--- a/src/citrine/informatics/predictors.py
+++ b/src/citrine/informatics/predictors.py
@@ -176,6 +176,23 @@ class GraphPredictor(Serializable['GraphPredictor'], Predictor):
     def _post_dump(self, data: dict) -> dict:
         data['display_name'] = data['config']['name']
         return data
+    
+    @classmethod
+    def _pre_build(cls, data: dict) -> dict:
+        for i, predictor in enumerate(data['config']['predictors']):
+            if isinstance(predictor, dict):
+                data['config']['predictors'][i] = GraphPredictor.stuff_predictor_into_envelope(predictor)
+        return data
+
+    @staticmethod
+    def stuff_predictor_into_envelope(predictor: dict) -> dict:
+        """Insert a serialized embedded predictor into a module envelope, to facilitate deser."""
+        return dict(
+            module_type='PREDICTOR',
+            config=predictor,
+            status='EMBEDDED',
+            active=False
+        )
 
     def __str__(self):
         return '<GraphPredictor {!r}>'.format(self.name)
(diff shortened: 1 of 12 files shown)


src/citrine/informatics/predictors.py
```diff
@@ -159,21 +162,43 @@ class GraphPredictor(Serializable['GraphPredictor'], Predictor):
...
         return data
 
+    @staticmethod
+    def stuff_predictor_into_envelope(predictor: dict) -> dict:
+        """Insert a serialized embedded predictor into a module envelope, to facilitate deser."""
+        return dict(
+            module_type='PREDICTOR',
+            config=predictor,
+            active=False,
+            schema_id='43c61ad4-7e33-45d0-a3de-504acb4e0737'  # TODO: what should this be?
```
bfolie: I don't understand the function of schema_id, so I'm not sure what to do about this. Right now it's required for deserializing all predictors. Options:
1. Make it optional for predictors except the graph predictor (like we do in this PR for `status`)
2. Insert an arbitrary value like is done here (this is the schema_id for graph predictors, so probably wouldn't want to use that)
3. Check to see what type of predictor this is and insert the relevant schema_id

src/citrine/informatics/predictors.py
```diff
@@ -81,7 +81,7 @@ class SimpleMLPredictor(Serializable['SimplePredictor'], Predictor):
     latent_variables = properties.List(properties.Object(Descriptor), 'config.latent_variables')
     training_data = properties.Object(DataTable, 'config.training_data')
     typ = properties.String('config.type', default='Simple', deserializable=False)
-    status = properties.String('status', serializable=False)
+    status = properties.Optional(properties.String(), 'status', serializable=False)
```
bfolie: I made `status` optional for all non-graph predictors so that they could be embedded and not have a status. We could also give all embedded predictors a status like "EMBEDDED". I'm not sure what would make more sense from the user's point of view

PR #253, opened 2020-03-12
Bug/pla 3008 notebook connection reset error

crleblanc:
# Citrine Python PR

## Description 
This PR is a proposed bugfix for https://citrine.atlassian.net/browse/PLA-3008.  If a ConnectionResetError or ConnectionError exception is raised while running a request, this will reinitialize the session and try the request again.  This does not use recursion.

I view this as a low risk PR as there is no substantial change to the logic.

### PR Type:

### DIFF
--- a/src/citrine/_session.py
+++ b/src/citrine/_session.py
@@ -82,7 +82,7 @@ class Session(requests.Session):
         try:
             response = super().request(method, uri, **kwargs)
         except (ConnectionError, ConnectionResetError):
-            logger.debug('Connection Reset, creating a new session')
+            logger.debug('Connection Error, creating a new session')
             super().__init__()
             response = super().request(method, uri, **kwargs)
 
--- a/src/citrine/_session.py
+++ b/src/citrine/_session.py
@@ -81,7 +81,7 @@ class Session(requests.Session):
 
         try:
             response = super().request(method, uri, **kwargs)
-        except ConnectionResetError:
+        except (ConnectionError, ConnectionResetError):
             logger.debug('Connection Reset, creating a new session')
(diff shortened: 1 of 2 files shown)


src/citrine/_session.py
```diff
@@ -79,7 +79,12 @@ def checked_request(self, method: str, path: str,
             logger.debug('\t{}: {}'.format(k, v))
         logger.debug('END request details.')
 
-        response = super().request(method, uri, **kwargs)
+        try:
+            response = super().request(method, uri, **kwargs)
+        except (ConnectionError, ConnectionResetError):
+            logger.debug('Connection Error, creating a new session')
+            super().__init__()
```
mcgalcode: What exactly does this do? Does this reset the state of the citrine session? How are the parameters that were used to instantiate the citrine session (e.g. scheme, host, api key) persisted?

crleblanc: The citrine-python Session is a subclass of Requests.Session.  The CP Session.__init__() calls the superclass's (Requests.Session's) __init__() with no arguments.  I had a look as Request's Session.__init__() and this is where urllib3 connections are created.

All of the arguments passed to CP's Session.__init__() are left unchanged so they should persist as normal.  I'm not sure which attributes are used by the superclass, e.g. refresh_token.

This PR doesn't try to refresh the token if one of these errors are seen.  This is something we can look into if required.

This change has not been tested with Jupyter notebooks to see if this addresses the possible timeout issue.

mcgalcode: Got it. This sounds great to me. I'll merge.

PR #258, opened 2020-03-13
Handle non-json responses to REST calls without erroring

dekatzenel:
# Citrine Python PR

## Description 
Handle non-json or invalid json responses from the backend without erroring, e.g. status code only with no payload

### PR Type:

### DIFF
--- a/src/citrine/_session.py
+++ b/src/citrine/_session.py
@@ -5,6 +5,8 @@ from datetime import datetime, timedelta
 
 from requests import Response
 
+from json.decoder import JSONDecodeError
+
 from citrine.exceptions import (
     NotFound,
     Unauthorized,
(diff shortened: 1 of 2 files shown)


dleblanc:
So instead of always returning JSON from the backend we're changing citrine-python to be more lenient in what it accepts?

dekatzenel:
> So instead of always returning JSON from the backend we're changing citrine-python to be more lenient in what it accepts?

They're not mutually exclusive, but I think we want this for a couple of reasons:

1. The content-type fix isn't sufficient; citrine-python can't parse the empty string, and figuring out what is the right thing to make the backend return is a harder problem than making the frontend not error when it doesn't get json
2. As I pointed out yesterday, there are some places in the code where we have less control over the response because it's being fully handled by akka http methods that we're calling

PR #261, opened 2020-03-16
Refactoring the retry logic to use a urllib3 Retry and HTTPAdapter

crleblanc:
This gives us fine grained control over the retry logic and backoff params.
We also control which HTTP status codes trigger a retry, including CloudFlare
specific codes.  This also switches from using super().request() to self.request()
for simplicity.

# Citrine Python PR

## Description 

The goal of this PR is to use a standard approach to HTTP retries that is simple and robust.  This approach uses the one suggested by the Request docs.  This is the approach used by the Pip package management tool which handles timeouts and other errors gracefully.

I'm happy to discuss and modify the HTTP codes we wish to retry on, the number of retries and the backoff params.  I picked arbitrary parameters based on other examples I saw.  This should cover just about everything we are likely to encounter.

This will work for all relevant requests made using the Citrine-python Session class (get/put/post/etc).

I removed a couple unit tests since the retry logic is now outside of citrine-python.

### PR Type:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,3 +8,4 @@ sphinxcontrib-apidoc==0.3.0
 boto3==1.9.226
 botocore==1.12.226
 deprecation==2.0.7
+urllib3==1.25.7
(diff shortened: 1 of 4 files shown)


crleblanc:
I tried running this code in a Jupyter notebook, leaving it idle for an hour.  I was able to reuse the session instance for new requests.

I also ran this notebook without a network connection, reconnecting a few minutes after running a call to client.projects.list().  This call worked successfully so I think this retry logic is working as expected.

PR #266, opened 2020-03-24
Adding "DataSource" non-module objects, that serve as training data

maxhutch:
# Citrine Python PR

## Description 
The DataTable class is replaced by the AraTableDataSource class.
The new CSVDataSource is added to support pla-3485.

This is an implementation of the [DataSources design](https://github.com/CitrineInformatics/ai-engine-documentation/blob/master/docs/data-sources.md#data-sources) and provides client support for the backend functionality found in https://github.com/CitrineInformatics/platform-backend/pull/1049

### PR Type:


maxhutch:
@Imperssonator note that this will be a breaking change when deployed.  `DataTable` instances will need to be replaced with `AraTableDataSource` instances, which are very similar but use integers for the version numbers and don't support `latest`.

setup.py
```diff
@@ -57,7 +57,7 @@ def run(self):
 
 
 setup(name='citrine',
-      version='0.9.0',
+      version='0.9.1',
```
maxhutch: Patch since the changes are only to ALPHA functionality.

src/citrine/informatics/data_sources.py
```diff
@@ -0,0 +1,99 @@
...
+    Parameters
+    ----------
+    file_link: FileLink
+        link to the CSV file to read the data from
+    column_definitions: Mapping[str, Descriptor]
+        Map the column headers to the descriptors that will be used to interpret the cell contents
+
+    """
+
+    type = properties.String('type', default='csv_data_source', deserializable=False)
```
maxhutch: @talltea note

src/citrine/informatics/data_sources.py
```diff
@@ -0,0 +1,99 @@
...
+    Parameters
+    ----------
+    table_id: UUID
+        Unique identifier for the table
+    table_version: int
+        Version number for the table
+
+    """
+
+    type = properties.String('type', default='hosted_table_data_source', deserializable=False)
```
maxhutch: @talltea note

src/citrine/informatics/data_sources.py
andrewmillspaugh: Haven't we been naming this variable `typ`?

maxhutch: I copied the source file from `descriptors.py` as a starting point, and [that used `type`](https://github.com/CitrineInformatics/citrine-python/blob/master/src/citrine/informatics/descriptors.py#L44).  I actually prefer `typ`; will change.

maxhutch: It looks like the rest of the `informatics` module uses `typ` too, so I'm going to change it in `descriptors.py` too.

tests/conftest.py
```diff
@@ -106,7 +107,8 @@ def valid_simple_ml_predictor_data():
             latent_variables=[y.dump()],
             training_data=dict(
                 table_id='e5c51369-8e71-4ec6-b027-1f92bdc14762',
-                table_version='latest'
+                table_version=0,
```
andrewmillspaugh: @maxhutch can we expect that version numbers start at 0?

andrewmillspaugh: Also it seems like "latest" is supported, at least in the api.

maxhutch: Good call-out for the version numbers. I'll check and add to the docstring too.

The story ACs include: "The predictor module references a specific version of the file in the data service." and its also in the ai-engine-documentation and was discussed a little here: https://github.com/CitrineInformatics/ai-engine-documentation/pull/43#issuecomment-602800680

I think it would be easy enough to accept a `str` or `int` so the constructor signature doesn't change in the future if we want to support "latest" (and also we would not throw a client error on "latest").  But, for product reasons, I don't think we want to persist "latest" in the predictor configuration; we'd want to convert it to the explicit latest version number.

maxhutch:
> This will be a breaking change, right? Is there a straightforward (relatively cheap) way to prevent this break?

Yes, but I think there probably is a way to make it not breaking.  Let me take a crack at it.

maxhutch:
> > This will be a breaking change, right? Is there a straightforward (relatively cheap) way to prevent this break?
> 
> Yes, but I think there probably is a way to make it not breaking. Let me take a crack at it.

Nevermind.

maxhutch:
https://github.com/CitrineInformatics/nextgen-devkit/pull/366 should unblock this.

PR #271, opened 2020-04-01
Add method for validate-templates endpoint

dekatzenel:
# Citrine Python PR

## Description 
Add method to hit backend endpoint /validate-templates

### PR Type:

### DIFF
--- a/src/citrine/resources/data_objects.py
+++ b/src/citrine/resources/data_objects.py
@@ -149,7 +149,7 @@ class DataObjectCollection(DataConceptsCollection[DataObjectResourceType], ABC):
                            ingredient_process_template: Optional[ProcessTemplate] = None)\
             -> List[ValidationError]:
         """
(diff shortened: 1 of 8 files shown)


dekatzenel:
> this seems like a simple PR. I don't see anything glaring. This just hits the endpoint for validation right? I don't think I saw anything else.

There's also a couple of small tweaks to how testing works more broadly in session.py

dekatzenel:
> please mark this as alpha. Afterwards, I, David Garcia, will approve this Pull Request.

Upon reflection, this is beta because the core backend business logic is already established and well-tested

bfolie:
> Upon reflection, this is beta because the core backend business logic is already established and well-tested

I don't think the [Beta] tag is being used in citrine-python (since it's 0.x.x, isn't everything beta?). But I agree it's not alpha, since unlike Ara we don't expect it to be volatile.

PR #272, opened 2020-04-01
[PLA-3619] Feature/molecular structure descriptor

sparadiso:
# Citrine Python PR

## Description 
This PR just adds a MolecularStructureDescriptor for use in DataSource and AI Engine module instances.

### PR Type:

### DIFF
--- a/src/citrine/informatics/descriptors.py
+++ b/src/citrine/informatics/descriptors.py
@@ -101,6 +101,7 @@ class MolecularStructureDescriptor(Serializable['MolecularStructureDescriptor'],
     ----------
     key: str
         The column header key corresponding to this descriptor
+
     """
 
     key = properties.String('descriptor_key')
--- a/src/citrine/informatics/descriptors.py
+++ b/src/citrine/informatics/descriptors.py
@@ -96,7 +96,8 @@ class InorganicDescriptor(Serializable['InorganicDescriptor'], Descriptor):
 
 class MolecularStructureDescriptor(Serializable['MolecularStructureDescriptor'], Descriptor):
     """
-    [ALPHA] Material descriptor for an organic molecule (accepts SMILES, IUPAC, and InChI String values).
+    [ALPHA] Material descriptor for an organic molecule
+    (accepts SMILES, IUPAC, and InChI String values).
 
     Parameters
     ----------
--- a/src/citrine/informatics/descriptors.py
+++ b/src/citrine/informatics/descriptors.py
@@ -96,7 +96,7 @@ class InorganicDescriptor(Serializable['InorganicDescriptor'], Descriptor):
 
 class MolecularStructureDescriptor(Serializable['MolecularStructureDescriptor'], Descriptor):
     """
-    [ALPHA] Material descriptor for an organic molecule (accepts SMILES and InChI String values).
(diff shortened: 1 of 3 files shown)


src/citrine/informatics/descriptors.py
```diff
@@ -94,6 +94,33 @@ def __init__(self, key: str, threshold: float = 1.0):
         self.threshold = threshold
 
 
+class MolecularStructureDescriptor(Serializable['MolecularStructureDescriptor'], Descriptor):
+    """
+    [ALPHA] Material descriptor for an organic molecule (accepts SMILES and InChI String values).
```
bfolie, line 99: According to Mithril we also support IUPAC names

alexandraj777:
What are the checkboxes for?

alexandraj777:
Is this PR documentation of a previously existing feature, or solution design, or something else?

src/citrine/informatics/descriptors.py
```diff
@@ -19,7 +18,8 @@ def get_type(cls, data) -> Type[Serializable]:
         return {
             "Real": RealDescriptor,
             "Inorganic": InorganicDescriptor,
-            "Categorical": CategoricalDescriptor
+            "Categorical": CategoricalDescriptor,
+            "Organic": MolecularStructureDescriptor
```
alexandraj777, line 22: Adding a hanging comma onto the end of all multiline lists will help reduce spurious lines from your PRs. Specifically, here, you didn't really edit line 21, yet it shows up in the PR because you needed to add a comma so that you could add line 22. If you add a comma to the end of line 22 now, you'll avoid issues like this in the future

src/citrine/informatics/descriptors.py
```diff
@@ -94,6 +94,33 @@ def __init__(self, key: str, threshold: float = 1.0):
...
+    key = properties.String('descriptor_key')
+    typ = properties.String('type', default='Organic', deserializable=False)
+
+    def __eq__(self, other):
+        try:
+            attrs = ["key", "typ"]
+            return all([
+                self.__getattribute__(key) == other.__getattribute__(key) for key in attrs
+            ])
+        except Exception:
```
alexandraj777, line 117: Could you be more specific as to the exception? Allowing for any Exception, and not logging it, might cover up unintended developer error in the future

src/citrine/informatics/descriptors.py
sparadiso: Good call, thanks!

src/citrine/informatics/descriptors.py
sparadiso: Ah, yes. Thank you. I will add that.

src/citrine/informatics/descriptors.py
sparadiso: Good catch - this was a lazy copy/paste from above. I'll take this opportunity to try and infer the original intent here.

sparadiso:
> Is this PR documentation of a previously existing feature, or solution design, or something else?
Sorry, @alexandraj777, I missed this. This adding a class to represent a configuration object that already exists in the platform. We shouldn't release it until it's been incorporated into end to end tests, but it is expected to be valid input for modules without modification.

maxhutch:
> We shouldn't release it until it's been incorporated into end to end tests, but it is expected to be valid input for modules without modification.

Should we hold back on merging this, then?  Otherwise, its going to block a release...

PR #274, opened 2020-04-02
Add Dataset.register_all()

dekatzenel:
# Citrine Python PR

## Description 
Add Dataset.register_all() convenience method. Expected to be used for seeding.

### PR Type:

### DIFF
--- a/src/citrine/resources/dataset.py
+++ b/src/citrine/resources/dataset.py
@@ -12,6 +12,7 @@ from taurus.entity.template.measurement_template import MeasurementTemplate
 from taurus.entity.template.parameter_template import ParameterTemplate
 from taurus.entity.template.process_template import ProcessTemplate
 from taurus.entity.template.property_template import PropertyTemplate
+from taurus.util import writable_sort_order
 
 from citrine._session import Session
 from citrine._rest.collection import Collection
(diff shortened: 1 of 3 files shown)


src/citrine/resources/dataset.py
```diff
@@ -221,6 +221,16 @@ def register(self, data_concepts_resource: ResourceType) -> ResourceType:
...
 
+    def register_all(self, data_concepts_resources: List[ResourceType]) -> List[ResourceType]:
+        """
+        Register multiple data concepts resources to each of their appropriate collections.
+
+        :param data_concepts_resources: the resources to register. Can be different types.
+
+        :return the registered versions
+        """
+        return [self.register(resource) for resource in data_concepts_resources]
```
ventura-rivera, line 232: I believe you will run into a problem here: For example, consider a MaterialSpec linked to a ProcessSpec. Both of the items are in the `data_concepts_resources` list, but you will get an error if MaterialSpec is registered before ProcessSpec.

In devkit, I filtered by each type, then registered in an order so I don't get an error. (See [this function](https://github.com/CitrineInformatics/nextgen-devkit/blob/master/seeding/register.py#L19))

src/citrine/resources/dataset.py
```diff
@@ -221,6 +222,58 @@ def register(self, data_concepts_resource: ResourceType) -> ResourceType:
         if isinstance(data_concepts_resource, ConditionTemplate):
             return self.condition_templates.register(data_concepts_resource)
 
+    def register_all(self, data_concepts_resources: List[ResourceType]) -> List[ResourceType]:
+        """
+        Register multiple data concepts resources to each of their appropriate collections.
+
+        Does so in an order that is guaranteed to store all linked items before the item that
```
maxhutch, line 229: Why not use [taurus.util.writeable_sort_order](https://github.com/CitrineInformatics/gemd-python/blob/master/gemd/util/impl.py#L240) (note that it is still available under the taurus import; the source just moved to the `gemd` tree.  It will have the same function as below, but more DRY.  It should also make test coverage easier :-D

ventura-rivera: I had no idea this was a thing!

PR #276, opened 2020-04-02
Feature/mandatory version bumps

dekatzenel:
# Citrine Python PR

## Description 
Test if trying to merge into branches that aren't master works

DO NOT MERGE

### PR Type:


maxhutch:
@dekatzenel : given that we're releasing on tags rather than master, I don't think there is anything special about the builds on branches anymore.  We have "Require branches to be up to date before merging" enabled in the repo settings, so there shouldn't be any difference in the code being tested.  If you remove the branch test, I will still approve this PR.

dekatzenel:
This is the wrong PR - it was a sanity check that things still worked when you pushed against a branch other than master

PR #278, opened 2020-04-03
Add dataset to data concepts classes

asantas93:
# Citrine Python PR

## Description 
Readies client for new `dataset` field which will appear on all data concepts classes, if they were created from the backend's response.

### PR Type:

### DIFF
--- a/src/citrine/resources/data_concepts.py
+++ b/src/citrine/resources/data_concepts.py
@@ -49,7 +49,10 @@ class DataConcepts(PolymorphicSerializable['DataConcepts'], DictSerializable, AB
     Custom json support object, which knows how to serialize and deserialize DataConcepts classes.
     """
 
-    client_specific_fields = {"audit_info": AuditInfo}
+    client_specific_fields = {
+        "audit_info": AuditInfo,
+        "dataset": UUID,
+    }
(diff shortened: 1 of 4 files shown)


src/citrine/resources/data_concepts.py
```diff
@@ -85,13 +95,21 @@ def from_dict(cls, d: dict):
...
-                setattr(obj, "_{}".format(field), clazz.build(popped[field]))
+            value = popped[field]
+            if value is None:
+                deserialized = None
+            elif issubclass(clazz, DictSerializable):
+                if not isinstance(value, dict):
+                    raise TypeError(
+                        "{} must be a dictionary or None but was {}".format(field, value))
+                deserialized = clazz.build(value)
+            elif issubclass(clazz, Property):
```
maxhutch, line 106: To avoid the namespace collision with the property attribute, could we use this as `properties.Property` (as we have in some other places?)

asantas93: That is actually a python syntax error, surprisingly. I could give it an alias though.

maxhutch: anything to disambiguate it

PR #285, opened 2020-04-07
Add seeding utility for sorting templating from data objects

dekatzenel:
# Citrine Python PR

## Description 
Add seeding utility for sorting templating from data objects. Useful when seeding needs to be different for templates vs data objects

### PR Type:

### DIFF
--- a/src/citrine/seeding/sort_gems.py
+++ b/src/citrine/seeding/sort_gems.py
@@ -2,7 +2,7 @@ from taurus.entity.template.attribute_template import AttributeTemplate
 from taurus.entity.template.base_template import BaseTemplate
 
 
-def sort_templates_from_objects(gems):
+def split_templates_from_objects(gems):
     """
     Sort the provided gems into two lists.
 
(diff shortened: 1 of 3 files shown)


src/citrine/seeding/sort_gems.py
```diff
@@ -0,0 +1,22 @@
...
+    The second list contains only data objects (specs and runs).
+
+    Useful when the procedure for seeding templates needs to be distinct from the procedure for
+    seeding data objects.
+    """
+    templates = []
+    data_objects = []
+    for gem in gems:
+        if isinstance(gem, AttributeTemplate) or isinstance(gem, BaseTemplate):
+            templates.append(gem)
```
ventura-rivera, line 19: I am not sure if I caught this in the devkit pr, but attribute_templates will have to be registered before object templates so you might have to sort this differently. Does that sorting matter here? or is the main purpose to separate templates from other objects?

dekatzenel: My idea was that the registration code would handle the sorting (which it currently does). I can add a comment to that effect if that would be helpful

ventura-rivera: Ok, if thats the case, I think this is fine by me. A comment would be really helpful. But one last question: would it make more sense to rename this (because it is splitting/separating templates rather than sorting)? I don't have a critical issue with the name, but that would've helped me understand the intention better.

PR #288, opened 2020-04-08
Pla 3725/rename inorganic chemical descriptor

bfolie:
# Citrine Python PR

## Description 
Renames InorganicDescriptor to ChemicalFormulaDescriptor and removes the `threshold` argument from the constructor. The backend implementation still has a threshold parameter so we need to take it into account during serde, but it's less exposed to the user now.

If a user tries to create an `InorganicDescriptor` they will instead get a `ChemicalFormulaDescriptor` and they will see a deprecation warning.

Corresponding changes to nextgen-devkit (merged): https://github.com/CitrineInformatics/nextgen-devkit/pull/414/files#diff-b9e1cad38f41a2a025680df6f77d2aacR6-R9

### PR Type:

### DIFF
--- a/src/citrine/informatics/descriptors.py
+++ b/src/citrine/informatics/descriptors.py
@@ -21,7 +21,7 @@ class Descriptor(PolymorphicSerializable['Descriptor']):
         return {
             "Real": RealDescriptor,
             "Inorganic": ChemicalFormulaDescriptor,
-            "Categorical": CategoricalDescriptor
+            "Categorical": CategoricalDescriptor,
             "Organic": MolecularStructureDescriptor,
         }[data["type"]]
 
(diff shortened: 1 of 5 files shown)


andyczerwonka:
We'll have to reach out DS so that they can update notebooks.

bfolie:
> We'll have to reach out DS so that they can update notebooks.

To make the transition smoother I updated the PR to include a deprecated stub for `InorganicDescriptor`. This should mean that nobody's notebook will break, but they will get a warning message about deprecation.

@Imperssonator  is included here because he's the primary user. We'll figure out a way to make sure everybody else knows as well.

andyczerwonka:
> To make the transition smoother I updated the PR to include a deprecated stub for `InorganicDescriptor`. This should mean that nobody's notebook will break, but they will get a warning message about deprecation.

I saw that, which is awesome. That’s how I found this PR. 👍

src/citrine/informatics/descriptors.py
```diff
@@ -89,9 +91,15 @@ def __eq__(self, other):
...
 
-    def __init__(self, key: str, threshold: float = 1.0):
+    def __init__(self, key: str):
         self.key: str = key
-        self.threshold = threshold
+
+
+def InorganicDescriptor(key: str, threshold: Optional[float] = 1.0):
+    """[DEPRECATED] Use ChemicalFormulaDescriptor instead."""
+    logger.warning("InorganicDescriptor is deprecated and will soon be removed. "
```
maxhutch, line 100: I believe the standard pattern for deprecations is to use the warnings package, e.g.:
https://github.com/CitrineInformatics/gemd-python/blob/master/taurus/__init__.py#L4
which was pattern-copied from a deprecated python standard module.

bfolie: changed

PR #293, opened 2020-04-13
Add back the PR version bump check

maxhutch:
Use the "TRAVIS_PULL_REQUEST" environment variable to toggle the version check on only for pull requests that have the necessary git info.

https://docs.travis-ci.com/user/environment-variables/#convenience-variables

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,9 @@ script:
 - flake8 src
 - cd docs; make html; cd ..;
 - touch ./docs/_build/html/.nojekyll
-- if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then bash ./scripts/validate-version-bump.sh; fi
+# only run the validate-version-bump.sh script on PR builds against master, since those have access to a "master" reference to git show the version
+# following https://docs.travis-ci.com/user/environment-variables/#convenience-variables
+- if [ "$TRAVIS_PULL_REQUEST" == "true" ] && [ "$TRAVIS_BRANCH" == "master" ]; then bash ./scripts/validate-version-bump.sh; fi
 
 deploy:
   - provider: pages
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ script:
 - flake8 src
 - cd docs; make html; cd ..;
 - touch ./docs/_build/html/.nojekyll
+- ./scripts/validate-version-bump.sh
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -12,6 +12,8 @@ script:
 - flake8 src
 - cd docs; make html; cd ..;
 - touch ./docs/_build/html/.nojekyll
+- if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then bash ./scripts/validate-version-bump.sh; fi
```
dekatzenel, line 15: why not == "true"?

maxhutch: :shrug: I literally copied this from the travis docs:
https://docs.travis-ci.com/user/pull-requests/#pull-requests-and-security-restrictions

dekatzenel: That's stupid enough that it might be worth a comment

maxhutch: :fire:

.travis.yml
```diff
@@ -12,7 +12,10 @@ script:
 - flake8 src
 - cd docs; make html; cd ..;
 - touch ./docs/_build/html/.nojekyll
-- if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then bash ./scripts/validate-version-bump.sh; fi
+# only run the validate-version-bump.sh script on PR builds against master, since those have access to a "master" reference to git show the version
```
dekatzenel, line 15: I actually spot checked this and PRs against other branches appear to have access to a "master" reference as well. Agreed that it's only necessary for PRs against master, though

scripts/validate-version-bump.sh
```diff
@@ -1,7 +1,5 @@
 #!/bin/bash
 
-set -eux
```
dekatzenel: Why remove this? If you want to remove the logging, that's just the x. Also, if you're removing any part of this, please update the `trap` accordingly

maxhutch: I added back eu and updated the trap.  Removing the x was to quiet the logs a bit; I had assumed it was added there to sanity check that the script was working as expected.  I'm not a bash expert, though...

dekatzenel: I like -x because it creates a useful debugging error trail when things go wrong unexpectedly, but it does seem to be off by default at Citrine.

PR #296, opened 2020-04-14
Prepare objects for store before trying to validate them

dekatzenel:
# Citrine Python PR

## Description 
Prepare objects for store in the same way that register() does prior to running template validation on them. Prevents 500s on the backend which expects a specific format for incoming data

### PR Type:


dekatzenel:
> While it seems pretty reasonable to do this client-side, the fact that the datasvc 500's without it seems like an issue. Is there a separate PR / ticket to address that?

Just filed https://citrine.atlassian.net/browse/PLA-3840

PR #300, opened 2020-04-17
Taurus -> GEMD migration

kroenlein:
# Citrine Python PR
This PR is intended to
* Modify all requirements so that installs rely on gemd-python not citrine-taurus
* Update all in-library reference from Taurus to GEMD to avoid spurious deprecation warnings

## Description 
As described above.  This is motivated by impending roll-outs to a customer site.

I am less than confident in the corrections to Pipfile and Pipfile.lock, so extra care should be taken about verifying those edits.

### PR Type:


Pipfile
```diff
@@ -16,7 +16,7 @@ requests-mock = "*"
 
 [packages]
 citrine = {editable = true,path = "."}
-taurus = {git = "https://github.com/CitrineInformatics/taurus.git", editable = true, ref="develop"}
+gemd = {git = "https://github.com/CitrineInformatics/gemd-python.git", editable = true, ref="develop"}
```
bfolie, line 19: It looks like gemd doesn't have a develop branch. Did it used to have one? In that case I think it should be changed to `ref="master"`

Pipfile.lock
```diff
@@ -199,10 +199,10 @@
             ],
             "version": "==0.1.7"
         },
-        "taurus": {
+        "gemd": {
             "editable": true,
-            "git": "https://github.com/CitrineInformatics/taurus.git",
-            "ref": "e043283fab2f18112d8e71f102657527fc09efe9"
+            "git": "https://github.com/CitrineInformatics/gemd-python.git",
+            "ref": "1733aa751683b915666f30a24ca289cfcf9da305"
```
bfolie, line 205: Did you run `pipfile lock` or did you copy a git hash?

Pipfile
kroenlein: Yeah, it got moved into masterflow as well.  Good catch.

kroenlein:
```suggestion
gemd = {git = "https://github.com/CitrineInformatics/gemd-python.git", editable = true, ref="master"}
```

Pipfile.lock
kroenlein: Manually copied the git hash out of the 0.6.2 release, as tagged in PyPi.

PR #301, opened 2020-04-20
Docs/data sources

bfolie:
# Citrine Python PR

## Description 
Adds documentation on how to use data source objects. This is mostly copied from the ai-engine-documentation repo.

Also reorganizes the "AI Engine" section of the docs to have a top-level table of contents.

### PR Type:


docs/source/workflows/getting_started.rst
```diff
@@ -0,0 +1,80 @@
+.. _getting-started:
```
bfolie, line 1: This file is copied from what used to be in workflows/index.rst

docs/source/workflows/index.rst
```diff
@@ -1,83 +1,18 @@
...
-Validation status can be one of the following states:
-
--  **Created:** The module/workflow has been registered with a project and has been queued for validation.
--  **Validating:** The module/workflow is currently validating. The status will be updated to one of the subsequent states upon completion.
--  **Invalid:** Validation completed successfully but found errors with the workflow/module.
--  **Ready:** Validation completed successfully and found no errors.
--  **Error:** Validation did not complete. An error was raised during the validation process that prevented an invalid or ready status to be determined.
-
-Validation of a workflow and all constituent modules must complete with ready status before the workflow can be executed.
+.. toctree::
```
bfolie, line 7: workflows/index.rst is now a table of contents

docs/source/workflows/data_sources.rst
```diff
@@ -0,0 +1,90 @@
...
+    from citrine.informatics.data_sources import CSVDataSource
+    from citrine.informatics.predictors import SimpleMLPredictor
+    from citrine.informatics.descriptors import RealDescriptor, CategoricalDescriptor, ChemicalFormulaDescriptor
+
+    file_link = dataset.files.upload("./data.csv", "bandgap_data.csv")
+
+    data_source = CsvDataSource(
+        file_link = file_link,
+        column_definitions = {
+            "Chemical Formula": ChemicalFormulaDescriptor("formula"),
```
yoolheekim, line 41: (minor request) it could be useful to add that the key here references the column header, and explicitly state that the descriptor key does not need to match the column header key

docs/source/workflows/data_sources.rst
```diff
@@ -0,0 +1,90 @@
...
+            "Crystallinity": CategoricalDescriptor("Crystallinity", categories=[
+                "Single crystalline", "Amorphous", "Polycrystalline"])
+        }
+    )
+
+    predictor = SimpleMLPredictor(
+        name = "Band gap predictor",
+        description = "Predict the band gap from the chemical formula and crystallinity",
+        inputs = [
+            data_source.column_definitions["Chemical Formula"],
```
yoolheekim, line 52: (minor request) would be useful to explicitly state that the column_definitions are identified by the column headers, and not by the descriptor keys

bfolie: updated

docs/source/workflows/data_sources.rst
bfolie: updated

docs/source/workflows/data_sources.rst
yoolheekim: (question) is it not required to use `data_source.column_definitions` here? can you reference instead a descriptor object like `ChemicalFormulaDescriptor("formula")` ? 

also, what I meant is that it would be nice to state that the key in the `column_definitions` should refer to the csv column header and not the descriptor key (i.e. "formula")

docs/source/workflows/data_sources.rst
```diff
@@ -0,0 +1,94 @@
...
+
+The :class:`~citrine.informatics.data_sources.CSVDataSource` draws data from a CSV file stored on the data platform and annotates it by mapping each column name to a :class:`~citrine.informatics.descriptors.Descriptor`.
+The file is referenced via a :class:`~citrine.resources.file_link.FileLink`.
+Each FileLink references an explicit version of the CSV file.
+Uploading a new file with the same name will produce a new version of the file with a new FileLink.
+
+The columns in the CSV are extracted and parsed by a mapping of column header names to user-created descriptors.
+Columns in the CSV that are not mapped with a descriptor are ignored.
+
+Assume that a file data.csv exists with the following contents:
```
bfolie, line 18: @floparedes @maxhutch is there a limit on the number of rows/columns/cells in a CSV data source? I see limits on Ara tables here and total number of training data for a predictor, but am not sure which parts transfer to CSV data source: https://citrine.atlassian.net/wiki/spaces/PLA/pages/251297805/AI+Engine+limitations+and+questions+for+CP+Onboarding

maxhutch: There is not currently a limit.  I think we should cap the file size because we read the whole file into memory in orion.  cc: @andyczerwonka @jtharris

docs/source/workflows/data_sources.rst
bfolie: >  is it not required to use data_source.column_definitions here? can you reference instead a descriptor object like ChemicalFormulaDescriptor("formula") ?

It is not. You can create the descriptor directly but it needs to match the descriptor in the CSV data source, hence the comment on lines 54-55.

> what I meant is that it would be nice to state that the key in the column_definitions should refer to the csv column header and not the descriptor key (i.e. "formula")

I think that's covered in lines 40-41. "`column_definitions` maps a column header to a descriptor" states that the key in the dictionary refers to a column header. And "the column header and the descriptor key do not need to be identical" states that it doesn't refer to the descriptor key.

If you have different wording in mind, feel free to make suggested changes.

yoolheekim: OK, that makes sense to me. I'm good with this! Thanks

PR #302, opened 2020-04-20
updating original object uids in register_all

ventura-rivera:
# Citrine Python PR

## Description 

### PR Type:

### DIFF
--- a/src/citrine/resources/dataset.py
+++ b/src/citrine/resources/dataset.py
@@ -30,6 +30,7 @@ from taurus.entity.template.measurement_template import MeasurementTemplate
 from taurus.entity.template.parameter_template import ParameterTemplate
 from taurus.entity.template.process_template import ProcessTemplate
 from taurus.entity.template.property_template import PropertyTemplate
+from taurus.entity.base_entity import BaseEntity
 from taurus.util import writable_sort_order
 
 ResourceType = TypeVar('ResourceType', bound='DataConcepts')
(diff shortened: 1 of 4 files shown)


ventura-rivera:
> Is there a way to add even a basic citrine-python test?

I'm looking through the factory objects to see how I can do this

src/citrine/resources/dataset.py
```diff
@@ -231,9 +231,14 @@ def register_all(self, data_concepts_resources: List[ResourceType]) -> List[Reso
 
         :return the registered versions
         """
-        return [self.register(resource) for resource
```
maxhutch, line 234: Is this change necessary?  I would have thought that `self.register(resource)` would return the resource with the uid set.

tests/utils/session.py
```diff
@@ -86,12 +86,12 @@ def checked_delete(self, path: str, **kwargs) -> dict:
         self.calls.append(FakeCall('DELETE', path, params=kwargs.get('params')))
         return self._get_response()
 
-    def _get_response(self):
+    def _get_response(self, default_response: dict = {}):
```
dekatzenel, line 89: python has a thing about default mutable arguments: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

src/citrine/resources/dataset.py
dekatzenel: More specifically, the issue occurs when trying to store an object that has a reference to an object that doesn't have a uid

src/citrine/resources/dataset.py
```diff
@@ -231,9 +232,14 @@ def register_all(self, data_concepts_resources: List[ResourceType]) -> List[Reso
...
-                in (sorted(data_concepts_resources,
-                           key=lambda resource: writable_sort_order(resource.typ)))]
+        resources = list()
+        for resource in (sorted(data_concepts_resources,
+                         key=lambda resource: writable_sort_order(resource.typ))):
+            registered_resource = self.register(resource)
+            if isinstance(registered_resource, BaseEntity):
+                resource.uids = registered_resource.uids
```
dekatzenel, line 240: Updating uids in place is unexpected behavior that is worth documenting on the method (including a quick explanation of what this supports)

tests/utils/session.py
ventura-rivera: Fixed this (used the pattern in the link)

src/citrine/resources/dataset.py
ventura-rivera: Added explanation on behavior to the docstrings

ventura-rivera:
@dekatzenel Yes, I can do that

PR #304, opened 2020-04-22
Mock versioning automation proposal

asantas93:
# Citrine Python PR

## Description 
Approached based upon the interface of https://github.com/semantic-release/semantic-release, which is designed for JS projects.

How it works:
* All version state is stored in git tags
* setup.py reads version from git at runtime
* Commit messages are prefixed with bump 'sizes' which are used to create a new git tag in Travis
* PyPi releases continue to be triggered from git tags

What this means
* We don't commit any information about version, so there aren't any merge conflicts around editing it
* Post-merge travis job cannot run concurrently safely — I think it’s possible our fast-forward-only setting makes this happen, but if not travis also lets you configure max parallelism, and 1 doesn’t seem too bad given how fast the pipeline is
* We no longer have to manually release to pypi, but we still have the ability to do it manually with a manual git tag and push
* The version bump type can be edited during PR review if needed
* We may accumulate a lot of bump type messages in our commit log when a PR has a lot of iterations and we don't use rebase.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ python:
 - '3.6'
 install:
 - git tag
+- echo $TRAVIS_COMMIT_MESSAGE
 - pip install -U -r requirements.txt
 - pip install -U -r test_requirements.txt
 - pip install --no-deps -e .
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,6 @@ python:
 - '3.5'
 - '3.6'
 install:
-- git tag
-- echo $TRAVIS_COMMIT_MESSAGE
 - pip install -U -r requirements.txt
 - pip install -U -r test_requirements.txt
 - pip install --no-deps -e .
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -16,7 +16,26 @@ script:
...
+      elif msg.startswith('[major]')
+          bump = 1.0.0
+      elif msg.startswith('[pre]')
+          bump = None
+      else
+          exit 1 "Invalid semver bump"
+      if bump
+          tag = prev_tag + bump
+          git tag $tag
+          git push --tags
```
maxhutch, line 38: Note that this is effectively going back to continuous deployment of every merge to master.  Currently, our release process involves manually triggering verification jobs before cutting releases; If we make this change, we probably need to gate PRs to master on those tests against the PR code, which could significantly add to jenkins load.

setup.py
```diff
@@ -56,8 +56,14 @@ def run(self):
...
 
+def get_git_tag():
+    # probably not valid subprocess usage, but you get the idea
+    import subprocess
+    return subprocess.run('git tag -l | HEAD -1').stdout
+
+
 setup(name='citrine',
-      version='0.16.4',
+      version=get_git_tag(),
```
maxhutch, line 66: Does this work with pypi-hosted releases?  Namely, if this is released to and then installed from pypi, do I get an error?  Does `pip show` work?

maxhutch: While we don't currently, I might expect that we want to expose the client version as something like:
```python
client = Citrine(...)
print(client.version)
```
Does this pattern play nice with that?

PR #306, opened 2020-04-22
Feature/pla 3508 expose gemd relation searches

asantas93:
# Citrine Python PR

## Description 
Adds numerous searches for GEMD objects based upon their relationships to one another. These searches are still using the collection-style syntax to avoid a major design change such as giving every `DataConcepts` object a `Session` object, e.g. to get a process's output material, you must write

```python
my_project.material_runs.get_by_process(my_process.uids['id'])
```
rather than
```python
my_process.output_material
```
If the latter syntax is desirable, we can add it later and reuse the collection methods added in this PR.

### PR Type:


src/citrine/resources/material_spec.py
```diff
@@ -105,12 +112,43 @@ def filter_by_template(self,
...
-        raw_objects = self.session.cursor_paged_resource(self.session.get_resource,
-                                                         path,
-                                                         per_page=per_page,
-                                                         version="v1")
-        return (self.build(raw) for raw in raw_objects)
+        if per_page is not None:
+            logger.warning('The per_page parameter will be ignored. Please remove it.')
+        return self.list_by_template(uid=template_id, scope=template_scope)
+
+    def list_by_template(self, uid: Union[UUID, str], scope: str = 'id') -> Iterator[MaterialSpec]:
```
asantas93, line 119: These methods are pretty copypasta-y. Unfortunately, our data model isn't quite consistent enough to be able to add a valid `list_by_template` to `ObjectSpecCollection`.

src/citrine/resources/ingredient_run.py
```diff
@@ -119,3 +120,58 @@ class IngredientRunCollection(ObjectRunCollection[IngredientRun]):
...
+            The unique ID of the ingredient spec whose ingredient run usages are to be located.
+        scope
+            The scope of `uid`.
+        Returns
+        -------
+        Iterator[IngredientRun]
+            The ingredient runs using the specified ingredient spec.
+
+        """
+        return self._get_relation('ingredient-specs', uid=uid, scope=scope)
```
dekatzenel, line 140: Maybe use IngredientSpecCollection._collection_key (and similar elsewhere)?

src/citrine/resources/data_concepts.py
```diff
@@ -571,3 +571,45 @@ def delete(self, uid: Union[UUID, str], scope: str = 'id', dry_run: bool = False
...
+        Returns
+        -------
+        Iterator[ResourceType]
+            Objects in this collection which are somehow related to the object with
+            provided uid and scope.
+
+        """
+        params = {}
+        if self.dataset_id is not None:
+            params['dataset_id'] = str(self.dataset_id)
```
dekatzenel, line 606: do we know to use this on the backend?

src/citrine/resources/material_run.py
```diff
@@ -143,10 +150,30 @@ def get_history(self, scope, id) -> Type[MaterialRun]:
...
+        Returns
+        -------
+        MaterialRun
+            The output material of the specified process, or None if no such material exists.
+
+        """
+        return next(
+            self._get_relation(relation='process-runs', uid=uid, scope=scope, per_page=1), None)
+
+    @deprecation.deprecated(details='Use list_by_spec instead.')
```
dekatzenel, line 172: Don't forget to update the citrine-python docs that refer to using this.

src/citrine/resources/material_spec.py
dekatzenel: We could have yet another layer of abstract indirection after ObjectSpec/ObjectSpecCollection IngredientSpec/IngredientSpecCollection doesn't inherit from

src/citrine/resources/measurement_run.py
```diff
@@ -105,3 +106,40 @@ class MeasurementRunCollection(ObjectRunCollection[MeasurementRun]):
     def get_type(cls) -> Type[MeasurementRun]:
         """Return the resource type in the collection."""
         return MeasurementRun
+
+    def list_by_spec(self, uid: Union[UUID, str], scope: str = 'id') -> Iterator[MeasurementRun]:
```
dekatzenel, line 110: also implement list_by_template?

src/citrine/resources/process_run.py
```diff
@@ -103,3 +104,21 @@ class ProcessRunCollection(ObjectRunCollection[ProcessRun]):
     def get_type(cls) -> Type[ProcessRun]:
         """Return the resource type in the collection."""
         return ProcessRun
+
+    def list_by_spec(self, uid: Union[UUID, str], scope: str = 'id') -> Iterator[ProcessRun]:
```
dekatzenel, line 108: also implement list_by_template?

src/citrine/resources/material_spec.py
```diff
@@ -105,12 +112,43 @@ def filter_by_template(self,
...
+            The scope of `uid`.
+        Returns
+        -------
+        Iterator[MaterialSpec]
+            The material specs using the specified material template.
+
+        """
+        return self._get_relation('material-templates', uid=uid, scope=scope)
+
+    def get_by_process(self, uid: Union[UUID, str], scope: str = 'id') -> Optional[MaterialSpec]:
```
dekatzenel, line 137: should this be called list_by_process?

src/citrine/resources/ingredient_run.py
asantas93: The collection key is slightly off because it uses an underscore. Nothing a `replace` couldn't fix but I'm not sure how much string manipulation we want to do for this. Thoughst?

src/citrine/resources/data_concepts.py
asantas93: Yep https://github.com/CitrineInformatics/platform-backend/blob/master/code/faraday/dc-data-objects/src/main/scala/io/citrine/dataobjects/api/StorableRoute.scala#L263

src/citrine/resources/measurement_run.py
asantas93: I was holding off on implementing it since it wasn't asked for explicitly. To me, adding `list_by_template` implies its implementation is efficient, but it's not. If the user would use the intermediate values (specs), then it would be better for them to write that code themselves so they can take advantage of those values.

src/citrine/resources/ingredient_run.py
dekatzenel: Is it worth trying to make use of the dataset agnostic path template, then?

src/citrine/resources/material_spec.py
asantas93: Validations ensure that there is at most 1 material associated with a process.

src/citrine/resources/ingredient_run.py
asantas93: It would be awkward in this case because the dataset agnostic path you want would be from the relation's collection, yet the build method you want to use is in this collection. I don't see much utility to the dataset agnostic path (and other shared string templates) anyway...it eliminates a relatively low amount of copypasta and adds a lot more opacity to what the final rendered endpoint will actually look like.

src/citrine/resources/material_spec.py
asantas93: I would be inclined to do that if this method had any complexity to it, but as it is, it's a simple one-liner. An additional class which doesn't reflect a group of object types in our GEMD docs doesn't seem worth it to me to reduce 2 one-line methods to 1 one-line method.

docs/source/data_entry.rst
```diff
@@ -68,9 +68,9 @@ and :func:`~citrine.resources.data_concepts.DataConceptsCollection.filter_by_nam
 can help refine the listing to make the target object easier to find.
 
 For material runs and specs, there exist additional types of filtering:
-:func:`~citrine.resources.material_run.MaterialRunCollection.filter_by_spec`,
-:func:`~citrine.resources.material_run.MaterialRunCollection.filter_by_template`,
-and :func:`~citrine.resources.material_spec.MaterialSpecCollection.filter_by_template`.
+:func:`~citrine.resources.material_run.MaterialRunCollection.list_by_spec`,
```
dekatzenel, line 71: Also worth mentioning new implementations? Or maybe not if it's alpha

asantas93: I was going to but I'm not sure if it's worth the time if we're going to push for different search syntax. @djack do you want me to add this before DE takes a look?

ventura-rivera:
Thanks for the heads-up, the implementation of `get_by_process` and deprecation marks of `filter_by`'s look good to me!

PR #310, opened 2020-04-23
Add test to codify expected behavior when uids is an empty dict

dekatzenel:
# Citrine Python PR

## Description 
Link replacement is surprisingly forgiving. Codify existing behavior when there is a empty uids dict so that we can decide if it's correct or not

### PR Type:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ language: python
 python:
 - '3.5'
 - '3.6'
+- '3.7'
+- '3.8'
 install:
 - pip install -U -r requirements.txt
 - pip install -U -r test_requirements.txt
--- a/docs/source/workflows/data_sources.rst
+++ b/docs/source/workflows/data_sources.rst
@@ -15,15 +15,21 @@ Uploading a new file with the same name will produce a new version of the file w
 The columns in the CSV are extracted and parsed by a mapping of column header names to user-created descriptors.
 Columns in the CSV that are not mapped with a descriptor are ignored.
 
+An optional list of identifiers can be specified.
+Each identifier is a column header name.
+These may overlap with the keys defined in the mapping from column header names to descriptors if a column should be parsed as data and used as an identifier.
+Identifiers should be globally unique.
+No two rows should contain the same value.
+
 Assume that a file data.csv exists with the following contents:
 
 .. code::
 
-    Chemical Formula,Gap,Crystallinity
-    Bi2Te3,0.153,Single crystalline
-    Mg2Ge,0.567,Single crystalline
-    GeTe,0.7,Amorphous
-    Sb2Se3,1.15,Polycrystalline
+    Chemical Formula,Gap,Crystallinity,Sample ID
+    Bi2Te3,0.153,Single crystalline,0
+    Mg2Ge,0.567,Single crystalline,1
+    GeTe,0.7,Amorphous,2
+    Sb2Se3,1.15,Polycrystalline,3
 
 That file could be used as the training data for a predictor as:
 
@@ -44,7 +50,8 @@ That file could be used as the training data for a predictor as:
             "Gap": RealDescriptor("Band gap", lower_bound=0, upper_bound=20, units="eV"),
             "Crystallinity": CategoricalDescriptor("Crystallinity", categories=[
                 "Single crystalline", "Amorphous", "Polycrystalline"])
-        }
+        },
+        identifiers = ["Sample ID"]
     )
 
     predictor = SimpleMLPredictor(
--- a/docs/source/workflows/design_spaces.rst
+++ b/docs/source/workflows/design_spaces.rst
@@ -1,7 +1,7 @@
 Design spaces
 =============
 
-A design space defines a finite or infinite set of materials that can be made.
+A design space defines a set of materials that should be searched over when performing a material design.
 Currently, there are two design spaces:
 
 -  `ProductDesignSpace <#product-design-space>`__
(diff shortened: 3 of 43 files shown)


asantas93:
The behavior I would expect is that you get a client-side exception as soon as you attempt to reference a GEMD object which does not have uids, e.g.

```python
spec = ProcessSpec('foo')
run = ProcessRun('bar', spec=spec)  # throws exception
```

I would expect the same behavior with the gemd version of the models as well, although I don't feel as strongly in that context.

dekatzenel:
> The behavior I would expect is that you get a client-side exception as soon as you attempt to reference a GEMD object which does not have uids, e.g.
> 
> ```python
> spec = ProcessSpec('foo')
> run = ProcessRun('bar', spec=spec)  # throws exception
> ```
> 
> I would expect the same behavior with the gemd version of the models as well, although I don't feel as strongly in that context.

Note: citrine-python adds a CitrineId by default when instantiating an object, so the example above is using GEMD objects, at least for the spec.

That change feels like just enough work to be worth its own ticket, since we'd have to either implement the validation for every link manually or maybe do something clever in LinkOrElse

asantas93:
> Note: citrine-python adds a CitrineId by default when instantiating an object, so the example above is using GEMD objects, at least for the spec.

IMO we should really NOT do this. The user should be forced to specify an external ID to link with manually if they want to use a data object which hasn't yet been written to the backend.

maxhutch:
I'm not sure that I quite grok the test that this PR is adding.

Independently, if we're talking about changing the behavior of client-side links between objects, we should loop in @kroenlein and @EricLundberg

dekatzenel:
> I'm not sure that I quite grok the test that this PR is adding.

It is testing what happens when both type and uids exist, and uids is an empty dict. The answer turns out to be that translation to uid doesn't happen.

maxhutch:
> > I'm not sure that I quite grok the test that this PR is adding.
> 
> It is testing what happens when both type and uids exist, and uids is an empty dict. The answer turns out to be that translation to uid doesn't happen.

Gotcha.  Do we know why `replace_objects_with_links` is being used instead of [substitute_links](https://github.com/CitrineInformatics/gemd-python/blob/master/gemd/util/impl.py#L73) cc @bfolie ?

bfolie:
> Gotcha. Do we know why `replace_objects_with_links` is being used instead of [substitute_links](https://github.com/CitrineInformatics/gemd-python/blob/master/gemd/util/impl.py#L73) cc @bfolie ?

I don't know. My recollection is that `replace_objects_with_links` was created to deal with some specific parts of serialization in the client, but isn't meant to be used by itself.

dekatzenel:
That change appears to break e2e tests. Example:
```
___________________________________________________________________________ test_empirical_formula ___________________________________________________________________________

template_dataset = <citrine.resources.dataset.Dataset object at 0x7faddf546908>, dataset = <citrine.resources.dataset.Dataset object at 0x7fadb9e68ac8>

    def test_empirical_formula(template_dataset: Dataset, dataset: Dataset):
        proc = ProcessSpec(name='process')
    
        good_tmpl = PropertyTemplate(
            name='good_tmpl',
            bounds=CompositionBounds(components=['Na', 'Cl']),
        )
    
        good_prop_good_tmpl = PropertyAndConditions(
            Property(
                name='good_prop',
                value=EmpiricalFormula(formula='NaCl'),
                template=good_tmpl
            )
        )
    
        bad_tmpl = PropertyTemplate(
            name='bad_tmpl',
            bounds=CompositionBounds(components=['bad', 'tmpl']),
        )
    
        good_prop_bad_tmpl = PropertyAndConditions(
            Property(
                name=bad_tmpl.name,
                value=EmpiricalFormula(formula='NaCl'),
                template=bad_tmpl
            )
        )
    
        bad_prop_good_tmpl = PropertyAndConditions(
            Property(
                name='bad_formula',
                value=EmpiricalFormula(formula='garbage'),
                template=good_tmpl
            )
        )
    
        dataset.process_specs.register(proc)
        template_dataset.property_templates.register(good_tmpl)
        template_dataset.property_templates.register(bad_tmpl)
    
        with pytest.raises(BadRequest) as expt:
            dataset.material_specs.register(MaterialSpec(
                name='fail_bad_formula',
                properties=[bad_prop_good_tmpl],
                process=proc
            ))
>       expt_failure_id("empirical_formula.invalid", expt)

tests/strike-team-data/test_template_validation.py:303: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

id = 'empirical_formula.invalid', expt_info = <ExceptionInfo BadRequest tblen=5>

    def expt_failure_id(id, expt_info):
        bad_request: BadRequest = expt_info.value
>       return any(id in f.failure_id for f in bad_request.api_error.validation_errors)
E       AttributeError: 'NoneType' object has no attribute 'validation_errors'

tests/utils/__init__.py:17: AttributeError
----------------------------------------------------------------------------- Captured log call ------------------------------------------------------------------------------
_session.py                118 ERROR    400 POST projects/84fc440a-a152-4cdb-8301-c86cb7797634/datasets/b25423d3-2702-455c-9088-32c5ed80655e/material-specs
_session.py                119 ERROR    The request content was malformed:
Parsed JSON values do not match with class constructor
args=fail_bad_formula,ProcessSpec(process,None,Set(),Set(),None,Uids(Set(CitrineId(183d753d-9eeb-415f-a3ec-a76c4bfaa726))),Set(),None,Set(),None,None),Set(PropertyAndConditions(Property(bad_formula,EmpiricalFormula(garbage),Some(PropertyTemplate(good_tmpl,CompositionBounds(Set(Cl, Na)),Uids(Set(CitrineId(5b9d394a-a23b-4735-9713-c83110ab562c))),None,Set(),None,None)),unknown,Set(),None),Set())),None,Uids(Set(CitrineId(a0bf63ce-1a56-407c-8953-3e7d474d55a6))),Set(),None,Set(),None,None
arg types=java.lang.String,io.citrine.faraday.api.model.objects.ProcessSpec,scala.collection.immutable.Set$Set1,scala.None$,io.citrine.faraday.api.model.Uids,scala.collection.immutable.Set$EmptySet$,scala.None$,scala.collection.immutable.Set$EmptySet$,scala.None$,scala.None$
executable=Executable(Constructor(public io.citrine.faraday.api.model.objects.MaterialSpec(java.lang.String,io.citrine.faraday.api.model.LinkOrElse,scala.collection.immutable.Set,scala.Option,io.citrine.faraday.api.model.Uids,scala.collection.immutable.Set,scala.Option,scala.collection.immutable.Set,scala.Option,scala.Option)))
cause=java.lang.ClassCastException@1c4e99c2
types comparison result=MATCH,io.citrine.faraday.api.model.objects.ProcessSpec(sun.misc.Launcher$AppClassLoader@18b4aac2) !<: io.citrine.faraday.api.model.LinkOrElse(sun.misc.Launcher$AppClassLoader@18b4aac2),MATCH,MATCH,MATCH,MATCH,MATCH,MATCH,MATCH,MATCH

```

maxhutch:
The gemd-python `substitute_links` approach doesn't work because citrine-python overrides `as_dict` in a way that breaks it.  I'm sure we'll need to deal with that at some point, but it is not this day.  Tests pass both in repo and when this version is used with the e2e tests against development.

PR #314, opened 2020-04-24
Testing against and documenting python >=3.5 support

maxhutch:
This adds tests for 3.7 and 3.8, and adds classifiers to setup.py


maxhutch:
Relates to, but does not actually depend on, https://github.com/CitrineInformatics/gemd-python/pull/98.  Since no code needed to change to pass 3.7 or 3.8 tests, no redeploy of either repo is actually needed.

PR #318, opened 2020-04-24
[PLA-3993] Added dry_run to dataset registers, fixed some docstrings

sesevgen:
# Citrine Python PR

## Description 
Very small addition to enable dry run registers through Dataset.

### PR Type:


ventura-rivera:
@sesevgen could you add a test where the dry run validation is expected to fail and succeed?

sesevgen:
@ventura-rivera I haven't actually looked into the implementation of dry_run. This change is just propagating that call from dataset to the data collections for our ease of use. I'd have to look into the relevant code for that test, but I had assumed it already existed when the original dry_run functionality was implemented. If not, I can create a separate PR if there's a need.

PR #320, opened 2020-04-29
Add caveat about CSV files

bfolie:
# Citrine Python PR

## Description 
Add caveat about using CSV files saved with Excel.

### PR Type:

### DIFF
--- a/docs/source/workflows/data_sources.rst
+++ b/docs/source/workflows/data_sources.rst
@@ -15,6 +15,10 @@ Uploading a new file with the same name will produce a new version of the file w
 The columns in the CSV are extracted and parsed by a mapping of column header names to user-created descriptors.
 Columns in the CSV that are not mapped with a descriptor are ignored.
 
+The CSV must be UTF-8 encoded ASCII text with CRLF line terminators.
+The best way to ensure this is to save the file in a "Comma Separated Values" format and not any of the other small CSV variations.
+For example, Excel's default CSV format, "CSV UTF-8 (Comma Delimited)" is not parseable.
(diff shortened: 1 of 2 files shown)


docs/source/workflows/data_sources.rst
```diff
@@ -15,6 +15,10 @@ Uploading a new file with the same name will produce a new version of the file w
 The columns in the CSV are extracted and parsed by a mapping of column header names to user-created descriptors.
 Columns in the CSV that are not mapped with a descriptor are ignored.
 
+The CSV must be UTF-8 encoded ASCII text with CRLF line terminators.
+The best way to ensure this is to save the file in a "Comma Separated Values" format and not any of the other small CSV variations.
```
bfolie, line 19: I compared the working and failed files some more and the difference doesn't actually seem to be the encoding. They're both UTF-8 encoded according to Python
```
with open("file.csv") as f:
    print(f)
```
But when I use the terminal command `file file.csv`, the failed files just say "ASCII text" while the working files say "ASCII text, with CRLF line terminators"

@andyczerwonka @talltea do you have a concise explanation for what this difference?

andyczerwonka:
@bfolie @talltea `U+FEFF` is is a Byte-Order Marker (BOM), probably something that is inserted using the Excel Mac version. We should be able to handle this on the server side - I will argue that this is a bug. I will write a failing test and log an issue. I think we can close the PR once we're written this up.

bfolie:
Logged here: https://citrine.atlassian.net/browse/PLA-4013

PR #322, opened 2020-04-30
Fix incorrect tag param in list_by_tag

asantas93:
# Citrine Python PR

## Description 
`list_by_tag` actually just lists by dataset because `tag` isn't actually the correct search parameter. 

### PR Type:

### DIFF
--- a/src/citrine/resources/data_concepts.py
+++ b/src/citrine/resources/data_concepts.py
@@ -542,7 +542,7 @@ class DataConceptsCollection(Collection[ResourceType], ABC):
             Every object in this collection.
 
         """
-        params = {'tag': tag}
+        params = {'tags': [tag]}
         if self.dataset_id is not None:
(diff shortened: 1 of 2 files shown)


src/citrine/resources/data_concepts.py
```diff
@@ -542,7 +542,7 @@ def list_by_tag(self, tag: str, per_page: int = 100) -> Iterator[ResourceType]:
             Every object in this collection.
 
         """
-        params = {'tag': tag}
+        params = {'tags': [tag]}
```
maxhutch, line 545: https://github.com/CitrineInformatics/platform-backend/blob/master/code/faraday/dc-data-objects/src/main/scala/io/citrine/dataobjects/api/StorableRoute.scala#L198 checks out

bfolie:
Is there a Mock test somewhere that needs to be updated?

PR #323, opened 2020-05-01
PLA-3994: list pagination 

talltea:
# Citrine Python PR

## Description 
*This is still WIP*
This PR attempts to address these two main issues:
1. the list endpoints don't make use of the module type filtering functionality of the back end and rely on serialization errors to filter modules client side
2. the pagination of the list endpoint stops prematurely when the  number modules on the client side is less than the number requested by page after the client does filtering and deserialization

Solution summaries:
1. add a `_module_type` field that can be used in the list endpoint (so far only to predictors, but I'm also going to add it to other supported modules)
2. refactor the pagination code to allow for use of the `next` field provided in the backend response field to determine if we are at the end of the pagination

Things I'm still working on:
1. Editing docstrings and type annotations
2. Ensuring full test coverage

### PR Type:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,7 @@ class PreBuildCommand(build_py):
 
 
 setup(name='citrine',
-      version='0.17.7',
+      version='0.17.8',
       url='http://github.com/CitrineInformatics/citrine-python',
       description='Python library for the Citrine Platform',
       author='Citrine Informatics',
--- a/src/citrine/_rest/collection.py
+++ b/src/citrine/_rest/collection.py
@@ -144,11 +144,10 @@ class Collection(Generic[ResourceType]):
         for element in collection:
             try:
                 yield self.build(element)
-            except(KeyError, ValueError) as e:
+            except(KeyError, ValueError):
                 # TODO:  Right now this is a hack.  Clean this up soon.
                 # Module collections are not filtering on module type
                 # properly, so we are filtering client-side.
-                print(e)
                 pass
 
     def _page_params(self,
--- a/src/citrine/_rest/collection.py
+++ b/src/citrine/_rest/collection.py
@@ -119,7 +119,8 @@ class Collection(Generic[ResourceType]):
         """
         path = self._get_path()
 
-        params = self._page_params(page, per_page)
+        module_type = getattr(self, '_module_type', None)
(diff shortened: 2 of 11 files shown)


bfolie:
Assuming the paginator works as expected in the backend (if `page` is too large then it returns an empty list and no value of `next_uri`), then this approach looks like it will work.

talltea:
> Assuming the paginator works as expected in the backend (if `page` is too large then it returns an empty list and no value of `next_uri`), then this approach looks like it will work.

I've confirmed this branch works with `project.predictors.list()` against development as expected

andyczerwonka:
>     1. Are we actually using the `next` url to fetch the next page?  There are lots of benefits to doing this, a big one being that we don't have to thread any knowledge of paging throughout the client.

I would like to capture this, as I think this is the whole point of the API and something we should be doing. Saying that, I'd like to get over the dev0 and log a ticket for this.

talltea:
> 1. Are we actually using the `next` url to fetch the next page?  There are lots of benefits to doing this, a big one being that we don't have to thread any knowledge of paging throughout the client.

We're not currently using it. I agree there's a lot of benefit of using it, but I didn't want to pull that change in to this PR because that would have been a larger refactor.

> 2. Are there any other missing module types?  The specific bug that we are addressing in development is with design spaces, not predictors, right?

Absolutely. I missed that in my PR description as a todo, but I will be going through and adding the other module types supported. The ones I see are `DESIGN_SPACE, PROCESSOR, PREDICTOR, SCORER, DESIGN_WORKFLOW, CONSTRAINT, OBJECTIVE, SCORE, PERFORMANCE_WORKFLOW`

talltea:
> I think we should add the `_module_type` field to `Workflow`. Otherwise LGTM

I ran into a problem with this since the backend distinguishes between performance and design workflows on the list endpoint, but the client doesn't. So it would require a slightly different fix client side. Since that change isn't necessary to get us past the dev0, I was going to handle it in a separate pr.

andyczerwonka:
>  Since that change isn't necessary to get us past the dev0, I was going to handle it in a separate pr.

Let's write up a separate issue for that, but yes, I agree. @bfolie are we okay with that?

PR #326, opened 2020-05-05
PLA-3998: predictor report bugfix 2

bfolie:
# Citrine Python PR

## Description 
Improve on the rushed #325 

This makes the logic clearer: we traverse the JSON dictionary until we get to the terminal key. If a key is ever missing traversal stops and we use the default value (if applicable). If the JSON has unexpected structure, that's an error.

### PR Type:

### DIFF
--- a/src/citrine/_serialization/properties.py
+++ b/src/citrine/_serialization/properties.py
@@ -83,6 +83,8 @@ class Property(typing.Generic[DeserializedType, SerializedType]):
             if next_value is None:
                 value = self.default
                 break
+            else:
+                value = next_value
 
         base_class = _get_base_class(data, self.serialization_path)
         return self.deserialize(value, base_class=base_class)
--- a/src/citrine/_serialization/properties.py
+++ b/src/citrine/_serialization/properties.py
@@ -83,7 +83,9 @@ class Property(typing.Generic[DeserializedType, SerializedType]):
             next_value = value.get(field)
             if next_value is None:
                 # This occurs if a `field` is unexpectedly not present in the data dictionary
-                # Stop traversing and use the default value
+                # or if its value is null.
+                # Use the default value and stop traversing, even if we have not yet reached
+                # the last field in the serialization path.
                 value = self.default
                 break
             else:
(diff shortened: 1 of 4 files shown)


src/citrine/_serialization/properties.py
```diff
@@ -79,10 +79,13 @@ def deserialize_from_dict(self, data: dict) -> DeserializedType:
         value = data
         fields = self.serialization_path.split('.')
         for field in fields:
-            if isinstance(value, dict):
-                value = value.get(field, self.default)
-            else:
+            next_value = value.get(field)
+            if next_value is None:
```
asantas93, line 83: The break you added here looks right. However, this isn't what I was suggesting in the other PR. I was saying you should check if `value is None`. In that case it will be unsafe to do a `get` on it, so you should also use a default value and break in that case. I don't think this code will handle the key where the "root" key doesn't exist.

asantas93: (your `next_value is None` does seem appropriate though, not saying to remove it)

bfolie: Since `value` is set from `next_value`, the only way you'd have `value=None` and try to perform a `.get` on it is if the thing that you expect to be a JSON dictionary is actually a `None`. If we're expecting to deserialize a JSON dictionary and instead get `None` or some other not-dict, then I think that's a bad.

asantas93: I think you're right that that's the only case in which this code fails. If we assume that the input JSON is never `None` (I don't know if this is a valid assumption or not), then this code does the same thing as what I suggested. However, to me the way that it's written doesn't communicate that part of the purpose of the `is None` check is to make sure the next iteration doesn't do a `get` on `None`. If I were the one writing this function, I would either rewrite it to communicate that intent by also doing the check on `value`, or, if I resented the verbosity of writing it that way, I'd just add an inline comment explaining the dual-purpose of the check.

bfolie: I added some comments. I don't know if they'll be entirely to your satisfaction, though, since I'm thinking about the logic a bit differently. I think the point of the `is None` check is because it doesn't make sense to keep traversing the dictionary if a key isn't present. An attempted `get` on a `None` (or the `self.default` value) is in my mind a secondary effect, it's just how the logical mistake manifests as an exception.

PR #329, opened 2020-05-05
Pla 3998/handle duplicate descriptors

bfolie:
# Citrine Python PR

## Description 
Gracefully handle multiple descriptors in a predictor report that have the same key.
PLA-4056

### PR Type:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,7 @@ class PreBuildCommand(build_py):
 
 
 setup(name='citrine',
-      version='0.18.4',
+      version='0.18.5',
       url='http://github.com/CitrineInformatics/citrine-python',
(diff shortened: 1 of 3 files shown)


bfolie:
How should we handle a descriptor that is missing entirely? That seems like an even bigger bug, and right now throws an exception. We haven't yet seen evidence that there are predictor reports with missing descriptors, but out of an abundance of caution should we have the client throw a warning in that case as well?

bfolie:
> Is the plan to promote this to an error after a backend fix?

I'd like to, but I don't know if we can. There will be existing predictors that stick around with this problem.

bfolie:
nextgen-devkit tests marked as xfail pass when run against this branch

PR #331, opened 2020-05-08
Add predictor responses method

asantas93:
# Citrine Python PR

## Description 
Adds the ability to get predictor responses (descriptors) given a set of inputs.

### PR Type:

### DIFF
--- a/src/citrine/resources/descriptors.py
+++ b/src/citrine/resources/descriptors.py
@@ -29,7 +29,7 @@ class DescriptorMethods:
         Returns
         -------
         List[Descriptor]
-            The computable responses of the predictor given the provided inputs.
+            The computable responses of the predictor given the provided input space (as descriptors).
 
         """
         response = self.session.post_resource(
--- a/src/citrine/resources/descriptors.py
+++ b/src/citrine/resources/descriptors.py
@@ -17,19 +17,20 @@ class DescriptorMethods:
     def from_predictor_responses(self, predictor: Predictor,
                                  inputs: List[Descriptor]) -> List[Descriptor]:
         """
-        [ALPHA] Get responses for a predictor, given a set of inputs.
+        [ALPHA] Get responses for a predictor, given an input space.
 
         Parameters
         ----------
         predictor : Predictor
             The predictor whose available responses are to be computed.
         inputs : List[Descriptor]
-            The inputs to the predictor
+            The input space to the predictor.
 
         Returns
         -------
         List[Descriptor]
-            The computable responses of the predictor given the provided input space (as descriptors).
+            The computable responses of the predictor given the provided input space (as
+            descriptors).
 
         """
         response = self.session.post_resource(
--- a/src/citrine/resources/descriptors.py
+++ b/src/citrine/resources/descriptors.py
@@ -14,7 +14,8 @@ class DescriptorMethods:
         self.project_id = project_id
         self.session: Session = session
 
-    def from_predictor_responses(self, predictor: Predictor, inputs: List[Descriptor]) -> List[Descriptor]:
+    def from_predictor_responses(self, predictor: Predictor,
+                                 inputs: List[Descriptor]) -> List[Descriptor]:
         """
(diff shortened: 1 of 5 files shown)


src/citrine/resources/project.py
```diff
@@ -103,6 +104,11 @@ def predictors(self) -> PredictorCollection:
         """Return a resource representing all visible predictors."""
         return PredictorCollection(self.uid, self.session)
 
+    @property
+    def descriptors(self) -> DescriptorMethods:
```
asantas93, line 108: Does it actually make sense to have to do this in the context of a project?

src/citrine/resources/descriptors.py
```diff
@@ -0,0 +1,40 @@
...
+        inputs The inputs to the predictor
+
+        Returns
+        -------
+        List[Descriptor]
+            The computable responses of the predictor given the provided inputs.
+
+        """
+        response = self.session.post_resource(
+            path='/projects/{}/material-descriptors/predictor-responses'.format(self.project_id),
```
asantas93, line 34: Does it actually make sense to have to do this in the context of a project?

bfolie: We're not using any of the project resources of permissions, so no, the api route does not need to be on the `/projects/` path.

That said, we should decouple what makes sense for API design from what makes sense to show the user. The latter decision should involve product. We did the solution design exercise (I'm sorry for not directing you to this sooner) and the results are here: https://github.com/CitrineInformatics/ai-engine-documentation/blob/master/docs/predictors.md#retrieving-predictor-responses

Our plan was to expose the method as `predictor.get_responses(inputs)`. I realize now that's not so easy because a given `predictor` doesn't necessarily have `Session` information. But we should still try and make the user interface as readable as possible.

bfolie: I think I like `descriptors.from_predictor_responses(predictor, inputs)`. It seems like it will be easy for users to grasp.

maxhutch: What if its a graph predictor that references other predictors by UID? I think we do want to keep just about everything scoped in a project because, even if it doesn't require access checks now it might in the future.

bfolie: Are you suggesting we handle graph predictors with references now, or that we might in the future? I agree with the latter, but to be clear I don't think the backend supports the former right now (but @asantas93 will know for sure)

maxhutch: This type signature doesn't exclude them.  Graph predictors do support references already.  The backend [seems to](https://github.com/CitrineInformatics/platform-backend/pull/1180/files#diff-176be3fadd0613e04391b60237f80672R64).

src/citrine/resources/descriptors.py
```diff
@@ -0,0 +1,40 @@
...
+        self.session: Session = session
+
+    def from_predictor_responses(self, predictor: Predictor,
+                                 inputs: List[Descriptor]) -> List[Descriptor]:
+        """
+        [ALPHA] Get responses for a predictor, given a set of inputs.
+
+        Parameters
+        ----------
+        predictor The predictor whose available responses are to be computed.
```
maxhutch, line 24: I don't think this part of the docstring is formatted quite correctly.  We use [numpy-style](https://numpydoc.readthedocs.io/en/latest/format.html#sections)

src/citrine/resources/descriptors.py
```diff
@@ -0,0 +1,40 @@
...
+        inputs The inputs to the predictor
```
maxhutch, line 25: Its important to note that these are descriptors and not values.

src/citrine/resources/descriptors.py
```diff
@@ -0,0 +1,40 @@
...
+
+        Returns
+        -------
+        List[Descriptor]
+            The computable responses of the predictor given the provided inputs.
```
maxhutch, line 30:
```suggestion
            The computable responses of the predictor given the provided input space (as descriptors).
```
Again, to avoid the confusion between the descriptors and values.

PR #343, opened 2020-05-26
Fixes to allow us to list the ara definitions

dleblanc:
This changes some ara definitions stuff to allow listing of ara definitions.

Specifically:
 - using the proper collection key
 - removing the individual key
 - using properties from both version and definition when building an AraDefinition

Ken's e2e test exercises this code path.

### PR Type:

### DIFF
--- a/src/citrine/resources/ara_definition.py
+++ b/src/citrine/resources/ara_definition.py
@@ -12,7 +12,7 @@ from citrine._session import Session
 from citrine.resources.process_template import ProcessTemplate  # noqa: F401
 from citrine.ara.columns import Column, MeanColumn, IdentityColumn, OriginalUnitsColumn
 from citrine.ara.rows import Row
-from citrine.ara.variables import Variable, IngredientIdentifierByProcessTemplateAndName,\
+from citrine.ara.variables import Variable, IngredientIdentifierByProcessTemplateAndName, \
     IngredientQuantityByProcessAndName, IngredientQuantityDimension
 
 
@@ -208,7 +208,8 @@ class AraDefinitionCollection(Collection[AraDefinition]):
     _path_template = 'projects/{project_id}/ara-definitions'
     _collection_key = 'definitions'
 
-    # NOTE: This isn't actually an 'individual key' - both parts (version and definition) are necessary
+    # NOTE: This isn't actually an 'individual key' - both parts (version and
+    # definition) are necessary
     _individual_key = None
 
     def __init__(self, project_id: UUID, session: Session):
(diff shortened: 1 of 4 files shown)


src/citrine/resources/ara_definition.py
```diff
@@ -41,6 +41,7 @@ class AraDefinition(Resource["AraDefinition"]):
 
     @staticmethod
     def _get_dups(lst: List) -> List:
+        # Hmmn, this looks like a potentially costly operation?!
         return [x for x in lst if lst.count(x) > 1]
```
kroenlein, line 45:
```suggestion
    def _get_dups(lst: List) -> List:
        seen = set()
        dups = [x for x in lst if x in seen or seen.add(x)]  # seen.add returns None
        return dups
```
But perhaps this is an off-topic optimizaton

src/citrine/resources/ara_definition.py
```diff
@@ -221,10 +225,10 @@ def get_with_version(self, definition_uid: Union[UUID, str],
 
     def build(self, data: dict) -> AraDefinition:
         """[ALPHA] Build an individual Ara Definition from a dictionary."""
-        defn = AraDefinition.build(data['ara_definition'])
-        defn.definition_uid = data['definition_id']
-        defn.version_number = data['version_number']
-        defn.version_uid = data['id']
+        defn = AraDefinition.build(data['version']['ara_definition'])
```
kroenlein, line 228: Would it make sense for `version` to be a constant, given the number of times it appears?  I was originally thinking just use `_individual_key`, but that got cleared in this PR.

src/citrine/resources/ara_definition.py
dleblanc: I agree that's a better approach. I'm going to leave it out of the PR for now just to keep focused.

src/citrine/resources/ara_definition.py
dleblanc: I'll just extract those parts of the dict - it's only in this scope, and used 3 times.

dleblanc:
> Makes sense if it clears tests. Did you identify why we were incorrectly getting clean bills of health?

All the interactions are faked out in the unit tests, and they were representing a different JSON schema than what actually gets returned by the real API. In addition they were doing some decode->re-encode logic that doesn't work given that the AraDefinition munges bits of information from both the Ara 'definition', the 'version', as well as the 'ara_definition' JSON blob. That's not great, but a larger issue and not something I intend to fix here.

I've fixed them to represent the schema properly, and to not bother with the decode/re-encode business. I've also removed some overspecification in the tests and added some relevant asserts on the returned ara definitions.

dleblanc:
Rebuild!

dleblanc:
Rebuild!

tests/resources/test_ara_definition.py
```diff
@@ -241,81 +243,89 @@ def test_add_all_ingredients(session, project):
...
+    registered = collection.register(ara_definition)
 
     # Then
-    assert 1 == session.num_calls
-    expect_call = FakeCall(
-        method="POST",
-        path="projects/{}/ara-definitions".format(project_id),
-        json={"definition": collection.build(ara_definition).dump()}
-    )
-    assert session.last_call == expect_call
```
kroenlein, line 262: Good clean-up

PR #346, opened 2020-05-27
Feature/pla 3507 batch write

asantas93:
# Citrine Python PR

## Description 
Adds new batch write method for data concepts collections, `register_all`.

Note that there will likely be changes to API limits for batch size and to the validation error response structure in the near future. What those changes will be is unclear at the moment.

My initial testing shows a reduction in ingest time for the breads dataset (308 objects of mixed types) from 53s to 6s, and a reduction of write time for 100 empty process specs from 17s to 0.67s.

### PR Type:

### DIFF
--- a/tests/resources/test_material_run.py
+++ b/tests/resources/test_material_run.py
@@ -42,6 +42,19 @@ def test_register_material_run(collection, session):
     assert "<Material run 'Test MR 123'>" == str(registered)
 
 
+def test_register_all(collection, session):
+    runs = [MaterialRunFactory(name='1'), MaterialRunFactory(name='2'),  MaterialRunFactory(name='3')]
+    session.set_response({'objects': [r.dump() for r in runs]})
+    registered = collection.register_all(runs)
+    assert [r.name for r in runs] == [r.name for r in registered]
+    assert len(session.calls) == 1
+    assert session.calls[0].method == 'PUT'
+    assert session.calls[0].path == 'projects/{}/datasets/{}/material-runs/batch'.format(
+        collection.project_id, collection.dataset_id)
+    with pytest.raises(RuntimeError):
+        MaterialRunCollection(collection.project_id, dataset_id=None, session=session).register_all([])
+
+
 def test_dry_run_register_material_run(collection, session):
     # Given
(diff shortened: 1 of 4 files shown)


tests/resources/test_material_run.py
```diff
@@ -42,6 +42,19 @@ def test_register_material_run(collection, session):
     assert "<Material run 'Test MR 123'>" == str(registered)
 
 
+def test_register_all(collection, session):
+    runs = [MaterialRunFactory(name='1'), MaterialRunFactory(name='2'),  MaterialRunFactory(name='3')]
```
ventura-rivera, line 46: What if the materials are linked to processes (using gemd-python objects). If you use register_all on the processes first, will the uid info for the processes be updated on the materials?
I basically don't want to run into this problem:
https://citrine.atlassian.net/browse/PLA-3935

ventura-rivera: This tests checks for that situation on `dataset.register_all`:
https://github.com/CitrineInformatics/citrine-python/blob/master/tests/resources/test_dataset.py#L301

asantas93: I did not code in that functionality. It's definitely something I could add, although I'm not sure I like the idea of the client having to manually resolve the differences between objects pre-write and post-write (more than just `uids` change), or of this method having side-effects on its inputs. It would be nice if users could just use the returned list, which is in the same order as its inputs so the association from input to output is easy to make.

asantas93: I just looked at this issue again, and I don't think this is a result of not setting uids client-side. I mean, it is because of the way our normalization/serialization logic works, but I think the underlying cause is that we add and then remove the `auto` uid client-side. Then there's no uid on the process and the material referencing it is serialized, our link swapper just ignores the process instead of complaining that it can't make a link when no uid is present. If we generated ids that we didn't subsequently drop, this wouldn't be a problem.

asantas93:
Merged this PR, but @ventura-rivera we should talk about how to address this class of issues more generally.

ventura-rivera:
@asantas93 All good, I think I understand what I can do with it for now. Excited to try it out myself! I can set up a meeting with you to talk about this

PR #347, opened 2020-05-28
Documented how to retrieve descriptors from simple mixture predictors.

latture:
This PR documents the responses of simple mixture predictors and how to retrieve them using a project.


latture:
@maxhutch @gregor-robinson I forgot I had this open. Could I get a quick review, so we can close this PR?

PR #348, opened 2020-05-28
Bugfix/remove enum34

kroenlein:
# Citrine Python PR

## Description 
The primary goal of this PR to making citrine-python compatible with the lastest version of GEMD.
* It removes the reference to enum34 in its Pipfile.lock.
* It updates a reference to a flatten method in a test that uses a now broken interface. 

### PR Type:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,7 @@ class PreBuildCommand(build_py):
 
 
 setup(name='citrine',
-      version='0.30.0',
+      version='0.30.1',
       url='http://github.com/CitrineInformatics/citrine-python',
       description='Python library for the Citrine Platform',
       author='Citrine Informatics',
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -57,7 +57,7 @@ def run(self):
 
 
 setup(name='citrine',
-      version='0.30.0',
+      version='0.30.1',
```
kroenlein, line 60: Or should this be 0.31.0, since it's adapting to breaking changes in GEMD?

tests/_serialization/test_taurus_interop.py
```diff
@@ -19,5 +19,5 @@ def test_flatten():
     )
     spec = ProcessSpec(name="spec", template=template)
 
-    flat = flatten(spec)
+    flat = flatten(spec, scope='testing')
```
kroenlein, line 22: Calls to flatten now require an explicit choice of scope for dissecting the tree.

setup.py
maxhutch: I'd be conservative and make it 0.31.0 for that reason, since we expect users to reference GEMD classes directly (e.g. for value types).

PR #349, opened 2020-06-01
Update dataset.register_all to use data_concepts register_all

asantas93:
# Citrine Python PR

## Description 
Improves performance of `dataset.register_all` by using the `register_all` method on the data concepts collection.

### PR Type:

### DIFF
--- a/src/citrine/resources/dataset.py
+++ b/src/citrine/resources/dataset.py
@@ -1,4 +1,5 @@
 """Resources that represent both individual and collections of datasets."""
+from collections import defaultdict
 from typing import TypeVar, List
 from uuid import UUID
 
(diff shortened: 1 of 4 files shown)


src/citrine/resources/dataset.py
```diff
@@ -252,12 +252,22 @@ def register_all(self, data_concepts_resources: List[ResourceType],
 
         """
         resources = list()
-        for resource in (sorted(data_concepts_resources,
-                                key=lambda resource: writable_sort_order(resource.typ))):
-            registered_resource = self.register(resource, dry_run)
-            if isinstance(registered_resource, BaseEntity):
-                resource.uids = registered_resource.uids
-            resources.append(resource)
+        by_type = {}
```
maxhutch, line 255: You can use a [defaultdict](https://docs.python.org/3.6/library/collections.html#collections.defaultdict) here

src/citrine/resources/dataset.py
```diff
@@ -252,12 +252,22 @@ def register_all(self, data_concepts_resources: List[ResourceType],
...
+        for obj in data_concepts_resources:
+            if obj.typ not in by_type:
+                by_type[obj.typ] = []
+            by_type[obj.typ].append(obj)
+        typ_groups = sorted(list(by_type.values()), key=lambda x: writable_sort_order(x[0]))
+        batch_size = 50
+        for typ_group in typ_groups:
+            num_batches = len(typ_group) // batch_size
+            for batch_num in range(num_batches + 1):
+                batch = typ_group[batch_num: batch_num + batch_size]
```
maxhutch, line 265:
```suggestion
                batch = typ_group[batch_num * batch_size: (batch_num + 1)  * batch_size]
```

asantas93: oof

asantas93:
> LGTM.
> 
> Do we have e2e test coverage of this method? The fact that the indexing issue still passed tests suggests we're not getting much out of the mocks here, so we should be sure it gets verified.

Maybe one of the table builder tests would fail. I can add a more explicit case for this.

PR #351, opened 2020-06-02
Remove CitrineID from object constructor

kroenlein:
# Citrine Python PR

## Description 
This PR is to change default behavior of citrine-python objects derived from gemd-python BaseEntities so that a CitrineID is only populated immediately prior to sending the object to the server (or, more generally, when we try to serialize things).  This should bring behavior in gemd-python and citrine-python more in line.

### PR Type:

### DIFF
--- a/src/citrine/resources/condition_template.py
+++ b/src/citrine/resources/condition_template.py
@@ -46,10 +46,12 @@ class ConditionTemplate(AttributeTemplate, Resource['ConditionTemplate'], GEMDCo
     def __init__(self,
                  name: str,
                  bounds: BaseBounds,
-                 uids: Optional[Dict[str, str]] = dict(),
+                 uids: Optional[Dict[str, str]] = None,
                  description: Optional[str] = None,
                  tags: Optional[List[str]] = None
                  ):
+        if uids is None:
+            uids = dict()
         DataConcepts.__init__(self, GEMDConditionTemplate.typ)
         GEMDConditionTemplate.__init__(self, name=name, bounds=bounds, tags=tags,
                                        uids=uids, description=description)
--- a/src/citrine/resources/condition_template.py
+++ b/src/citrine/resources/condition_template.py
@@ -5,7 +5,6 @@ from citrine._rest.resource import Resource
 from citrine._serialization.properties import List as PropertyList
 from citrine._serialization.properties import Optional as PropertyOptional
 from citrine._serialization.properties import String, Mapping, Object
-from citrine._utils.functions import set_default_uid
 from citrine.resources.attribute_templates import AttributeTemplate, AttributeTemplateCollection
 from citrine.resources.data_concepts import DataConcepts
 from gemd.entity.bounds.base_bounds import BaseBounds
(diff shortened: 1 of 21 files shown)


src/citrine/resources/condition_template.py
```diff
@@ -47,13 +46,13 @@ class ConditionTemplate(AttributeTemplate, Resource['ConditionTemplate'], GEMDCo
     def __init__(self,
                  name: str,
                  bounds: BaseBounds,
-                 uids: Optional[Dict[str, str]] = None,
+                 uids: Optional[Dict[str, str]] = dict(),
```
kroenlein, line 49: This pattern appears repeatedly.  It is necessary because citrine._serialization.properties actually overloads the uids setter and requires a dictionary rather than an Optional pattern as other fields have.  Addressable, but not the immediate purpose of this PR.

src/citrine/resources/data_concepts.py
```diff
@@ -363,6 +365,8 @@ def register(self, model: ResourceType, dry_run=False):
         # know how to replace the objects with link-by-uids. loads() converts this string into
         # nested gemd objects, and then the final dumps() converts that to a json-ready string
         # in which all of the object references have been replaced with link-by-uids.
+
+        GEMDJson(scope=CITRINE_SCOPE).dumps(model)  # This apparent no-op populates uids
```
kroenlein, line 369: Structurally, it might make sense to modify gemd.entity.dict_serializable so that dump generates an object with an ID, but there is value in the non-mutating dump method.

tests/_util/test_functions.py
```diff
@@ -5,22 +5,12 @@
...
-def test_set_default_id_non_uuid():
-    with pytest.raises(ValueError):
-        set_default_uid({'id': 'not-valid'})
-
-
-def test_set_default_id_non_dict():
-    with pytest.raises(Exception):
-        set_default_uid('not a dict')
-
-
```
kroenlein, line 23: The method that this was testing has been removed.

tests/serialization/test_material_run.py
```diff
@@ -48,7 +48,7 @@ def test_serialization():
 def test_process_attachment():
     """Test that a process can be attached to a material, and that the connection survives serde"""
     cake = MaterialRun('Final cake')
-    cake.process = ProcessRun('Icing')
+    cake.process = ProcessRun('Icing', uids={'id': '12345'})
     cake_data = cake.dump()
```
kroenlein, line 52: The terrible UID here is possibly the best argument for modifying dict_serializable.

kroenlein:
https://jenkins.corp.citrine.io/job/nextgen-ITs/5472/

src/citrine/resources/data_concepts.py
asantas93: Non-blocking: I would still prefer that we just call a method specifically for setting IDs before registering. IMO serialization should not have side-effects and we should just make it easy for the user to meet all the requirements for preserving links through serialization with explicit convenience methods.

kroenlein: I generally agree, except I don't know if there's actually a clean solution here.  In either case, this requires a change to the external library.

src/citrine/resources/data_concepts.py
```diff
@@ -397,6 +401,10 @@ def register_all(self, models: List[ResourceType], dry_run=False) -> List[Resour
             raise RuntimeError("Must specify a dataset in order to register a data model object.")
         path = self._get_path()
         params = {'dry_run': dry_run}
+
+        json = GEMDJson(scope=CITRINE_SCOPE)  # This apparent no-op populates uids
```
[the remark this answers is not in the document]
kroenlein: It was intentional, but probably louder than in needed to be.

PR #354, opened 2020-06-04
Make MIME types insensitive to capitals in file extension

kroenlein:
# Citrine Python PR

## Description 
At present, a capitalized file extension (e.g., .JPG) causes a key error as the MIME type dictionary in the mimetypes package is all lower case.  This patches this immediate issue, though leaves open the issue of throwing exceptions if a file extension doesn't align with our expectation.

I've added an XFail for the latter problem.  Maybe we should just change the invocation to a `get` with default `'application/octet-stream'`, but I don't feel comfortable executing on that without discussion.

### PR Type:

### DIFF
--- a/tests/resources/test_file_link.py
+++ b/tests/resources/test_file_link.py
@@ -95,7 +95,7 @@ def test_upload(mock_isfile, mock_stat, mock_open, mock_boto3_client, collection
     mock_open.return_value.__enter__.return_value = 'Random file contents'
     mock_boto3_client.return_value = FakeS3Client({'VersionId': '3'})
 
-    dest_name = 'foo.txt'
+    dest_names = ['foo.txt', 'foo.TXT']  # Verify that capitalization in extension is fine
     file_id = '12345'
     version = '13'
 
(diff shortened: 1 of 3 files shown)


tests/resources/test_file_link.py
```diff
@@ -95,7 +95,59 @@ def test_upload(mock_isfile, mock_stat, mock_open, mock_boto3_client, collection
...
+        ]
+    }
+
+    for dest_name in dest_names:
+        session.set_responses(uploads_response, file_info_response)
+        file_link = collection.upload(dest_name)
+
+        url = 'projects/{}/datasets/{}/files/{}/versions/{}'\
+            .format(collection.project_id, collection.dataset_id, file_id, version)
+        assert file_link.dump() == FileLink(dest_name, url=url).dump()
```
bfolie, line 131: How is this testing that the extension is correctly read? I don't see any check of the `upload_json` structure. Or is the test that this simple occurs without throwing an exception, because previously there would have been an exception on `mimetypes.types_map['TXT']`?

kroenlein: Yes, it's just a test that capitalization no longer fails.  I could add an `assert mimetype1 == mimetype2` test following if explicit case insensitivity makes more sense, but partly this is a band-aid to better typing approaches.

kroenlein: I dove in, and (presuming I'm not missing anything) implementing that test would involve a bunch of dark art at the least.

kroenlein:
https://citrine.atlassian.net/browse/PLA-4395

PR #355, opened 2020-06-05
Feature/pla 4305 xor gem table variables

asantas93:
# Citrine Python PR

## Description 
Adds new XOR variable for GEM table building. Also fixes bug where `IngredientQuantityInOutput` would serialize with the type hint of `IngredientIdentifierInOutput`, which may have cause the backend to deserialize an instance of the latter rather than the former.

### PR Type:

### DIFF
--- a/src/citrine/ara/variables.py
+++ b/src/citrine/ara/variables.py
@@ -640,7 +640,7 @@ class XOR(Serializable['XOR'], Variable):
     typ = properties.String('type', default="xor", deserializable=False)
 
     def _attrs(self) -> List[str]:
-        return ["name", "headers", "variables"]
+        return ["name", "headers", "variables", "typ"]
 
     def __init__(self, *, name, headers, variables):
         self.name = name
--- a/src/citrine/ara/variables.py
+++ b/src/citrine/ara/variables.py
@@ -37,12 +37,9 @@ class Variable(PolymorphicSerializable['Variable']):
         pass  # pragma: no cover
 
     def __eq__(self, other):
-        try:
-            return type(self) == type(other) and all([
-                getattr(self, key) == getattr(other, key) for key in self._attrs()
-            ])
-        except AttributeError:
-            return False
+        return type(self) == type(other) and all([
+            getattr(self, key) == getattr(other, key) for key in self._attrs()
+        ])
 
     @classmethod
     def get_type(cls, data) -> Type[Serializable]:
--- a/src/citrine/ara/variables.py
+++ b/src/citrine/ara/variables.py
@@ -54,7 +54,7 @@ class Variable(PolymorphicSerializable['Variable']):
             AttributeByTemplateAndObjectTemplate, IngredientIdentifierByProcessTemplateAndName,
             IngredientLabelByProcessAndName, IngredientQuantityByProcessAndName,
             RootIdentifier, AttributeInOutput, IngredientIdentifierInOutput,
-            IngredientQuantityInOutput
+            IngredientQuantityInOutput, XOR
         ]
         res = next((x for x in types if x.typ == data["type"]), None)
         if res is None:
@@ -588,7 +588,7 @@ class IngredientQuantityInOutput(Serializable['IngredientQuantityInOutput'], Var
     ingredient_name = properties.String('ingredient_name')
     quantity_dimension = properties.Enumeration(IngredientQuantityDimension, 'quantity_dimension')
     process_templates = properties.List(properties.Object(LinkByUID), 'process_templates')
-    typ = properties.String('type', default="ing_id_in_output", deserializable=False)
+    typ = properties.String('type', default="ing_quantity_in_output", deserializable=False)
 
     def _attrs(self) -> List[str]:
         return ["name", "headers", "ingredient_name", "process_templates", "typ"]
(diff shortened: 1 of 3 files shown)


src/citrine/ara/variables.py
```diff
@@ -588,7 +588,7 @@ class IngredientQuantityInOutput(Serializable['IngredientQuantityInOutput'], Var
     ingredient_name = properties.String('ingredient_name')
     quantity_dimension = properties.Enumeration(IngredientQuantityDimension, 'quantity_dimension')
     process_templates = properties.List(properties.Object(LinkByUID), 'process_templates')
-    typ = properties.String('type', default="ing_id_in_output", deserializable=False)
+    typ = properties.String('type', default="ing_quantity_in_output", deserializable=False)
```
kroenlein, line 591: Drive-by bug, right?  Is worth breaking it out to figure out why it didn't have test coverage?

src/citrine/ara/variables.py
```diff
@@ -604,3 +604,45 @@ def __init__(self, *,
...
+
+    XOR can only operate on inputs with the same output type. For example, you may XOR
+    :class:`~citrine.ara.variables.RootIdentifier` with
+    :class:`~citrine.ara.variables.IngredientIdentifierByProcessTemplateAndName` because they both
+    produce simple strings, but not with :class:`~citrine.ara.variables.IngredientQuantityInOutput`
+    which produces a real numeric quantity.
+
+    The input variables to XOR need not exist elsewhere in the table definition, and the name and
+    headers assigned to them have no bearing on how the table is constructed. That they are
+    required at all is simply a limitation of the current API.
```
kroenlein, line 624: Well written and complete.

src/citrine/ara/variables.py
```diff
@@ -604,3 +604,45 @@ def __init__(self, *,
...
+
+    name = properties.String('name')
+    headers = properties.List(properties.String, 'headers')
+    variables = properties.List(properties.Object(Variable), 'variables')
+    typ = properties.String('type', default="xor", deserializable=False)
+
+    def _attrs(self) -> List[str]:
+        return ["name", "headers", "variables", "typ"]
+
+    def __init__(self, *, name, headers, variables):
```
kroenlein, line 645: As a dodge, can we put defaults in here?  Or would overlapping names collide in later processing?  If `name` and `headers` are no-op arguments, we could bypass the need for people to do bookkeeping.

tests/ara/test_variables.py
```diff
@@ -8,6 +8,7 @@
 
 @pytest.fixture(params=[
     RootInfo(name="root name", headers=["Root", "Name"], field="name"),
+    XOR(name="root name or sample_type", headers=["Root", "Info"], variables=[RootInfo(name="root name", headers=["Root", "Name"], field="name"), RootInfo(name="root name", headers=["Root", "Sample Type"], field="sample_type")]),
```
kroenlein, line 11: This technically checks the functionality, but it sounds like a pretty implausible scenario.

src/citrine/ara/variables.py
asantas93: Are you asking if we can derive `name` and `headers` from `variables`?

src/citrine/ara/variables.py
asantas93: hmm I think I can probably add a serde test that would have detected this issue. Let me see.

asantas93: https://github.com/CitrineInformatics/citrine-python/blob/3ea5cfb5324b4623e0e8953c78c4234f9dccf84c/tests/ara/test_variables.py#L35 was actually the test I was planning to write, but it wasn't catching the issue because the one type's field's was a subset of the other and the `Variable.__eq__` method had different behavior depending on which variable is on the left/right. https://github.com/CitrineInformatics/citrine-python/blob/1dcc8c10013c127a5df8287bfd4e7402f192b753/src/citrine/ara/variables.py#L42

asantas93: The new `__eq__` method in my latest commit also checks types, which caused a test failure without the type hint update in this PR.

src/citrine/ara/variables.py
```diff
@@ -37,12 +37,9 @@ def _attrs(self) -> List[str]:
         pass  # pragma: no cover
 
     def __eq__(self, other):
-        try:
-            return all([
-                self.__getattribute__(key) == other.__getattribute__(key) for key in self._attrs()
-            ])
-        except AttributeError:
-            return False
+        return type(self) == type(other) and all([
```
kroenlein, line 40: Do we want strict type equality?  What if we subclass?

asantas93: Would we want to different child types to be able to be equal?

asantas93: There are variables with the exact same data other than type hint right now. The type is definitely a distinguishing factor right now.

asantas93: Sorry, that isn't accurate. I have been documenting the *column* definitions recently and it is true of them. For variables, there is no possible for exact JSON equality when you remove the type hint as of right now, but I wouldn't be surprised if we did run into that in the future since there is a lot of attribute overlap, and these `Variable` classes really model functions rather than data.

src/citrine/ara/variables.py
kroenlein: Sorry, I had this backwards in my brain, and we thinking we could avoid the issue where the `variables` we are feeding in required names and headers for no good reason.  Ignore.

src/citrine/ara/variables.py
kroenlein: Excellent.

PR #357, opened 2020-06-10
Pla 4404/joint design space

Imperssonator:
# Citrine Python PR

## Description 
Add builders to construct enumerated design spaces that mix simple mixture DoFs and independent processing DoFs.

### PR Type:


src/citrine/builders/design_spaces.py
```diff
@@ -0,0 +1,96 @@
+import pandas as pd
+from itertools import product
+
+def EnumeratedProductDesignSpaceData(dof_grids:dict):
```
maxhutch, line 4:
```suggestion
def enumerate_data_from_product(dof_grids:dict):
```
function names should be `snake_case` and verb-y rather than noun-y.

maxhutch: What is the utility of this method over the current product design space?

src/citrine/builders/design_spaces.py
```diff
@@ -0,0 +1,96 @@
...
+        {'degree of freedom':[list, of, values]} for each DoF
+    '''
+    design_space_tuples = list(product(*dof_grids.values()))
+    design_space_cols = list(dof_grids.keys())
+    df_ds = pd.DataFrame(data=design_space_tuples, columns=design_space_cols)
+    candidates = df_ds.to_dict('records')
+    return candidates
+
+
+def SimpleMixtureProductDesignSpaceData(formulation_grids:dict,
```
maxhutch, line 21:
```suggestion
def enumerate_simple_mixture_from_ingredients(formulation_grids:dict,
```
Again, `snake_case` and verb-y.

src/citrine/builders/design_spaces.py
```diff
@@ -0,0 +1,96 @@
...
+                                        balance_component:str
+                                        ):
```
maxhutch, line 23:
```suggestion
                                        ) -> List[Mapping[str, Any]]:
```
I _think_ this is the right type.

src/citrine/builders/design_spaces.py
```diff
@@ -0,0 +1,96 @@
...
+    ''' Given a dict of lists or tuples of ingredient amounts (must be
+    fractions of some kind, i.e. between 0-1), create
+    a DataFrame that is the outer product of all possible combinations of
+    ingredients *except* for the balance_component. The balance_component
+    will then be filled in as 1-(all other ingredients) for each row.
+    
+    Many points generated through this process will be invalid (negative
+    balance ingredient amount). These rows are filtered out as a last step,
+    if the balance ingredient amount is below the lowest value in its list
```
maxhutch, line 32: This is a bit confusing to me.  Why include the balance component in the grids if its value is dicated by the others?  How about `balance_minimum: float = 0` and `balance_maximum: float = 1` arguments to make clear that the balance component quantity could lie anywhere between the min and the max?

src/citrine/builders/design_spaces.py
```diff
@@ -0,0 +1,96 @@
...
+                     max(formulation_grids[balance_component])
+    )
+    form_ds = form_ds[form_ds[balance_component]>=balance_range[0]]
+    form_ds = form_ds[form_ds[balance_component]<=balance_range[1]]
+    data = form_ds.to_dict('records')
+
+    return data
+
+
+def JointDesignSpaceData(data_list:list):
```
maxhutch, line 78:
```suggestion
def build_outer_product_of_candidates(data_list:list):
```
I'm not sure if `Joint` or `Product` or `OuterProduct` is more clear.  The language we converged on for the on-platform solution is `Product` (as an simple extension of the univariate product currently available).  See [ai-engine-documentation](https://github.com/CitrineInformatics/ai-engine-documentation/blob/master/docs/design-spaces.md#productdesignspace-modifications-to)

src/citrine/builders/design_spaces.py
```diff
@@ -0,0 +1,96 @@
...
+    '''
+    
+    ds_list = [pd.DataFrame(data) for data in data_list]
+    
+    for df in ds_list:
+        df['join_key']=0
+    
+    df_join=ds_list[0]
+    for df in ds_list[1:]:
+        df_join=pd.merge(df_join,df,on='join_key')
```
maxhutch, line 93: This logic definitely needs a test.

src/citrine/builders/design_spaces.py
bfolie: I remember having this discussion and deciding that "outer product" is not technically correct. "Cartesian product" is closer.

src/citrine/builders/design_spaces.py
Imperssonator: None on its own, but this one can be combined with other design spaces, and used to make Cartesian product design spaces within other functions (see the simple mixture DS)

src/citrine/builders/design_spaces.py
Imperssonator: 1. You might want to change the balance component, this allows that to be done much more easily
2. It keeps all the ingredients together in a single dict, which just feels nicer
3. You can just pass (min, max) as the dict value for the balance component - fewer arguments

I would rather clarify the docstring than break out the balance component...

src/citrine/builders/design_spaces.py
Imperssonator: ^Right, I remember that too. The real difficulty here is that the first function, the regular old Product Design Space, is the Cartesian Product of its dimensions... this function is also a Cartesian product, but it's like a meta-Cartesian product. You're passing in design spaces that themselves could already be Cartesian Products.

`build_cartesian_product_from_subspaces`....

src/citrine/builders/design_spaces.py
Imperssonator: How about for the input (`formulation_grids`)? A function call looks like:

```
form_ds = SimpleMixtureProductDesignSpaceData({
    'balance_comp':[0.3,0.9],
    'comp1':[0,0.01,0.02,0.03,0.04,0.05],
    'comp2':[0.1,0.11,0.12,0.13,0.14,0.15],
    'comp3':[0.1,0.15,0.2,0.25,0.3],
    'comp4':[0,0.01,0.02,0.03,0.04,0.05]
},
balance_component='balance_comp')
```

Imperssonator: Mapping[str, Sequence], right?

Imperssonator:
I updated the function calls to be more `verb-like`.

I'm not sure about having these return an EnumeratedDesignSpace. I like that these enable the user to interact directly with the candidates, display a dataframe in-line so they can check correctness, etc. It's trivial for them to take the output of any of these functions and pass the result to `data` of an `EumeratedDesignSpace`. Wouldn't that also kind of pollute the core API with a bunch of custom enumerated design space types?

src/citrine/builders/design_spaces.py
maxhutch: The thing after the `-> ` is the type of the return value..  How about `Mapping[str, List[float]]` for `formulation_grids`?

maxhutch:
> I'm not sure about having these return an EnumeratedDesignSpace. I like that these enable the user to interact directly with the candidates, display a dataframe in-line so they can check correctness, etc. 

Why not define `toDataframe(design_space: EnumeratedDesignSpace)` so they can parse the candidates of any enumerated design space?

> Wouldn't that also kind of pollute the core API with a bunch of custom enumerated design space types?

These are all just one type: `EnumeratedDesignSpace`.  That's the power of the builder pattern: you have multiple ways to construct an object, but, once you've constructed it, its state is always represented in the same way.  Think multiple constructors but more explicit.

src/citrine/builders/design_spaces.py
bfolie: The ease-of-change is compelling, as is limiting the number of arguments.

Would it be useful to also print a warning message if the balance component maps to a list with the number of elements != 2? Something like "Warning: balance component 'water' mapped to [0.4, 0.5, 0.6], which has more than 2 elements. The min and max values are used to constrain the allowed amount of 'water'; all other values will be ignored."

Imperssonator: ^Yes

src/citrine/builders/design_spaces.py
Imperssonator: Yep, note that I have updated this with a new commit

Imperssonator:
@maxhutch @bfolie round-up of things that sound like current asks to get this approved...

1. settle on names. Now they are snake_case and verb-y and use 'cartesian' instead of 'outer' when products are being talked about. I've got `enumerate_cartesian_product`, `enumerate_simple_mixture_cartesian_product`, and `cartesian_join_design_spaces` (should I add `enumerate_` to the start of that one too?)

2. I believe I'm now fully flake8 compliant. VSCode does not have any more red lines.

3. Do you really want these functions to return `EnumeratedDesignSpace` objects rather than just candidate lists that can be fed in as `data`? That would have a couple implications: (1) (minor) `name` and `description` would be added as pass-through arguments (and what happens when I join two design spaces with different names?), and (2) (pretty involved) `descriptors` would become a (required?) pass-through argument... This adds extra machinery to be able to use these functions. I like how light-weight they are now. I would rather the user provide their list of Descriptors when they've finished joining all their design spaces and pass the `data` to a single `EnumeratedDesignSpace` at the end of the process. The intent of these functions is that the intermediate results might not be your intended end result, so it's odd to be creating full-on EnumeratedDesignSpaces at every step of the process.

4. Is it on me to write a test for the join-logic in `cartesian_join_design_spaces`? I don't know where/how to begin doing that.

Imperssonator:
Oh and 5. I still need to add a warning for the `balance_component` argument to the simple mixture function

maxhutch:
> 1. settle on names. Now they are snake_case and verb-y and use 'cartesian' instead of 'outer' when products are being talked about. I've got `enumerate_cartesian_product`, `enumerate_simple_mixture_cartesian_product`, and `cartesian_join_design_spaces` (should I add `enumerate_` to the start of that one too?)

Those names seem reasonable; we can review them again at the end.

> 2. I believe I'm now fully flake8 compliant. VSCode does not have any more red lines.

There are still failures.  You can see them in the travis build logs, e.g. [here](https://travis-ci.org/github/CitrineInformatics/citrine-python/jobs/700331596#L532).

> 3. Do you really want these functions to return `EnumeratedDesignSpace` objects rather than just candidate lists that can be fed in as `data`? That would have a couple implications: (1) (minor) `name` and `description` would be added as pass-through arguments (and what happens when I join two design spaces with different names?), and (2) (pretty involved) `descriptors` would become a (required?) pass-through argument... This adds extra machinery to be able to use these functions. I like how light-weight they are now. I would rather the user provide their list of Descriptors when they've finished joining all their design spaces and pass the `data` to a single `EnumeratedDesignSpace` at the end of the process. The intent of these functions is that the intermediate results might not be your intended end result, so it's odd to be creating full-on EnumeratedDesignSpaces at every step of the process.

I still feel that they should.  In the case of the simple mixtures, we know what the descriptors are so populating them is actually helpful.  In the case of the product/join, we'd get them on the inputs so there's nothing else the user needs to specify.  Its true that they'd be new for the enumerated product, but the user would need to add them later so its just order of operations in my mind.  Name and description would be pass-through, as you said.  This should make it easier to download a design space, use it in a product, and then immediately upload the result.

> 4. Is it on me to write a test for the join-logic in `cartesian_join_design_spaces`? I don't know where/how to begin doing that.

Yes.  Add a file `tests/builders/design_spaces.py` and put the tests there. Most of the tests in citrine-python so far target http or json.  You can find an example of a meatier logical test [here in gemd-python](https://github.com/CitrineInformatics/gemd-python/blob/master/gemd/util/tests/test_substitute_links.py).  Note that you'll need to cover 100% of the source lines with tests (the test runner will check that as soon as the sources are used at all in the tests).

src/citrine/builders/design_spaces.py
```diff
@@ -1,40 +1,59 @@
...
 
 def enumerate_cartesian_product(
-    dof_grids: Mapping[str, Sequence],
-) -> Sequence[Mapping[str, Any]]:
-    '''[ALPHA] Given a dict of lists or tuples of values for each degree of freedom,
+    design_grids: Mapping[str, Sequence],
+    descriptors: List[Descriptor],
+    name: str = 'Enumerated Cartesian product design space',
+    description: str = '',
+) -> EnumeratedDesignSpace:
```
Imperssonator, line 13: @maxhutch I'm making `name` and `description` kwargs with defaults to keep these functions lightweight. `descriptors` however is required - but I do not perform any kind of checks to ensure the descriptors are correct.

src/citrine/builders/design_spaces.py
```diff
@@ -55,19 +74,39 @@ def enumerate_simple_mixture_cartesian_product(
...
         "balance component must be in formulation_grids' keys"
     non_balance_comps = [comp for comp in list(formulation_grids.keys())
                          if comp != balance_component]
+    non_balance_descriptors = [dd for dd in descriptors 
+                               if dd.key != balance_component]
+    non_balance_grids = {
+        k: v for k, v in formulation_grids.items()
+        if k in non_balance_comps
+        }
+
```
Imperssonator, line 97: A bit of additional overhead here in order to use the `enumerate_cartesian_product()` function within this simple mixture function (need to create dummy descriptor list)

src/citrine/builders/design_spaces.py
```diff
@@ -87,36 +126,57 @@ def enumerate_simple_mixture_cartesian_product(
...
     
+    # Set a dummy column to do the join
     for df in ds_list:
         df['join_key'] = 0
     
+    # Perform the join
     df_join = ds_list[0]
     for df in ds_list[1:]:
         df_join = pd.merge(df_join, df, on='join_key')
+
```
Imperssonator, line 169: @maxhutch this is the logic we would need to replace from pandas. Very simple outer join...

tests/builders/test_design_spaces.py
```diff
@@ -0,0 +1,74 @@
...
+        'gamma': ['a','b','c']
+    }
+    basic_space = enumerate_cartesian_product(
+        design_grids=design_grids,
+        descriptors=[alpha, beta, gamma],
+        name='basic space',
+        description=''
+    )
+    return basic_space
+
```
Imperssonator, line 26: making a basic enumerated space here as a fixture

tests/builders/test_design_spaces.py
```diff
@@ -0,0 +1,74 @@
...
+    joint_space = cartesian_join_design_spaces(
+        design_space_list=ds_list,
+        name='Joined enumerated design space',
+        description='',
+    )
+    return joint_space
+
+
+def test_enumerated_cartesian(basic_cartesian_space):
+    """Check data length"""
```
Imperssonator, line 62: @maxhutch I noticed this in e.g.
[the original design_spaces tests](https://github.com/CitrineInformatics/citrine-python/blob/master/tests/informatics/test_design_spaces.py#L32)
the test function has one of the fixtures listed as an argument... with no `()` after it... then within the test function the variable appears as if it is the output of actually running the fixture function... is this a common pattern?

maxhutch: Yes, [that is the way](https://docs.pytest.org/en/2.8.7/fixture.html).

src/citrine/builders/design_spaces.py
```diff
@@ -0,0 +1,190 @@
...
+    Parameters
+    ----------
+    design_space_list: List[EnumeratedDesignSpace]
+        A list of EnumeratedDesignSpaces
+
+    """
+    assert 'join_key' not in [ds.data[0].keys() for ds in design_space_list], \
+        '"join_key" cannot be a descriptor key when using this function'
+
```
maxhutch, line 164: What happens if some of the design spaces share descriptor keys?  If that's not allowed, then we should check it and raise a `ValueError` if they do.

tests/builders/test_design_spaces.py
```diff
@@ -0,0 +1,76 @@
...
+    assert len(basic_cartesian_space.data) == 45
+
+
+def test_simple_mixture(simple_mixture_space):
+    """Check data length"""
+    assert len(simple_mixture_space.data) == 25
+
+
+def test_joined(joint_design_space):
```
maxhutch, line 73: This only combines two design spaces.  If we want to accept a list, I think we should test that at least 3 work as well.  The third one could be really simple (e.g. just 2 candidates in 1 D).

src/citrine/builders/design_spaces.py
```diff
@@ -0,0 +1,190 @@
...
+                                         description=description,
+                                         descriptors=descriptors,
+                                         data=data)
+    return design_space
+
+
+def enumerate_simple_mixture_cartesian_product(
+    formulation_grids: Mapping[str, Sequence[float]],
+    balance_component: str,
+    descriptors: List[Descriptor],
```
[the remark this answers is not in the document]
Imperssonator: Hmmmmmmmmmmmm............... as much as I think the client desperately needs automated descriptor generation in general, I'm actually not sure this is the place to do it. I think that kind of thing should happen during Table generation. Like "this ingredient amount is a `mass_fraction` therefore it is a RealDescriptor on [0,1]. In general what I do is usually pass Predictor.inputs to my DesignSpace, which helps guarantee that I've mapped my keys correctly.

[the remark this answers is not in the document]
maxhutch: I general I agree, but, given that we're normalizing to zero (via the balance element) we know the lower and upper bounds and the units.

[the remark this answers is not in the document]
Imperssonator: I think this does raise a more general question about the API, which is, "why do I need to pass descriptors into design spaces at all?" IMO, Descriptors should live in one place: **Table headers** (and in some sense they already do, as Bounds!). Every other module should just be passed `keys` (would _vastly_ simplify their instantiation), and then use those keys to pull the descriptor from the Table(s) being used for the general workflow.

src/citrine/builders/design_spaces.py
```diff
@@ -0,0 +1,190 @@
...
+    filtered out of the design space. In other words, formulation_grids could
+    look like:
+
+    {'balance component':[0.8, 0.85, 0.9],
+     'other component':[0.1, 0.2, 0.3, 0.4, 0.5]}
+
+    and the result would be:
+
+    [{'balance component': 0.8, 'other component': 0.2},
+     {'balance component': 0.9, 'other component': 0.1}]
```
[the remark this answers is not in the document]
Imperssonator: If you wanna get really crazy, there's [doctest](https://docs.python.org/3/library/doctest.html), which turns examples in docstrings into tests...

src/citrine/builders/design_spaces.py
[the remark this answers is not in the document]
Imperssonator: I think what I'll do is if they _don't_ pass descriptors, I will auto-generate.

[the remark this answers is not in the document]
maxhutch: What table headers would you pull the bounds from?  A design space that is built on top of a data source should not have to re-define its descriptors; I agree about that.  But there's no table here.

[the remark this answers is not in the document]
Imperssonator: I know there isn't a Table here - the Design Space would have to make that connection through the data source of whatever Predictor it's paired with in whatever Workflow it's in. Ah I see, I think what I'm suggesting goes against the whole "modularity" angle. I'm operating under the assumption that you pretty much make a new design space every time you make a new model.

src/citrine/builders/design_spaces.py
Imperssonator: added

src/citrine/builders/design_spaces.py
```diff
@@ -0,0 +1,190 @@
+import pandas as pd
```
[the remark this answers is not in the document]
Imperssonator: All of the functions use pandas. Added the try/except up top.

Imperssonator:
Review roundup:

Changes to functions:
- changed the formulation function to `enumerate_formulation_grid`
- changed some terms in the docstrings to `grid`
- auto-generating descriptors
- raising ValueErrors rather than assertions
- added check for uniqueness of all descriptors in join function
- raise ImportError for pandas

Changes to tests:
- introduced an `overspec_mixture_space` that should get significantly down-filtered by the narrow range of balance_component
- introduced tests on uniqueness of candidates
- introduced tests on completeness of grid propagation to candidate lists
- introduced test for merging 3 design spaces

src/citrine/builders/design_spaces.py
```diff
@@ -0,0 +1,205 @@
...
+) -> EnumeratedDesignSpace:
+    """[ALPHA] Enumerate a Cartesian product from 1-D grids.
+
+    Given a dict of lists or tuples of values for each design descriptor,
+    create the list of candidates representing the Cartesian product of all
+    values for each descriptor
+
+    Parameters
+    ----------
+    design_grids: Mapping[str, Sequence]
```
maxhutch, line 25: Its just one grid, right?  `design_grid`

src/citrine/builders/design_spaces.py
```diff
@@ -0,0 +1,205 @@
...
+from itertools import product
+from typing import Mapping, Sequence, List
+from citrine.informatics.design_spaces import EnumeratedDesignSpace
+from citrine.informatics.descriptors import Descriptor, RealDescriptor
+
+
+def enumerate_cartesian_product(
+    design_grids: Mapping[str, Sequence],
+    descriptors: List[Descriptor],
+    name: str = 'Enumerated Cartesian product design space',
```
maxhutch, line 14: I do think these static names will increase the rate at which duplicate names are written to the platform (if used).  We might want to balance it being a bit annoying to have to specify a name everywhere against that.

src/citrine/builders/design_spaces.py
```diff
@@ -0,0 +1,205 @@
...
+    data = df_ds.to_dict('records')
+    design_space = EnumeratedDesignSpace(name=name,
+                                         description=description,
+                                         descriptors=descriptors,
+                                         data=data)
+    return design_space
+
+
+def enumerate_formulation_grid(
+    formulation_grids: Mapping[str, Sequence[float]],
```
maxhutch, line 50: Singular, right?

src/citrine/builders/design_spaces.py
```diff
@@ -0,0 +1,205 @@
...
+    balance_component: str,
+    descriptors: List[Descriptor] = None,
+    name: str = 'Enumerated Formulation Grid',
```
maxhutch, line 53: Same as above.

src/citrine/builders/design_spaces.py
```diff
@@ -0,0 +1,205 @@
...
+    data = form_ds.to_dict('records')
+    design_space = EnumeratedDesignSpace(name=name,
+                                         description=description,
+                                         descriptors=descriptors,
+                                         data=data)
+    return design_space
+
+
+def cartesian_join_design_spaces(
+    design_space_list: List[EnumeratedDesignSpace],
```
maxhutch, line 152: `design_spaces`? or `subspaces`?  We probably want to use the same language as https://github.com/CitrineInformatics/ai-engine-documentation/blob/master/docs/design-spaces.md#productdesignspace-modifications-to

src/citrine/builders/design_spaces.py
```diff
@@ -0,0 +1,205 @@
...
+    for ds in design_space_list:
+        if 'join_key' in ds.data[0].keys():
+            raise ValueError('"join_key" cannot be a descriptor key when using this function')
+        all_keys.extend(list(ds.data[0].keys()))
+    if len(all_keys) != len(set(all_keys)):
+        raise ValueError('Duplicate keys are not allowed across design spaces')
+
+    # Convert data fields of EDS into DataFrames to prep for join
+    data_list = [ds.data for ds in design_space_list]
+    ds_list = [pd.DataFrame(data) for data in data_list]
```
maxhutch, line 182: I think these lines could be merged.

src/citrine/builders/design_spaces.py
Imperssonator: Could take on `datetime` as a dependency and generate a YY-MM-DD-hh:mm stamp. I would argue in favor of doing that for all modules. Required naming is generally annoying throughout the API.

src/citrine/builders/design_spaces.py
Imperssonator: oh wow line 181 was quite possibly the laziest thing I have ever done haha

src/citrine/builders/design_spaces.py
Imperssonator: I think originally I was going for the idea that each dimension (each item in the dict) was its own grid, but grid kind of always implies more than one dimension. There's not a great word for a 1-D grid, is there? I agree that singular grid makes more sense.

Imperssonator:
three minor changes:
- list of design spaces argument to the `join` function is now called `subspaces` to align with core design_space module
- `name` arg is required and has no default
- all args are keyword-only

PR #363, opened 2020-06-11
Revert "Revert "PLA-2151 Make bounds on object templates optional""

asantas93:
Reverts CitrineInformatics/citrine-python#362


src/citrine/_utils/functions.py
```diff
@@ -43,7 +43,7 @@ def validate_type(data_dict: dict, type_name: str) -> dict:
 
 def scrub_none(json):
     """
-    Recursively delete dictionary keys and remove list entries with the value ``None``.
+    Recursively delete dictionary keys with the value ``None``.
```
asantas93, line 46: Removing `None` from lists makes tuple deserialization from JSON arrays impossible when optional elements are missing.

PR #365, opened 2020-06-17
Require positional

kroenlein:
# Citrine Python PR

## Description 
This PR migrates the constructor behavior of citrine-python GEMD objects to require name (except IngredientRun) and to accept at most 1 positional argument (name).  This reconciles code behavior with gemd-python 0.10.0 release candidate ([draft PR](https://github.com/CitrineInformatics/gemd-python/pull/111)) and with Best Practice as described in CONTRIBUTING.

This also removes `name` and `labels` from the IngredientRun constructor, finishing the deprecation cycle for Ingredient Runs and Specs having potentially different values for name/labels.

This will likely have substantial downstream effects for repos that were not respecting these rules.

### PR Type:


src/citrine/resources/ingredient_run.py
```diff
@@ -102,8 +94,8 @@ def __init__(self,
                                    material=material, process=process,
                                    mass_fraction=mass_fraction, volume_fraction=volume_fraction,
                                    number_fraction=number_fraction,
-                                   absolute_quantity=absolute_quantity, labels=labels,
```
asantas93, line 105: I don't remember off the top of my head but do we magically populate these fields somehow? It is still a requirement that they are serialized with the current backend implementation of ingredient specs.

asantas93: Oh this is citrine-python (not gemd). I think the answer is yes, then.

asantas93: I'm still not sure it's a good idea to make this change while the backend requires these fields. I think it will require the user to get the ingredient spec before they can save a run, which isn't really necessary when using the API directly. I think we should remove them on the backend first if we want to remove them here so we don't force additional requests which aren't technically necessary.

kroenlein: I agree, part of why this is still a draft.  I can back off the IngredientRun migration if it will get in the way of getting the positional argument PR through, but there's a collision there since we don't want to add `IngredientRun.name` as a required argument.

asantas93: You don't want to have `name` after the `*`?

kroenlein: I'm looking at consequences, but removing all visibility on setting the name and labels from a user perspective seems like the consistent way to keep people from setting something that they don't really set.  [PLA-2437](https://citrine.atlassian.net/browse/PLA-2437)

kroenlein:
Needs to follow https://github.com/CitrineInformatics/gemd-python/pull/111

src/citrine/resources/ingredient_run.py
kroenlein: It's worth noting that setting names on IngredientRuns has been issuing deprecation warnings for months.

asantas93: Removing the ability to specify them in the constructor seems like a good thing to do, but it would be nice to still be able to see them upon deserialization from the backend. Maybe we can add these in a way similar to `client_specific_fields`? I.e. some additional property key collection called `read_only_fields` or something.

kroenlein: Right now, gemd-python handles this by explicitly stripping them from the dictionary during deserialization 

https://github.com/CitrineInformatics/gemd-python/pull/111/files#diff-85eb4121e9b5a966e163a4995a0c7bf3

I _think_ the citrine-python object should be able to just ride on that functionality.  I'll make sure, however.

kroenlein:
Rebuild me!

PR #367, opened 2020-06-18
Pla 4498/search projects

zoesyang:
# Citrine Python PR

## Description 
Modifies the collection and ProjectCollections classes to allow for project search. Ultimately, the goal is to leverage the project search feature in the e2e's.

### PR Type:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,7 @@ class PreBuildCommand(build_py):
 
 
 setup(name='citrine',
-      version='0.36.3',
+      version='0.36.4',
       url='http://github.com/CitrineInformatics/citrine-python',
       description='Python library for the Citrine Platform',
       author='Citrine Informatics',
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,7 @@ class PreBuildCommand(build_py):
 
 
 setup(name='citrine',
-      version='0.37.1',
+      version='0.38.0',
       url='http://github.com/CitrineInformatics/citrine-python',
       description='Python library for the Citrine Platform',
       author='Citrine Informatics',
--- a/src/citrine/_rest/collection.py
+++ b/src/citrine/_rest/collection.py
@@ -1,5 +1,5 @@
 from abc import abstractmethod
-from typing import Optional, Union, Generic, TypeVar, Iterable, Dict, Tuple
+from typing import Optional, Union, Generic, TypeVar, Iterable, Dict, Tuple, Callable
 from uuid import UUID
 
 from citrine._rest.paginator import Paginator
@@ -53,6 +53,7 @@ class Collection(Generic[ResourceType]):
             return self.build(data)
         except NonRetryableException as e:
             raise ModuleRegistrationFailedException(model.__class__.__name__, e)
+    
 
     def list(self,
              page: Optional[int] = None,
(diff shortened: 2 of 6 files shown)


src/citrine/_rest/collection.py
```diff
@@ -79,11 +80,56 @@ def list(self,
...
 
         """
-        return self._paginator.paginate(self._fetch_page,
+        return self._paginator.paginate(self._fetch_page_list,
                                         self._build_collection_elements,
                                         page,
                                         per_page)
 
+
+    def search(self, search_params: Optional[dict] = None,
```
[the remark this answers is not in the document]
zoesyang: I can move the search functionality into the ProjectCollection class. Wasn't sure whether or not it was okay to call the private fetch page function in the ProjectCollection class, but will change.

src/citrine/_rest/collection.py
```diff
@@ -79,11 +80,56 @@ def list(self,
...
                                         self._build_collection_elements,
                                         page,
                                         per_page)
 
+
+    def search(self, search_params: Optional[dict] = None, 
+                    page: Optional[int] = None, per_page: int = 100) -> Iterable[ResourceType]:
+        """
+            Search for elements matching search_params in the collection, and paginate over
+            the results. This differs from the list function, because it makes a POST request
```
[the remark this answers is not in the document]
zoesyang: In order to be able to search by different features by exact or substring match, it was easier to implement this as a POST request rather than a GET request with many query parameters

src/citrine/resources/project.py
```diff
@@ -365,3 +365,60 @@ def register(self, name: str, description: Optional[str] = None) -> Project:
...
+                "search_params": {
+                    "name": {
+                        "value": "Polymer",
+                        "search_method": "SUBSTRING"
+                    },
+                    "description": {
+                        "value": "polymer chain length",
+                        "search_method": "SUBSTRING"
+                    },
+                    "status": {
```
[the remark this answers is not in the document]
jfriend-citrine: Def looks like that field is a vestige or intended for soft delete behavior at some point.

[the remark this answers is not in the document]
maxhutch: Then shouldn't we remove it?

setup.py
```diff
@@ -57,7 +57,7 @@ def run(self):
 
 
 setup(name='citrine',
-      version='0.37.0',
+      version='0.37.1',
```
maxhutch, line 60: New functionality should be a minor version bump.

src/citrine/resources/project.py
```diff
@@ -365,3 +365,117 @@ def register(self, name: str, description: Optional[str] = None) -> Project:
...
+            The dict can constain any combination of (one or all) search specifications
+            for the name, description, and status fields of a project. For each parameter
+            specified, the "value" to match, as well as the "search_method" must be provided.
+            The available search_methods are "SUBSTRING" and "EXACT". The example above demonstrates
+            the input necessary to list projects with names including the word "Polymer,"
+            descriptions containing the phrase "polymer chain length," and statuses exactly
+            equal to "not-started."
+        page: int, optional
+            The "page" of results to list. Default is to read all pages and yield
+            all results.  This option is deprecated.
```
maxhutch, line 416: This is a new function, so we don't have to include `page` and `per_page` if they are deprecated.

src/citrine/resources/project.py
```diff
@@ -365,3 +365,117 @@ def register(self, name: str, description: Optional[str] = None) -> Project:
...
+        per_page: int, optional
```
maxhutch, line 417: Is this also deprecated?  If so, we can just leave it out.

maxhutch:
@zoesyang have you tried this with the e2e tests against development or devkit?  I don't especially trust our mocks, so it would be nice to know that this won't break the e2e when merged.

src/citrine/resources/project.py
zoesyang: Oh good point! I'll leave it out!

zoesyang: @maxhutch I'm having some issues removing the "page" parameter, because the pagination mechanism in the Pagination class assumes that there is a "page" parameter. Can we leave it in?

maxhutch: Can't you just pass in `page = None`?

maxhutch:
> @zoesyang have you tried this with the e2e tests against development or devkit? I don't especially trust our mocks, so it would be nice to know that this won't break the e2e when merged.

I ran this branch against development and didn't get any list-related failures (I got a few timeouts, obviously).  So :+1: on that front.

src/citrine/resources/project.py
zoesyang: The issue is that then `[this line](https://github.com/CitrineInformatics/citrine-python/blob/1b75914484b4a9138ef7ec7e5b18eced57da8375/src/citrine/_rest/paginator.py#L57)` gets passed "page", but page_fetcher does not have a "page" argument for search if we remove the argument. I tried passing page_kwarg = { 'page': page } if page is Not None else {} like this:  page_fetcher(**page_kwarg, per_page=per_page), but then tests that were expecting to get a "page" argument all fail

src/citrine/resources/project.py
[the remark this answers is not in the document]
jfriend-citrine: I will add a tech-debt ticket to eradicate the field after more analysis as to why it exists.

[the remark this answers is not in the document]
jfriend-citrine: https://citrine.atlassian.net/browse/PLA-4504

PR #368, opened 2020-06-18
Feature/pla 3799 list display tables for ara defn

dleblanc:
PLA 3799: adds a 'list for table config' listing to the tables collection.

This is safe to merge even if the backend code isn't merged yet, since we're only using unit tests here.

### PR Type:


dleblanc:
Rebuild!

src/citrine/resources/table.py
```diff
@@ -111,6 +111,35 @@ def build_versions(collection: Iterable[dict]) -> Iterable[Table]:
 
         return self._paginator.paginate(fetch_versions, build_versions, page, per_page)
 
+    def list_for_ara_definition(self,
```
ventura-rivera, line 114: Can we change to `list_by_ara_definition`?

tests/resources/test_table.py
```diff
@@ -98,6 +98,21 @@ def test_list_table_versions(collection, session):
     assert results[0].uid is not None
 
 
+def test_list_for_ara_definition(collection, session):
```
ventura-rivera, line 101: This would also need to change to `test_list_by_ara_definition`

src/citrine/resources/table.py
asantas93: I think we want this to be `list_by_config` if we're moving forward with the new terminology

ventura-rivera: I agree with this

djack:
@dleblanc @ventura-rivera bump - I'd like to get this merged and tested.

src/citrine/resources/table.py
dleblanc: Good call. Fixed.

tests/resources/test_table.py
dleblanc: Also fixed.

src/citrine/resources/table.py
```diff
@@ -111,6 +111,38 @@ def build_versions(collection: Iterable[dict]) -> Iterable[Table]:
 
         return self._paginator.paginate(fetch_versions, build_versions, page, per_page)
 
+    def list_by_config(self,
+                                ara_definition_uid: UUID,
```
asantas93, line 115: formatting looks off

dleblanc: Yeah, Intellij. Reformatted by hand.

PR #370, opened 2020-06-19
Fix ProjectCollection.search; use in find_collection

maxhutch:
# Citrine Python PR

## Description 

This adds a `get_resource_post` method to the session that behaves
like get_resource (in that it extracts the response json) but
POSTs with a body instead of GET.  That method is then
used in ProjectCollection.search.

### PR Type:

- [] I have written Numpy-style docstrings for every method and class.

### DIFF
--- a/src/citrine/_session.py
+++ b/src/citrine/_session.py
@@ -174,14 +174,6 @@ class Session(requests.Session):
         response = self.checked_get(path, **kwargs)
         return self._extract_response_json(path, response)
 
-    def get_resource_post(self, path: str, json: dict, **kwargs) -> dict:
-        """POST with a body to get a particular resource as JSON.
-
-        This is used when performing queries, e.g. projects.search.
-        """
-        response = self.checked_post(path, json=json, **kwargs)
-        return self._extract_response_json(path, response)
-
     def post_resource(self, path: str, json: dict, **kwargs) -> dict:
         """POST to a particular resource as JSON."""
         response = self.checked_post(path, json=json, **kwargs)
--- a/src/citrine/_session.py
+++ b/src/citrine/_session.py
@@ -174,6 +174,14 @@ class Session(requests.Session):
         response = self.checked_get(path, **kwargs)
         return self._extract_response_json(path, response)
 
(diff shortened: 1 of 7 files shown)


maxhutch:
I have used this to run the e2e tests (which is what surfaced the bug).  Prior to using search in `find_collection`, the search method was unused in the e2e tests.

src/citrine/_session.py
```diff
@@ -174,6 +174,14 @@ def get_resource(self, path: str, **kwargs) -> dict:
         response = self.checked_get(path, **kwargs)
         return self._extract_response_json(path, response)
 
+    def get_resource_post(self, path: str, json: dict, **kwargs) -> dict:
```
sparadiso, line 177: What is the difference between this and `post_resource`? Is this just for a more intuitive method name to use when resource retrieval is a POST?

tests/seeding/test_find_or_create.py
```diff
@@ -63,6 +63,25 @@ def list(self, page: Optional[int] = None, per_page: int = 100):
             else:
                 return self.projects[(page - 1)*per_page:page*per_page]
 
+        def search(self, search_params: Optional[dict] = None, per_page: int = 100):
```
sparadiso, line 66: unused?

maxhutch: find_collection calls `search` now, so this is used (and w/o it the tests fail).

src/citrine/_session.py
maxhutch: I just hadn't noticed that the content was the same as `post_resource`.  While there may be value in having different names, I think simplicity wins here.  I removed `get_resource_post` and am calling `post_resource` in search now.

src/citrine/resources/project.py
```diff
@@ -464,6 +464,6 @@ def _fetch_page_search(self, page: Optional[int] = None,
 
         path = self._get_path() + "/search"
 
-        return self._fetch_page(path=path, fetch_func=self.session.checked_post,
+        return self._fetch_page(path=path, fetch_func=self.session.post_resource,
```
zoesyang, line 467: Oh sweet!

Issue #166, opened 2020-01-09
Serialization properties cannot be defined in parent classes


maxhutch:
Many of our `PolymorphicSerializable` class hierarchies have shared members.  It would be nice to bring these members (e.g. `key` [descriptors](https://github.com/CitrineInformatics/citrine-python/blob/develop/src/citrine/informatics/descriptors.py)) into the parent class to communicate that they are part of the interface and to deduplicate their definition.  However, the magik in those properties causes an error to be thrown when this is attempted, e.g.
```
E       assert <citrine.informatics.descriptors.InorganicDescriptor object at 0x7fe6b29af630> == <citrine.informatics.descriptors.InorganicDescriptor object at 0x7fe6945453c8>
```

Interesting, I get a completely different error when I try to do this in #163 : 
```
tests/ara/test_variables.py:9: in <module>
    RootInfo("root name", ["Root", "Name"], "name"),
src/citrine/ara/variables.py:71: in __init__
    self.short_name = short_name
src/citrine/_serialization/properties.py:114: in __set__
    getattr(base_class, self.serialization_path).fset(obj, value_to_set)
E   AttributeError: 'NoneType' object has no attribute 'fset'
```

bfolie:
Resolved by #453

PR #375, opened 2020-06-26
Add experimental and experimental_reasons fields

maxhutch:
# Citrine Python PR

## Description 
These are present on all modules and workflows as type boolean
and list[str].  They are response-only fields populated
by the service.  Because they aren't part of the constructor, they aren't particularly visible
in the autodocs, so I went ahead and added them to the sphinx docs.

### PR Type:

### DIFF
--- a/src/citrine/informatics/design_spaces.py
+++ b/src/citrine/informatics/design_spaces.py
@@ -60,6 +60,12 @@ class ProductDesignSpace(Resource['ProductDesignSpace'], DesignSpace):
         serializable=False
     )
     active = properties.Boolean('active', default=True)
+    experimental = properties.Boolean("experimental", serializable=False)
+    experimental_reasons = properties.Optional(
+        properties.List(properties.String()),
+        'experimental_reasons',
+        serializable=False
+    )
 
     # NOTE: These could go here or in _post_dump - it's unclear which is better right now
     module_type = properties.String('module_type', default='DESIGN_SPACE')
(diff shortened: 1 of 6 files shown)


maxhutch:
I've been able to run this version locally against development.

src/citrine/informatics/design_spaces.py
```diff
@@ -60,6 +60,12 @@ class ProductDesignSpace(Resource['ProductDesignSpace'], DesignSpace):
         serializable=False
     )
     active = properties.Boolean('active', default=True)
+    experimental = properties.Boolean("experimental", serializable=False, default=True)
```
gregor-robinson, line 63: If `experimental_reasons` is to be populated when `experimental == True`, then it seems the default should be `False`. I imagine, however, that this default is proposed as a conservative choice during the transition to experimental annotations (and implicitly marking mature features as such by overriding this default).

maxhutch: Yeah, this is a very short term conservative default.  As soon as the backend change goes out to all deployments, we won't be triggering this default anymore and I'd be inclined to remove it.

maxhutch:
> this seems to be a lot of repetition, so perhaps this could be accomplished more tidily with a mix-in?

Unfortunately, these property definitions cannot easily be migrated into parent classes. I've tried to get that working on several occasions to no avail.  https://github.com/CitrineInformatics/citrine-python/issues/166

src/citrine/informatics/design_spaces.py
andyczerwonka: >  As soon as the backend change goes out to all deployments, we won't be triggering this default anymore and I'd be inclined to remove it.

Yep. makes sense.

PR #377, opened 2020-06-30
Add pre-register/update validation of enumerated design space size

maxhutch:
# Citrine Python PR

## Description 

EnumeratedDesignSpaces can be pretty big, so we want to return a helpful
error message rather than let the POST or PUT call fail because the
request body is too big.

### PR Type:

### DIFF
--- a/tests/resources/test_design_space.py
+++ b/tests/resources/test_design_space.py
@@ -1,6 +1,12 @@
 import uuid
+from random import random
 
+import pytest
+
+from citrine.informatics.descriptors import RealDescriptor
+from citrine.informatics.design_spaces import EnumeratedDesignSpace
 from citrine.resources.design_space import DesignSpaceCollection
+from tests.utils.session import FakeSession
 
 
 def test_design_space_build():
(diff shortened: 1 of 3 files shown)


tests/resources/test_design_space.py
```diff
@@ -39,3 +45,49 @@ def test_design_space_build():
...
+
+    # These still throw exceptions because we haven't setup the mock response
+    # But we can check the message to make sure it is NOT the exception from above
+    with pytest.raises(ValueError) as excinfo:
+        collection.register(just_right)
+    assert "only supports" not in str(excinfo.value)
+
+    with pytest.raises(ValueError) as excinfo:
+        collection.update(just_right)
+    assert "only supports" not in str(excinfo.value)
```
bfolie, line 93: Setting up the mock response is pretty simple, so I think we should do it before merging this PR

maxhutch:
> I would prefer not to be doing validation of design space size client-side. If we want to implement these limits asap then I agree this is the quickest way to do it, but I'd like to implement them in Orion and then remove this soon.
> 
> It does make sense to me that the client would check that the request body does not exceed size limits, but those are looser. What are the size limits on a request body, anyway? You said you saw errors around 15k, but presumably that translates into some number of bytes. What piece of software enforces that limit?

Wile we could impose a loose limit here and then a tighter limit during validation, that might be confusing.  Given that we need some limit here to avoid triggering server-side 500s on huge requests, I'd prefer to start with the actual limit and then relax it if we have a hard time keeping these in sync.  We should add the backend limit as well, ofc. I added https://citrine.atlassian.net/browse/PLA-4371 to capture that.

PR #381, opened 2020-07-02
Added fallback to octet-stream

kroenlein:
# Citrine Python PR

## Description 
This PR modifies the MIME type resolution so that any unrecognized upload is uploaded with MIME type `application/octet-stream` as per [common practice](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types).  

Issue was encountered trying to upload a file with an `.xlsx` file extension.  If we instead want to include the [formally correct MIME type](https://www.iana.org/assignments/media-types/media-types.xhtml), we should define a Citrine custom MIME type file, as per [mimetypes](https://docs.python.org/3/library/mimetypes.html#mimetypes.init).  This seems excessive as we will not be acting on them.

https://citrine.atlassian.net/browse/PLA-4395

### PR Type:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,7 @@ class PreBuildCommand(build_py):
 
 
 setup(name='citrine',
-      version='0.40.5',
+      version='0.40.6',
       url='http://github.com/CitrineInformatics/citrine-python',
       description='Python library for the Citrine Platform',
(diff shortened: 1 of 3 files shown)


kroenlein:
Rebuild me!

PR #382, opened 2020-07-02
modify file_link to use the mime type library's guess_type method 

djack:
First, use mimetypes' [guess_type](https://docs.python.org/2/library/mimetypes.html#mimetypes.guess_type) method, which includes commonly used but non-iana mime types (like xslx) before falling back to `application/octet-stream`

# Citrine Python PR

## Description 

### PR Type:

### DIFF
--- a/src/citrine/resources/file_link.py
+++ b/src/citrine/resources/file_link.py
@@ -196,6 +196,8 @@ class FileCollection(Collection[FileLink]):
     def _make_upload_request(self, file_path: str, dest_name: str):
         """
         Make a request to the backend to upload a file.
+        Uses mimetypes.guess_type to guess the mime type of the file, including 
+        commonly used but not IANA approved mime types (like .xlsx).
 
         Parameters
         ----------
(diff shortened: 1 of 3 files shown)


src/citrine/resources/file_link.py
```diff
@@ -127,7 +127,7 @@ def _as_dict_from_resource(self, file: dict):
         This is necessary because the database resource contains additional information that is
         not in the FileLink object, such as file size and the id of the user who uploaded the file.
 
-        Paramters
+        Parameters
```
kroenlein, line 130: Good catch

tests/resources/test_file_link.py
```diff
@@ -30,6 +30,18 @@ def valid_data() -> dict:
...
+    expected_xls = "application/vnd.ms-excel"
+    expected_txt = "text/plain"
+    expected_unk = "application/octet-stream"
+
+    assert collection._mime_type("asdf.xlsx") == expected_xlsx
+    assert collection._mime_type("asdf.XLSX") == expected_xlsx
+    assert collection._mime_type("asdf.xls") == expected_xls
+    assert collection._mime_type("asdf.TXT") == expected_txt
+    assert collection._mime_type("asdf.FAKE") == expected_unk
+
```
andyczerwonka, line 44: we should add a CSV test

PR #387, opened 2020-07-08
Add back design space size limit

maxhutch:
# Citrine Python PR

## Description 
The [first attempt](https://github.com/CitrineInformatics/citrine-python/pull/377) ran into issues with our existing e2e tests.  This increases the limit on the number of columns significantly (which has been checked against the post body limit) and the e2e tests are being thinned a bit.

### PR Type:
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

### Adherence to team decisions
- [x] I have added tests for 100% coverage
- [x] I have written Numpy-style docstrings for every method and class.
- [x] I have communicated the downstream consequences of the PR to others.
- [ ] I have bumped the version in setup.py

PR #388, opened 2020-07-08
Warn on experimental register and update

maxhutch:
# Citrine Python PR

## Description 
Add a warning when a register or update call returns an experimental resource.  This is part of PLA-4581.

### PR Type:

### DIFF
--- a/src/citrine/_rest/collection.py
+++ b/src/citrine/_rest/collection.py
@@ -210,7 +210,10 @@ class Collection(Generic[ResourceType]):
 
     def _check_experimental(self, data):
         if data.get("experimental", False):
-            msg = "The resource is experimental because: \n  {}".format(
+            uid = data.get("id")
+            typ = self._resource().__class__.__name__
+            msg = "The {} with id {} is experimental because: \n  {}".format(
+                typ, uid,
                 "\n  ".join(data.get("experimental_reasons") or ["Unknown reason"])
             )
             warnings.warn(msg)
--- a/src/citrine/_rest/collection.py
+++ b/src/citrine/_rest/collection.py
@@ -1,3 +1,4 @@
+import warnings
 from abc import abstractmethod
 from typing import Optional, Union, Generic, TypeVar, Iterable, Dict, Tuple, Callable
 from uuid import UUID
@@ -50,6 +51,7 @@ class Collection(Generic[ResourceType]):
         try:
             data = self.session.post_resource(path, model.dump())
             data = data[self._individual_key] if self._individual_key else data
+            self._check_experimental(data)
             return self.build(data)
         except NonRetryableException as e:
             raise ModuleRegistrationFailedException(model.__class__.__name__, e)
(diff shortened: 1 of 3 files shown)


maxhutch:
Some of https://github.com/CitrineInformatics/citrine-python/pull/387 bled into this because of poor hygiene on my end.  It should come after anyways.

src/citrine/_rest/collection.py
```diff
@@ -204,3 +207,10 @@ def _page_params(self,
         if module_type is not None:
             params["module_type"] = module_type
         return params
+
+    def _check_experimental(self, data):
+        if data.get("experimental", False):
+            msg = "The resource is experimental because: \n  {}".format(
+                "\n  ".join(data.get("experimental_reasons") or ["Unknown reason"])
+            )
+            warnings.warn(msg)
```
latture, line 216: We may want to consider always raising this warning. I can imagine this text might be the same for multiple modules, and we'd want to warn each time.

tests/resources/test_predictor.py
```diff
@@ -48,6 +48,20 @@ def test_register(valid_simple_ml_predictor_data, basic_predictor_report_data):
...
+    post_response = deepcopy(valid_simple_ml_predictor_data)
+    post_response["experimental"] = True
+    post_response["experimental_reasons"] = ["This is a test", "Of experimental reasons"]
+    session.post_resource.return_value = post_response
+    session.get_resource.return_value = basic_predictor_report_data
+    pc = PredictorCollection(uuid.uuid4(), session)
+    predictor = SimpleMLPredictor.build(valid_simple_ml_predictor_data)
+    with pytest.warns(UserWarning) as record:
+        pc.register(predictor)
+    assert len(str(record[0].message).split("\n")) == 3
```
latture, line 62: This is a rather opaque assert. Can we either add a comment about why this assertion exists or assert something about the contents of the warning?

src/citrine/_rest/collection.py
maxhutch: What do you mean `always`?

latture: From the [documentation](https://docs.python.org/3/library/warnings.html#module-warnings):

> The disposition of warnings can vary based on the warning category, the text of the warning message, and the source location where it is issued. Repetitions of a particular warning for the same source location are typically suppressed.

which basically says that subsequent warnings from this location with the same message will be suppressed. We could add some additional data to the text so a warning is only issued once per experimental resource, e.g. add the id to the message, or always raise the warning using a filter: `warnings.simplefilter('always', UserWarning)`. Ideally we'd use a custom warning class besides `UserWarning` so we only raise repeated instances of our warnings and not any `UserWarning` issued anywhere in our program.

maxhutch: Excellent point.  Here's what I get now:
```
  /home/maxhutch/src/citrine-python/src/citrine/_rest/collection.py:219: UserWarning: The Predictor with id c22ece8b-d5b8-4d17-976a-f6798b22c048 is experimental because: 
    Graph predictor contains more than 1 SimpleMLPredictors
```

PR #396, opened 2020-07-12
Change the design space size check to count cells

maxhutch:
## Description 

The client-side design space size limit is meant simply to catch
very large POSTs and PUTs that would cause a 5xx on the backend
because they are too big.  This changes "too big" from either
"too many rows" or "too many columns" to "too many cells", with
the actual number of cells unchanged.

NOTE this does not imply that we can actually handle any
EnumeratedDesignSpace that fits within this limit; simply that it
shoudln't cause a 5xx in the handler on the backend.

### PR Type:

### DIFF
--- a/tests/resources/test_design_space.py
+++ b/tests/resources/test_design_space.py
@@ -53,24 +53,17 @@ def test_design_space_limits():
     session = FakeSession()
     collection = DesignSpaceCollection(uuid.uuid4(), session)
 
-    too_long = EnumeratedDesignSpace(
+    too_big = EnumeratedDesignSpace(
         "foo",
         "bar",
-        descriptors=[RealDescriptor("x", 0, 1)],
-        data=[{"x": random()} for _ in range(2001)]
-    )
-
-    too_wide = EnumeratedDesignSpace(
-        "foo",
-        "bar",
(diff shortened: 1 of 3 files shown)


maxhutch:
I've tested this against a local e2e test update that tests {real values, categorical values with 20 char names} and {128x2000, 1x256000} and everything POSTs fine.

tests/resources/test_design_space.py
```diff
@@ -53,24 +53,17 @@ def test_design_space_limits():
...
+        descriptors=[RealDescriptor("R-{}".format(i), 0, 1) for i in range(128)],
+        data=[{"R-{}".format(i): random() for i in range(128)} for _ in range(2001)]
     )
 
     just_right = EnumeratedDesignSpace(
         "foo",
         "bar",
-        descriptors=[RealDescriptor("R-{}".format(i), 0, 1) for i in range(10)],
+        descriptors=[RealDescriptor("R-{}".format(i), 0, 1) for i in range(128)],
         data=[{"R-{}".format(i): random() for i in range(128)} for _ in range(1000)]
```
bfolie, line 67:
```suggestion
        data=[{"R-{}".format(i): random() for i in range(128)} for _ in range(2000)]
```

PR #397, opened 2020-07-13
Pla 4407/pla 4673/resource transfer

zoesyang:
# Citrine Python PR

## Description 
Adds a method to the projects class for transferring all resources from one project to another. This will ultimately serve to allow users to transfer resources out of a project in order to later delete that project. See [https://citrine.atlassian.net/browse/PLA-4407](url). Please note that the "impact assessment"/ dry-run capabilities of this feature have not been fully scoped, and therefore cannot yet be accessed by citrine-python.

### PR Type:


src/citrine/resources/project.py
```diff
@@ -210,6 +210,26 @@ def share(self,
             "resource": {"type": resource_type, "id": resource_id}
         })
 
+    def transfer_resources(self, receiving_project_id: str) -> bool:
+        """
+        Transfers all resources from this project to the project with id = receiving_project_id.
```
maxhutch, line 215: What permissions are required?  Are all resources supported?  Is the transaction atomic (i.e. is there any way that some of the resources would be transferred but others would fail to transfer?)

src/citrine/resources/project.py
```diff
@@ -221,6 +221,34 @@ def share(self,
             "resource": {"type": resource_type, "id": resource_id}
         })
 
+    def transfer_resource(self, receiving_project_uid: Union[str, UUID],
```
maxhutch, line 224: This is very minor and optional, but I might suggest swapping the order of these two arguments.  "Transfer {resource} to {recieving_project_uid}"

src/citrine/resources/project.py
```diff
@@ -221,6 +221,34 @@ def share(self,
...
+                          resource: Resource) -> bool:
+        """
+        Transfer ownership of a resource.
+
+        The new owner of the the supplied resource becomes the project
+        with uid = receiving_project_uid.
```
maxhutch, line 230: What access does the user need in order to make this call?  E.g. "You must be a lead in the originating project and have write access to the receiving project."

src/citrine/resources/project.py
zoesyang: I agree! that's more intuitive

src/citrine/resources/project.py
zoesyang: At the moment, you must be a lead of both projects, but we are looking into allowing the lead of one project transfer a resource to a project in which the user is a member.

PR #400, opened 2020-07-16
fix usage of project search in find_collection function

zoesyang:
# Citrine Python PR

## Description 
Updates the find_collection function to correctly use search parameters. This should help speed up tests.

### PR Type:

### DIFF
--- a/src/citrine/resources/project.py
+++ b/src/citrine/resources/project.py
@@ -402,7 +402,7 @@ class ProjectCollection(Collection[Project]):
                 },
             }
             The dict can constain any combination of (one or all) search specifications for the
-        name, description, and status fields of a project. For each parameter specified, the
+            name, description, and status fields of a project. For each parameter specified, the
             "value" to match, as well as the "search_method" must be provided. The available
             search_methods are "SUBSTRING" and "EXACT". The example above demonstrates the input
             necessary to list projects with the exact name "Polymer Project," and descriptions
--- a/src/citrine/resources/project.py
+++ b/src/citrine/resources/project.py
@@ -392,19 +392,17 @@ class ProjectCollection(Collection[Project]):
             A dict representing the body of the post request that will be sent to the search
             endpoint to filter the results ie.
             {
-                "search_params": {
-                    "name": {
-                        "value": "Polymer Project",
-                        "search_method": "EXACT"
-                    },
-                    "description": {
-                        "value": "polymer chain length",
-                        "search_method": "SUBSTRING"
-                    },
-                }
+                "name": {
+                    "value": "Polymer Project",
+                    "search_method": "EXACT"
+                },
+                "description": {
+                    "value": "polymer chain length",
+                    "search_method": "SUBSTRING"
+                },
             }
             The dict can constain any combination of (one or all) search specifications for the
-            name, description, and status fields of a project. For each parameter specified, the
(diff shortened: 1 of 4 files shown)


src/citrine/seeding/find_or_create.py
```diff
@@ -15,9 +15,11 @@ def find_collection(collection, name):
             # call list() to collapse the iterator, otherwise the NotFound
             # won't show up until collection_list is used
             collection_list = list(collection.search(search_params={
-                "name": {
-                    "value": name,
-                    "search_method": "EXACT"
+                "search_params": {
```
maxhutch, line 18: I think nesting a "search_params" key inside an argument called "search_params" is very confusing UX.  Could this be fixed by adjusting the logic in the search method [here](https://github.com/CitrineInformatics/citrine-python/blob/master/src/citrine/resources/project.py#L469) ?

zoesyang: In retrospect, I probably should have changed the ux for that - i was keeping it the same as what the product api request expects. I'll change it

zoesyang:
@maxhutch I changed the search_params to not include 'search_params' as a key. test_experimental.py::test_not_experimental e2e for orion should trigger the search

src/citrine/resources/project.py
```diff
@@ -392,19 +392,17 @@ def search(self, search_params: Optional[dict] = None,
...
+                    "search_method": "EXACT"
+                },
+                "description": {
+                    "value": "polymer chain length",
+                    "search_method": "SUBSTRING"
+                },
             }
             The dict can constain any combination of (one or all) search specifications for the
-            name, description, and status fields of a project. For each parameter specified, the
+        name, description, and status fields of a project. For each parameter specified, the
```
maxhutch, line 405:
```suggestion
            name, description, and status fields of a project. For each parameter specified, the
```

maxhutch: This change looks accidental.

zoesyang: It is! just pushed change

zoesyang:
> The client behavior looks great. Thanks!
> 
> > test_experimental.py::test_not_experimental e2e for orion should trigger the search
> 
> It does hit this route, but it doesn't ever test that the search only returns matching records. I'd guess that's how this bug was caught: there were calls that we thought would be lightweight (because they only returned a few matching records) that were actually returning everything that list would.
> 
> Could we add an e2e that asserts that only matching results are returned?

No problem! I can add an e2e for both the search functionality and the find_collection function in devkit once this is merged :)

zoesyang:
> I verified the basic change locally in lieu of an existing e2e:
> 
> ```
> In [3]: res = list(client.projects.search({"name": {"value": "[PRELOAD", "search_method": "SUBSTRING"}})
>    ...: )                                                                                               
> 
> In [4]: len(res)                                                                                        
> Out[4]: 2
> 
> In [5]: all = list(client.projects.search())                                                            
> 
> In [6]: len(all)                                                                                        
> Out[6]: 289
> ```

Do you still think it would be a good idea to have an e2e?

maxhutch:
> Do you still think it would be a good idea to have an e2e? Also thank you for verifying!

Yeah, especially because it should be fast.  I'm not sure what the backend test coverage of the different search options is, but the different fields and search methods should be checked _somewhere_.

PR #401, opened 2020-07-16
Version requirements bump

kroenlein:
# Citrine Python PR

## Description 
This PR increases the required version of gemd-python from 0.9 to 0.10 to reflect the latest GEMD release.  This has downstream implications for consumers of the library but is transparent from citrine-python's perspective.

### PR Type:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,7 @@ class PreBuildCommand(build_py):
 
 
 setup(name='citrine',
-      version='0.44.1',
+      version='0.45.0',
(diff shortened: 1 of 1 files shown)


kroenlein:
> So this means users importing bounds and such with this new version of citrine python will no longer be able to use positional arguments, right? Approve assuming "I have communicated..." means people know this.

This isn't that PR.  That breaking change is still upstream and I'll be careful about that one.  The only breaking change in this PR should be if people are explicitly invoking the `flatten` method.

PR #405, opened 2020-07-20
[PLA-4021] Forked expression predictor

latture:
This PR renames the current `ExpressionPredictor` to `DeprecatedExpressionPredictor` and defines a new `ExpressionPredictor` with an updated API. The new expression predictor requires aliases for all unknown expression arguments, and keys in `aliases` map to descriptors instead of descriptor keys.

I updated all docs that referenced the previous expression predictor and included details on how to migrate from the deprecated predictor the deprecated class's docstring.

### PR Type:

### DIFF
--- a/src/citrine/informatics/predictors.py
+++ b/src/citrine/informatics/predictors.py
@@ -2,7 +2,6 @@
 # flake8: noqa
 from typing import List, Optional, Type, Union, Mapping
 from uuid import UUID
-import warnings
 
 from citrine._serialization import properties as _properties
 from citrine._serialization.serializable import Serializable
@@ -338,10 +337,6 @@ class DeprecatedExpressionPredictor(Serializable['DeprecatedExpressionPredictor'
                  session: Optional[Session] = None,
                  report: Optional[Report] = None,
                  active: bool = True):
-        warnings.warn(
-            f"{self.__class__.__name__} is deprecated. Please use {ExpressionPredictor.__name__} instead.",
-            DeprecationWarning
-        )
         self.name: str = name
         self.description: str = description
         self.expression: str = expression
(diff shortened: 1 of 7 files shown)


src/citrine/informatics/predictors.py
```diff
@@ -226,8 +228,63 @@ def __str__(self):
         return '<GraphPredictor {!r}>'.format(self.name)
 
 
-class ExpressionPredictor(Serializable['ExpressionPredictor'], Predictor):
-    """[ALPHA] A predictor interface that allows calculator expressions.
+class DeprecatedExpressionPredictor(Serializable['DeprecatedExpressionPredictor'], Predictor):
```
maxhutch, line 231: Can we add a deprecated annotation here (we use https://pypi.org/project/Deprecated/ )

latture: Not to an entire class. (I tried.)

latture: We could deprecate the init, but the deprecation message isn't helpful. I figured if a user decided to use `DeprecatedExpressionPredictor` the deprecation was implied.

latture: We could rename the class and have a construction function that's deprecated 🤷

latture: I just tried using a function but that causes issues with our serialization logic.

bfolie: When I had to do this for `InorganicDescriptor`, I raised a warning: https://github.com/CitrineInformatics/citrine-python/blob/master/src/citrine/informatics/descriptors.py#L116

maxhutch: I remembered that it only worked for one of {functions, classes}, but I thought it was classes.  Sorry.  Explicitly raising the warning like in the example @bfolie referenced should do it.

PR #406, opened 2020-07-20
PLA-4435 add archived flag to predictors

talltea:
# Citrine Python PR

## Description 
Adds the `archived` flag to the client. There's a parallel PR to add this field to the backend.

### PR Type:


src/citrine/informatics/design_spaces.py
```diff
@@ -60,6 +60,7 @@ class ProductDesignSpace(Resource['ProductDesignSpace'], DesignSpace):
         serializable=False
     )
     active = properties.Boolean('active', default=True)
+    archived = properties.Boolean('archived', default=True)
```
bfolie, line 63: Why is the default True for design spaces?

src/citrine/informatics/predictors.py
```diff
@@ -218,6 +224,7 @@ def stuff_predictor_into_envelope(predictor: dict) -> dict:
         return dict(
             module_type='PREDICTOR',
             config=predictor,
+            # TODO: What's this for?
```
bfolie, line 227: Is this question directed at the `stuff_predictor_into_envelope` method?

src/citrine/informatics/design_spaces.py
talltea: woops, that's copy pasta

src/citrine/informatics/predictors.py
talltea: Yes, but mostly I was confused why the active flag was set to False here

bfolie: It was arbitrary. The deserialization code for modules assumes they come in this envelope with the `config` field pointing to the module itself and `module_type`, `active`, and `schema_id` being the other required fields. There are also some optional fields.
Embedded predictors don't come with this envelope, so rather than re-doing the serde code I just made a dummy envelope. The info in the `active` and `schema_id` fields gets thrown away upon deserialization so I just chose arbitrary values.

PR #415, opened 2020-07-31
Workflow execution version number

maxhutch:
# Citrine Python PR

## Description 

The `version_number` comes back from the service now; this just adds it to the `WorkflowExecution` model.

### PR Type:
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

### Adherence to team decisions
- [x] I have added tests for 100% coverage
- [x] I have written Numpy-style docstrings for every method and class.
- [ ] I have communicated the downstream consequences of the PR to others.
- [x] I have bumped the version in setup.py

PR #417, opened 2020-08-04
Revert "Workflow execution version number"

maxhutch:
Reverts CitrineInformatics/citrine-python#415


maxhutch:
The bug is actually in the BE, but its eaiser/faster to revert this (which relies on the BE behavior) and re-add it after the BE is working).

kroenlein:
Merged past @maxhutch to expedite nextgen fix

PR #419, opened 2020-08-05
Feature/ingredient identifier in output scope

ventura-rivera:
# Citrine Python PR

## Description 
`scope` is missing from `IngredientIdentifierInOutput`, but it can be handled by the backend. This PR adds that in.

cc'ing: @sdedkins for tracking visibility

### PR Type:

### DIFF
--- a/src/citrine/ara/variables.py
+++ b/src/citrine/ara/variables.py
@@ -581,22 +581,26 @@ class IngredientIdentifierInOutput(Serializable['IngredientIdentifierInOutput'],
     headers = properties.List(properties.String, 'headers')
     ingredient_name = properties.String('ingredient_name')
     process_templates = properties.List(properties.Object(LinkByUID), 'process_templates')
+    scope = properties.String('scope')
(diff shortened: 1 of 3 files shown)


src/citrine/ara/variables.py
```diff
@@ -581,22 +581,26 @@ class IngredientIdentifierInOutput(Serializable['IngredientIdentifierInOutput'],
...
-        return ["name", "headers", "ingredient_name", "process_templates", "type_selector", "typ"]
+        return ["name", "headers", "ingredient_name", "process_templates", "scope",
+                "type_selector", "typ"]
 
     def __init__(self, *,
                  name: str,
                  headers: List[str],
                  ingredient_name: str,
                  process_templates: List[LinkByUID],
+                 scope: str,
```
asantas93, line 597: For backwards compatibility, this should have a default value
```suggestion
                 scope: str = 'id',
```

ventura-rivera: Thanks! I'll commit this

PR #423, opened 2020-08-06
Abstract citrine scope

kroenlein:
# Citrine Python PR

## Description 
The string `'id'` is used for two different but adjacent concepts in citrine-python -- a field name and the Citrine reserved scope.  To improve legibility of code, this PR changes all references to the Citrine reserved scope to CITRINE_SCOPE, which is defined centrally.

### PR Type:


src/citrine/_utils/functions.py
```diff
@@ -86,10 +85,10 @@ def object_to_link_by_uid(json: dict) -> dict:
         uids = json['uids']
         if not isinstance(uids, dict) or not uids:
             return json
-        if 'id' in uids:
-            scope = 'id'
+        if CITRINE_SCOPE in uids:
+            scope = CITRINE_SCOPE
         else:
-            scope = list(uids.keys())[0]
+            scope = next(iter(uids))
```
kroenlein, line 91: This was a clean-up while I was looking at it.  Adding a None return with a more informative error condition if there's no scope could make sense.

src/citrine/gemtables/rows.py
```diff
@@ -1,5 +1,5 @@
 """Row definitions for GEM Tables."""
-from typing import Type, Optional, List  # noqa: F401
+from typing import Type, List  # noqa: F401
```
kroenlein, line 2: This was a clean-up while I was looking at it of an unused import.

src/citrine/gemtables/variables.py
```diff
@@ -1,12 +1,12 @@
 """Variable definitions for GEM Tables."""
 from abc import abstractmethod
-from typing import Type, Optional, List, Tuple, Union  # noqa: F401
+from typing import Type, Optional, List, Union  # noqa: F401
```
kroenlein, line 3: This was a clean-up while I was looking at it of an unused import.

src/citrine/_serialization/serializable.py
```diff
@@ -3,6 +3,8 @@
 
 Self = TypeVar('Self', bound='Serializable')
 
+CITRINE_SCOPE = 'id'
+
```
kroenlein, line 7: This object was chosen for where to define CITRINE_SCOPE because most other reasonable places would cause circular dependencies.  I'm not convinced it's the best place, but there is no clearer better place.

requirements.txt
```diff
@@ -1,4 +1,4 @@
-gemd==0.12.0
+gemd==0.12.1
```
kroenlein, line 1: Minor version bump on the default install of GEMD since a bug fix went out.

kroenlein:
Passed validation against nextgen-devkit: https://jenkins.corp.citrine.io/job/nextgen-ITs/7283

src/citrine/_serialization/serializable.py
maxhutch: What about [here](https://github.com/CitrineInformatics/citrine-python/blob/643e1f49814622a5bea668413e4295d5d7d2a9a9/src/citrine/resources/data_concepts.py#L19)

kroenlein: I missed one; nice catch.  Data Concepts would be a more elegant place for it to live, but `_utils.functions` gets imported there and `_utils.functions` needs the constant, too.  That's the spoiler on circularity.

maxhutch: only `_utils.functions.get_object_id` needs it, so you could move the import into that function body to break the cycle.  That function is not imported by data_concepts.py

src/citrine/resources/data_concepts.py
```diff
@@ -16,8 +16,6 @@
 from gemd.entity.link_by_uid import LinkByUID
 from gemd.json import GEMDJson
 
-CITRINE_SCOPE = 'id'
-
```
kroenlein, line 20: CITRINE_SCOPE constant deleted.

src/citrine/_serialization/serializable.py
kroenlein: I was just about to double check that.  I'll buy that reasoning.  Update in flight.

src/citrine/_utils/functions.py
```diff
@@ -82,14 +80,15 @@ def object_to_link(obj: Any) -> Any:
 
 def object_to_link_by_uid(json: dict) -> dict:
     """Convert an object dictionary into a LinkByUID dictionary, if possible."""
+    from citrine.resources.data_concepts import CITRINE_SCOPE
```
kroenlein, line 83: Also showed up here, but still an improvement.

PR #424, opened 2020-08-11
Sketch PlatformVocabularies to use the data platform to store descriptors

maxhutch:
# Citrine Python PR

## Description 

The PlatformVocabulary is a dict interface into a set of descriptors.
It can be conveniently loaded from the data platform by reading
and then converting templates into descriptors.

Also changed categories to a set.
### PR Type:

### DIFF
--- a/docs/source/workflows/descriptors.rst
+++ b/docs/source/workflows/descriptors.rst
@@ -17,7 +17,7 @@ Real Descriptor
 -----------------------------------
 :class:`~citrine.informatics.descriptors.RealDescriptor` is used to represent continuous variables.
 Each Real Descriptor must provide a lower and upper bound, which is used to both validate input data and as a prior when making predictions.
-If you are not sure what to use bounds, you may want to look at the attribute templates to see if another user has defined bounds for you.
+If you are not sure what bounds to use, you may want to look at the attribute templates to see if another user has defined bounds for you.
 It is better to err on the side of broader bounds rather than narrower ones.
 
 Additionally, each Real Descriptor defines the units of every variable associated with that descriptor.
--- a/docs/source/workflows/descriptors.rst
+++ b/docs/source/workflows/descriptors.rst
@@ -1,0 +1,105 @@
+Descriptors
+==========
+
+Descriptors allow users to define a controlled vocabulary with which to describe the physical context of an AI task.
+Each descriptor defines a term in that vocabulary, which is comprised of a name, a datatype, and bounds on that data type.
+If you are familiar with the GEMD data model, descriptors are roughly equivalent to `AttributeTemplate`s.
+
+The AI Engine currently supports _ kinds of descriptors:
+
+-  `Real Descriptors <#real-descriptor>`__
+-  `Categorical Descriptor <#categorical-descriptor>`__
+-  `Chemical Formula Descriptor <#chemical-formula-descriptor>`__
+-  `Molecular Structure Descriptor <#molecular-structure-descriptor>`__
+-  `Formulation Descriptor <#formulation-descriptor>`__
+
+Real Descriptor
+-----------------------------------
+:class:`~citrine.informatics.descriptors.RealDescriptor` is used to represent continuous variables.
+Each Real Descriptor must provide a lower and upper bound, which is used to both validate input data and as a prior when making predictions.
+If you are not sure what to use bounds, you may want to look at the attribute templates to see if another user has defined bounds for you.
+It is better to err on the side of broader bounds rather than narrower ones.
+
+Additionally, each Real Descriptor defines the units of every variable associated with that descriptor.
+Any `GEMD-compatible <https://citrineinformatics.github.io/gemd-python/depth/unit_parsing.html>`__ unit string may be used.
+If a variable is dimensionless, you can use an empty string.
+
+Categorical Descriptor
+-----------------------------------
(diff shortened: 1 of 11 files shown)


maxhutch:
I'm interested in feedback before writing docs/examples in sphinx and mocking the platform responses to get test coverage up to 100%.

Here's a quick usage example:
```
In [2]: project  = client.projects.get("74bfe934-4dcb-4c8d-adcf-e8d960790126")                          

In [3]: from citrine.builders.descriptors import PlatformVocabulary                                     

In [4]: v = PlatformVocabulary.from_templates(project, "cake-template-preload-66e3a714-5c58-44a6-bc83-04
   ...: ece919230d")                                                                                    

In [5]: list(v)                                                                                         
Out[5]: 
['chemical-formula',
 'molecular-structure',
 'color',
 'toothpick-test',
 'oven-temperature-setting',
 'expected-sample-mass',
 'cooking-time',
 'sample-mass',
 'oven-temperature']

In [6]: v["chemical-formula"]                                                                           
Out[6]: <citrine.informatics.descriptors.ChemicalFormulaDescriptor at 0x7f7072549828>

In [7]: v["cooking-time"]                                                                               
Out[7]: RealDescriptor(Cooking time, 0.0, 168.0, hour)

In [8]: v["sample-mass"]                                                                                
Out[8]: RealDescriptor(Sample Mass, 0.001, 10000.0, gram)
```

latture:
This approach looks good to me.

kroenlein:
I like the idea of directly connecting AttributeTemplates to Descriptors, particularly since making the step of GEM Table -> Predictor would be good UX.  I wonder if it makes sense to be building these in the context of a Project versus building it in the context of a particular Table.  That might address the ambiguity about whether this will be a Formulation or a Chemical Formula, though I understand that's largely handled by "Do you only allow elements?"

This also opens the door to if there should be migrations, either in the Bounds or in the Descriptors, to make the two stacks more compatible in melding.  It's unfortunate that there isn't strict 1:1 correspondence given that we control both stacks and there don't seem to be hard boundaries we're fighting with.

maxhutch:
cc @jspeerless @eantono

docs/source/workflows/descriptors.rst
```diff
@@ -0,0 +1,105 @@
...
+-  `Categorical Descriptor <#categorical-descriptor>`__
+-  `Chemical Formula Descriptor <#chemical-formula-descriptor>`__
+-  `Molecular Structure Descriptor <#molecular-structure-descriptor>`__
+-  `Formulation Descriptor <#formulation-descriptor>`__
+
+Real Descriptor
+-----------------------------------
+:class:`~citrine.informatics.descriptors.RealDescriptor` is used to represent continuous variables.
+Each Real Descriptor must provide a lower and upper bound, which is used to both validate input data and as a prior when making predictions.
+If you are not sure what to use bounds, you may want to look at the attribute templates to see if another user has defined bounds for you.
```
bfolie, line 20:
```suggestion
If you are not sure what bounds to use, you may want to look at the attribute templates to see if another user has defined bounds for you.
```

src/citrine/builders/descriptors.py
```diff
@@ -0,0 +1,125 @@
...
+        parameters = _from_collection(project.parameter_templates)
+        conditions = _from_collection(project.condition_templates)
+
+        res = {}
+        for k, v in chain(properties.items(), parameters.items(), conditions.items()):
+            try:
+                desc = template_to_descriptor(v)
+                res[k] = desc
+            except ValueError:
+                continue
```
bfolie, line 123: We had some problems with the "use exceptions to filter out the valid elements" pattern in the past. Specifically we used to use it when listing modules, and it ended up hiding some actual errors.

tests/builders/test_descriptors.py
```diff
@@ -0,0 +1,137 @@
...
+    assert v["density"] == entries["density"]
+    assert v["pressure"] == entries["pressure"]
+
+
+def test_from_template(fake_project: Project):
+    """Test that only correct scopes and bounds are loaded from templates."""
+    v = PlatformVocabulary.from_templates(fake_project, scope="my_scope")
+    assert len(v) == 1
+    assert list(v) == ["density"]
+    assert v["density"] == density_desc
```
bfolie, line 137: It took me several glances to realize that "volume" doesn't appear because we don't support integer descriptors.

src/citrine/builders/descriptors.py
```diff
@@ -0,0 +1,125 @@
...
+from citrine.resources.project import Project
+
+
+def template_to_descriptor(template: AttributeTemplate) -> Descriptor:
+    """
+    Convert a GEMD attribute template into an AI Engine Descriptor.
+
+    IntBounds cannot be converted because they have no matching descriptor type.
+    CompositionBounds cannot be converted because they may correspond to Formulations or
+    ChemicalFormulas.
```
bfolie, line 21: This doesn't match up with the current code.

tests/builders/test_descriptors.py
maxhutch: Commented for clarity.

src/citrine/builders/descriptors.py
maxhutch: Updated

src/citrine/builders/descriptors.py
maxhutch: Good call.  I sub-classed it and am only matching on that subclass now

PR #428, opened 2020-08-14
Feature/ds builders user warning

jspeerless:
# Citrine Python PR

## Description 
Raise warnings to users of the design space builders when attempting to build design spaces that are of a size to likely cause memory issues. This provides the user insight into why they may receive a MemoryError downstream. (Initiated by user feedback.)

### PR Type:

- [N/A] I have written Numpy-style docstrings for every method and class.

### DIFF
--- a/src/citrine/builders/design_spaces.py
+++ b/src/citrine/builders/design_spaces.py
@@ -2,8 +2,13 @@ try:
     import pandas as pd
 except ImportError:  # pragma: no cover
     raise ImportError('pandas>=0.25 is a requirement for the builders module')
+try:
+    import numpy as np
+except ImportError:  # pragma: no cover
+    raise ImportError('numpy is a requirement for the builders module')
 from itertools import product
 from typing import Mapping, Sequence, List
+from warnings import warn
 from citrine.informatics.design_spaces import EnumeratedDesignSpace
 from citrine.informatics.descriptors import Descriptor, RealDescriptor
 
@@ -36,6 +41,15 @@ def enumerate_cartesian_product(
         description for the EnumeratedDesignSpace
 
     """
+
+    # Check that the grid size is small enough to not cause memory issues
+    grid_size = np.prod(
+        [len(grid_points) for grid_points in design_grid.values()]
+    )
+    if grid_size > 2E8:
+        warn("Product design grid contains {n} grid points. This may cause memory issues downstream."
+             .format(n=grid_size))
+
     design_space_tuples = list(product(*design_grid.values()))
     design_space_cols = list(design_grid.keys())
     df_ds = pd.DataFrame(data=design_space_tuples, columns=design_space_cols)
@@ -118,6 +132,14 @@ def enumerate_formulation_grid(
         if k in non_balance_ingrs
     }
 
+    # Check that the grid size is small enough to not cause memory issues
+    non_balance_grid_size = np.prod(
+        [len(grid_points) for grid_points in non_balance_grids.values()]
+    )
+    if non_balance_grid_size > 2E8:
+        warn("Non-balance formulation grid contains {n} grid points. This may cause memory issues downstream."
+             .format(n=non_balance_grid_size))
+
     # Start by making a naive product design space of non-balance ingredients
     form_ds = pd.DataFrame(
         enumerate_cartesian_product(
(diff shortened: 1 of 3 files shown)


src/citrine/builders/design_spaces.py
```diff
@@ -36,6 +41,14 @@ def enumerate_cartesian_product(
         description for the EnumeratedDesignSpace
 
     """
+    # Check that the grid size is small enough to not cause memory issues
+    grid_size = np.prod(
+        [len(grid_points) for grid_points in design_grid.values()]
```
maxhutch, line 46: Here, the grid_size is the number of rows in `data`

src/citrine/builders/design_spaces.py
```diff
@@ -186,6 +199,13 @@ def cartesian_join_design_spaces(
     if len(all_keys) != len(set(all_keys)):
         raise ValueError('Duplicate keys are not allowed across design spaces')
 
+    # Check that the grid size is small enough to not cause memory issues
+    grid_size = np.prod([len(ds.data) * len(ds.data[0]) for ds in subspaces])
```
maxhutch, line 203: Here, `grid_size` seems to be trying to be the number of cells (rows x columns).  However, I don't think its quite right.  I think you want np.prod([len(ds.data) for ds in subspaces]) * sum([len(ds.data[0]) for ds in subspaces]) since the total number of columns is just the sum (not the product) of the subspaces.

src/citrine/builders/design_spaces.py
jspeerless: Changed to calculate total grid size

src/citrine/builders/design_spaces.py
jspeerless: Changed to (correctly) calculate total grid size.

jspeerless:
Pushed changes to:
 - Correct grid size calculations
 - Test warning text content

tests/builders/test_design_spaces.py
```diff
@@ -191,3 +193,99 @@ def test_exceptions(basic_cartesian_space, simple_mixture_space):
...
+    """Test that warnings are raised"""
+    # Test oversized formulation grid
+    with pytest.raises(UserWarning,
+                       match="Product design grid contains 1562500000 grid points. "
+                             "This may cause memory issues downstream."):
+        # Fail on error (so code stops running)
+        with warnings.catch_warnings():
+            warnings.simplefilter('error')
+            too_big_formulation_grid = {
+                'ing_F': np.linspace(0, 1),
```
maxhutch, line 208: [Blocking] The error state depends on the default [`num=50`](https://numpy.org/doc/stable/reference/generated/numpy.linspace.html), which is a bit opaque.  Could we add that parameter explicitly in the formulations test?  You're already providing it explicitly in the enumerated space tests.

tests/builders/test_design_spaces.py
```diff
@@ -191,3 +193,99 @@ def test_exceptions(basic_cartesian_space, simple_mixture_space):
             name='invalid join space 2',
             description=''
         )
+
+
+def test_oversize_warnings(large_joint_design_space):
+    """Test that warnings are raised"""
+    # Test oversized formulation grid
+    with pytest.raises(UserWarning,
+                       match="Product design grid contains 1562500000 grid points. "
```
maxhutch, line 202: [Non-blocking] This match may be too specific.  I would just do `match="1562500000"` so that the text of the warning can be changed more easily.  Likewise for the other tests.

tests/builders/test_design_spaces.py
```diff
@@ -191,3 +193,99 @@ def test_exceptions(basic_cartesian_space, simple_mixture_space):
...
+                'ing_K': np.linspace(0, 1)
+            }
+            enumerate_formulation_grid(
+                formulation_grid=too_big_formulation_grid,
+                balance_ingredient='ing_K',
+                name='too big mixture space',
+                description=''
+            )
+
+    # Test oversized enumerated space
```
maxhutch, line 222: [Non-blocking] This test is completely independent of the "Test oversized formulation grid", so it would be better for it to be define in its own test.  That way, if "Test oversized formulation grid" failed but this test somehow didn't, you'd see that this test passed.  Also, although these tests are pretty light, breaking code like this into multiple tests provides more opportunity for parallel execution of tests.

Likewise with "Test oversized joined spaces"

jspeerless: All separate now!

tests/builders/test_design_spaces.py
jspeerless: Agreed - done in last push

tests/builders/test_design_spaces.py
jspeerless: Explicitly called out now.

PR #433, opened 2020-08-26
Fix/tableconfig missing kwargs

ventura-rivera:
# Citrine Python PR

## Description 
add_columns and add_all_ingredients do not return all of the TableConfig kwargs. This fix adds the missing kwargs in.

### PR Type:

### DIFF
--- a/src/citrine/resources/table_config.py
+++ b/src/citrine/resources/table_config.py
@@ -156,7 +156,11 @@ class TableConfig(Resource["TableConfig"]):
             datasets=copy(self.datasets),
             rows=copy(self.rows),
             variables=copy(self.variables) + [variable],
-            columns=copy(self.columns) + columns
+            columns=copy(self.columns) + columns,
+            config_uid=copy(self.config_uid),
+            version_uid=copy(self.version_uid),
+            version_number=self.version_number,
+            definition_uid=copy(self.definition_uid)
         )
 
     def add_all_ingredients(self, *,
@@ -228,7 +232,11 @@ class TableConfig(Resource["TableConfig"]):
             datasets=copy(self.datasets),
             rows=copy(self.rows),
             variables=copy(self.variables) + new_variables,
-            columns=copy(self.columns) + new_columns
+            columns=copy(self.columns) + new_columns,
+            config_uid=copy(self.config_uid),
+            version_uid=copy(self.version_uid),
+            version_number=self.version_number,
+            definition_uid=copy(self.definition_uid)
         )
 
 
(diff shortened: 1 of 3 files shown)


ventura-rivera:
Adding @jspeerless for visibility.

src/citrine/resources/table_config.py
```diff
@@ -156,7 +156,11 @@ def add_columns(self, *,
             datasets=copy(self.datasets),
             rows=copy(self.rows),
             variables=copy(self.variables) + [variable],
-            columns=copy(self.columns) + columns
+            columns=copy(self.columns) + columns,
+            config_uid=copy(self.config_uid),
+            version_uid=copy(self.version_uid),
```
maxhutch, line 161: I don't think we want to copy the version_uid, since that is provided by the service when the config is registered.
```suggestion
```

src/citrine/resources/table_config.py
```diff
@@ -156,7 +156,11 @@ def add_columns(self, *,
...
+            version_number=self.version_number,
```
maxhutch, line 162: I don't think we want to copy the version_number, since that is provided by the service when the config is registered.
```suggestion
```

src/citrine/resources/table_config.py
```diff
@@ -228,7 +232,11 @@ def add_all_ingredients(self, *,
             datasets=copy(self.datasets),
             rows=copy(self.rows),
             variables=copy(self.variables) + new_variables,
-            columns=copy(self.columns) + new_columns
+            columns=copy(self.columns) + new_columns,
+            config_uid=copy(self.config_uid),
+            version_uid=copy(self.version_uid),
```
maxhutch, line 237: I don't think we want to copy the version_uid, since that is provided by the service when the config is registered.
```suggestion
```

src/citrine/resources/table_config.py
```diff
@@ -228,7 +232,11 @@ def add_all_ingredients(self, *,
...
+            version_number=self.version_number,
```
maxhutch, line 238: I don't think we want to copy the version_number, since that is provided by the service when the config is registered.
```suggestion
```

ventura-rivera:
> This looks right, but could we add a test? I think it should be straight-forward to extend the existing [`test_add_columns`](https://github.com/CitrineInformatics/citrine-python/blob/master/tests/resources/test_table_config.py#L242) to stick a config_uid in the empty_defn at the beginning and check that it is preserved in the output.

Good point, I'll extend that test!

PR #435, opened 2020-08-28
Feature/pla 5184 optional dataset unique name

dleblanc:
# Added an optional unique_name to datasets

This is dependent on backend PR: https://github.com/CitrineInformatics/platform-backend/pull/1427

This adds backwards-compatible support for the optional unique_name property. This allows us to have atomic create-or-update sematics for a dataset by supplying a unique name.

Provided the ability to use this in the Dataset collection's .register() call, but it's disabled by default (you can enable it by turning on a flag in the session). This is so we don't break when talking to old backends.

Also added a get_with_unique_name() method (please help with naming) which allows us to directly look up a dataset. This should help make testing (seeding?) and similar operations much faster and reliable, and reduce our reliance on locks.

### PR Type:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,7 @@ class PreBuildCommand(build_py):
 
 
 setup(name='citrine',
-      version='0.63.3',
+      version='0.63.1',
       url='http://github.com/CitrineInformatics/citrine-python',
       description='Python library for the Citrine Platform',
       author='Citrine Informatics',
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,7 @@ class PreBuildCommand(build_py):
 
 
 setup(name='citrine',
-      version='0.62.3',
+      version='0.63.3',
       url='http://github.com/CitrineInformatics/citrine-python',
       description='Python library for the Citrine Platform',
       author='Citrine Informatics',
--- a/src/citrine/resources/dataset.py
+++ b/src/citrine/resources/dataset.py
@@ -8,6 +8,7 @@ from citrine._rest.resource import Resource
 from citrine._serialization import properties
 from citrine._session import Session
 from citrine._utils.functions import scrub_none
+from citrine.exceptions import NotFound
 from citrine.resources.condition_template import ConditionTemplateCollection
 from citrine.resources.file_link import FileCollection
 from citrine.resources.ingredient_run import IngredientRunCollection
@@ -360,3 +361,17 @@ class DatasetCollection(Collection[Dataset]):
         full_model = self.build(data)
         full_model.project_id = self.project_id
         return full_model
+
+    def get_with_unique_name(self, unique_name: str) -> ResourceType:
+        """Get a Dataset with the given unique name."""
+        if unique_name is None:
+            raise ValueError("You must supply a unique_name")
+        path = self._get_path() + "?unique_name=" + unique_name
+        data = self.session.get_resource(path)
+
+        if len(data) == 1:
+            return self.build( data[0])
+        elif len(data) > 1:
+            raise RuntimeError("Received multiple results when requesting a unique dataset")
+        else:
+            raise NotFound(path)
--- a/src/citrine/resources/dataset.py
+++ b/src/citrine/resources/dataset.py
@@ -347,7 +347,16 @@ class DatasetCollection(Collection[Dataset]):
         path = self._get_path()
         dumped_dataset = model.dump()
         dumped_dataset["deleted"] = None
-        data = self.session.post_resource(path, scrub_none(dumped_dataset))
+
+        # Only use the idempotent put approach if a) a unique name is provided, and b) the session is configured to use
+        # it (default to False for backwards compatibility).
+        if model.unique_name is not None and self.session.use_idempotent_dataset_put:
+            # Leverage the create-or-update endpoint if we've got a unique name
+            data = self.session.put_resource(path, scrub_none(dumped_dataset))
+        else:
+            # Otherwise fall back to using the POST approach (which may fail if the resource already exists)
+            data = self.session.post_resource(path, scrub_none(dumped_dataset))
+
         full_model = self.build(data)
         full_model.project_id = self.project_id
         return full_model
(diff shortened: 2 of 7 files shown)


setup.py
```diff
@@ -57,7 +57,7 @@ def run(self):
 
 
 setup(name='citrine',
-      version='0.62.3',
+      version='0.63.3',
```
kroenlein, line 60: Strange version bump

src/citrine/resources/dataset.py
```diff
@@ -345,7 +348,31 @@ def register(self, model: Dataset) -> Dataset:
         path = self._get_path()
         dumped_dataset = model.dump()
         dumped_dataset["deleted"] = None
-        data = self.session.post_resource(path, scrub_none(dumped_dataset))
+
+        # Only use the idempotent put approach if a) a unique name is provided, and b)
+        # the session is configured to use it (default to False for backwards compatibility).
+        if model.unique_name is not None and self.session.use_idempotent_dataset_put:
+            # Leverage the create-or-update endpoint if we've got a unique name
+            data = self.session.put_resource(path, scrub_none(dumped_dataset))
```
kroenlein, line 356: Is this the correct step if a user has accidentally stumbled upon someone else's `unique_name`?  This seems like a particular risk if there's no way for a user to ask the platform what `unique_name`s have been used.

tests/resources/test_dataset.py
```diff
@@ -82,6 +84,74 @@ def test_register_dataset(collection, session):
...
+    dataset = collection.register(DatasetFactory(name=name, summary=summary, description=description, unique_name=unique_name))
+
+    expected_call = FakeCall(
+        method='PUT',
+        path='projects/{}/datasets'.format(collection.project_id),
+        json={'name': name, 'summary': summary, 'description': description, 'unique_name': unique_name}
+    )
+    assert session.num_calls == 1
+    assert expected_call == session.last_call
+    assert name == dataset.name
```
kroenlein, line 106: This has no failure modes being tested.

src/citrine/resources/dataset.py
djack: The POST endpoint will fail if the unique name is already in use. The PUT will update. Currently it's flagged off, but this is the intended behavior.

setup.py
dleblanc: Oh I see what you mean, yeah I'll fix that

src/citrine/resources/dataset.py
dleblanc: Dj is right. As of right now there are no datasets with a unique name, because the migration defaults to null. If a user tries to create (Post) two new distinct datasets with the same unique name, they'll get an error on the second one. That's by design.

For the create-or-update (Put) case, we specifically choose to update the given dataset on the second call (as identified by the unique name). This maps to the intended contract of .register(). We're just not using this by default because of backwards compatibility concerns. I imagine we'll use that by default at some stage (assuming the unique name is provided).

kroenlein: So what happens once this is deployed, and the second user tries to create a dataset with `unique_name = "my_dataset"`?  What is actually at risk of being clobbered? As best as I can think, this might mess up `name`, `summary` and `description`, but should be able to impact dataset ownership or access control, right?

dleblanc: An update won't change any grants, that only happens on creation (and is the same for POST vs PUT).

Yeah, the second write will overwrite the first one with the supplied properties. The actual query updates: name, summary, description, updated_by and updated_time. The UUID, unique_name, and all grants will stay the same.

tests/resources/test_dataset.py
dleblanc: With this method we test:
 - unique_name being None
 - no values being returned (this isn't an error, but is a branch)
 - > 1 results returned (that is an error)
 - exactly one value being returned

The actual endpoint can fail with an inappropriately formatted unique_name, but that's tested in the backend.

Not sure what else you're aiming to test? We already have coverage around session.get_resource, and collection.build. The purpose of this test is to ensure that with the given options that we invoke a PUT with the appropriate URL, and not the POST.

src/citrine/resources/dataset.py
dleblanc: And just for the curious: when we do an update on a dataset we don't have WRITE access on - we actually perform the UPSERT operation, get back the ID it's using, and check for Write access after the fact. If that fails, then the whole transaction gets rolled back and no updates are made. Kind of cool actually.

tests/resources/test_dataset.py
kroenlein: > unique_name being None
You mean this scenario is covered by the existing `test_register_dataset`?

It's probably fine.  Part of the problem of trying to do PRs during meetings, probably.

src/citrine/resources/dataset.py
djack: @kroenlein is it your contention (it seems like it is) that people should be able to create a dataset knowing only the unique name, but that they should only be able to PUT / UPDATE a dataset using an ID? That seems like a user-hostile constraint to me. If a user wants to be certain their chosen `unique_name` is available, they can do a GET on the unique name before attempting a PUT, or they can use the POST method. PUT/UPDATE should operate as exactly that - createOrUpdate.

kroenlein: I'm just trying to figure out if we've got a nasty tripping hazard.  Stumbling across an existing unique_name seems reasonably likely, so if the outcome would be a rename and objects in the wrong data set (hopefully discovered quickly), that's not so bad.  Ideally, a user trying to create a new dataset with an existing unique_name would bounce, but it's not obvious to me that we could differentiate between that case and the update case.  

Maybe if you want to update a dataset, you need to match both ID & unique_name?  It is a little user hostile, but a GET -> update -> PUT doesn't seem terrible if you want to edit a record on the platform.  I'm not gonna block on this.

dleblanc: If we don't want to klobber the contents of an existing dataset, we can call a POST with that unique name, catch the error, and handle it appropriately. Perhaps the distinction between create and update should be called out in the client, as .register() conflates the two.

For now I think this is fine (there's no documented client-visible change yet), but maybe we should think about parameters to DatasetCollection.register, or other ways to prevent mistakes.


The main benefit for us is that IF we're confident that our unique name is accurate, then we have one call that creates or updates the dataset based on that single criteria. This is particularly important for parallel e2e tests because this call is atomic and guaranteed to give consistent results. That's probably not so important for a typical user though.

PR #436, opened 2020-09-01
Bugfix/deprecate filter by and fix data object list

djack:
# Citrine Python PR

## Description 
The Data object "list" and "filter" methods are very slow, and fail with large numbers of data objects in a dataset. This PR:
- Adds deprecation warnings to all remaining "filter_by" methods. 
- replaces the implementation of "list" with a simple call to "list_all" that wraps the iterator in a list comprehension. 

### PR Type:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-gemd==0.13.0
+gemd==0.13.2
 requests==2.22.0
 pyjwt==1.7.1
 arrow==0.15.4
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-gemd==0.12.1
+gemd==0.13.0
 requests==2.22.0
 pyjwt==1.7.1
 arrow==0.15.4
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,7 @@ class PreBuildCommand(build_py):
 
 
 setup(name='citrine',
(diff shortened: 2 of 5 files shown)


setup.py
```diff
@@ -57,7 +57,7 @@ def run(self):
 
 
 setup(name='citrine',
-      version='0.63.1',
+      version='0.64.1',
```
kroenlein, line 60:
```suggestion
      version='0.64.0',
```
Is there a reason you didn't reset patch value?

kroenlein (dismissed):
Looks good.  Minor question, and a request to bump `gemd==0.13.0` in requirements.txt while you are in there.  Neither blocking.

setup.py
djack: not a good one

kroenlein (dismissed):

requirements.txt
```diff
@@ -1,4 +1,4 @@
-gemd==0.12.1
+gemd==0.13.0
```
kroenlein, line 1:
```suggestion
gemd==0.13.2
```
Releasing a 0.14 would be good, but that edition wasn't published because the feature was needed for the multicake example that I shelved for unique_name...

kroenlein (approved):
Makes sense

kroenlein (approved):

PR #439, opened 2020-09-03
Deduplicate list results by their uid

maxhutch:
# Citrine Python PR

## Description 

Since we're doing pagination-based listing, creates and deletes
to the collection being listed can result in index movement during
a multi-request list operation.  In particular, the repetition of
a list element from the end of one page to the beginning of the next
has been observed to occur when items are added to the list.  This
change keeps a list of the uids that have been encountered and skips
returning duplicates. If the listed object isn't a resource and
therefore doesn't have a uid, then no deduplication is performed.

This behavior can be toggled via a flag to assist future developers.

### PR Type:

### DIFF
--- a/src/citrine/resources/gemtables.py
+++ b/src/citrine/resources/gemtables.py
@@ -122,7 +122,9 @@ class GemTableCollection(Collection[GemTable]):
             for item in collection:
                 yield self.build(item)
 
-        return self._paginator.paginate(fetch_versions, build_versions, page, per_page)
+        return self._paginator.paginate(
+            # Don't deduplicate on uid since uids are shared between versions
+            fetch_versions, build_versions, page, per_page, deduplicate=False)
 
     def list_by_config(self,
                        table_config_uid: UUID,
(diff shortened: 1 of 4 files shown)


talltea (dismissed):
lgtm

src/citrine/resources/gemtables.py
```diff
@@ -122,7 +122,9 @@ def build_versions(collection: Iterable[dict]) -> Iterable[GemTable]:
             for item in collection:
                 yield self.build(item)
 
-        return self._paginator.paginate(fetch_versions, build_versions, page, per_page)
+        return self._paginator.paginate(
+            # Don't deduplicate on uid since uids are shared between versions
```
talltea, line 126: I'm guessing in the future we'll want to deduplicate on uid + version?

talltea (approved):

src/citrine/resources/gemtables.py
maxhutch: Or improve pagination more comprehensively.

PR #440, opened 2020-09-04
waiting utilities

sdedkins:
# Citrine Python PR

## Description 

### PR Type:


src/citrine/waiting/waiting.py
```diff
@@ -0,0 +1,77 @@
...
+from typing import Callable
+from pprint import pprint
+from citrine._rest.collection import Collection
+from citrine.resources.workflow_executions import (
+    WorkflowExecution,
+    WorkflowExecutionStatus,
+)
+
+
+def _wait_until(condition: Callable[[], bool], timeout: float = 900.0, interval: float = 3.0):
```
alexandraj777, line 11: The code would be more straightforward and readable if this `_wait_until` function was inline in `wait_until_validated` and `wait_until_finished`, rather than abstracted. Given that this function is only used twice, and requires the definition of separate functions that are passed around, you don't seems to be saving any lines of code by this abstraction.

src/citrine/waiting/waiting.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def wait_until_validated(
+    collection: Collection,
+    module,
+    print_status_info: bool = False,
+    timeout: float = 1800.0,
+    interval: float = 3.0,
+):
+    """Wait until module is validated."""
```
maxhutch, line 52: This will need a full docstring, since it is a public member.  Be sure to include the units for the times (seconds, right?)

src/citrine/waiting/waiting.py
```diff
@@ -0,0 +1,77 @@
...
+    _wait_until(is_validated, timeout=timeout, interval=interval)
+
+    if print_status_info:
+        print("\nStatus info:")
+        status_info = collection.get(module.uid).status_info
+        pprint(status_info)
+
+
+def wait_until_execution_is_finished(execution: WorkflowExecution):
+    """Wait until execution is finished."""
```
maxhutch, line 69: This will need a full docstring as well.

src/citrine/waiting/waiting.py
```diff
@@ -0,0 +1,77 @@
...
+        _print_validation_status(status, start)
+        return status != "VALIDATING"
+
+    _wait_until(is_validated, timeout=timeout, interval=interval)
+
+    if print_status_info:
+        print("\nStatus info:")
+        status_info = collection.get(module.uid).status_info
+        pprint(status_info)
+
```
maxhutch, line 66: We may want to return the final gotten module here.  Otherwise, the user will have to issue a follow-up `get` in order to access the status and status info.

src/citrine/waiting/waiting.py
maxhutch: I agree that inlining probably makes more sense.  In particular, it should allow for better error messaging, because the logical content of the condition can be used to populate the error message.

src/citrine/waiting/waiting.py
```diff
@@ -0,0 +1,77 @@
...
+    WorkflowExecutionStatus,
+)
+
+
+def _wait_until(condition: Callable[[], bool], timeout: float = 900.0, interval: float = 3.0):
+    """Poll until the provided condition is truthy or the timeout is reached."""
+    start = time()
+    while not condition() and time() - start < timeout:
+        sleep(interval)
+    assert condition(), "Timeout reached, but condition is still False."
```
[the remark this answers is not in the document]
maxhutch: I don't think we want to be throwing an assertion error here.  I think a `RuntimeError` would be more appropriate.  It would be particularly good to define our own error subclass (could be just at the top of this file) and throw that so that callers can catch that error specifically.  Here is [an example](https://github.com/CitrineInformatics/citrine-python/blob/master/src/citrine/builders/descriptors.py#L15) of that pattern.

src/citrine/waiting/waiting.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def _print_validation_status(
+    status: str, start_time: float, line_start: str = "", line_end: str = "\r"
+):
+    print(
+        f"{line_start}Status = {status:<25}Elapsed time"
```
maxhutch, line 23: f-strings aren't included in python 3.5, which is a version that we do current support.  Can you replace these with calls to [format](https://docs.python.org/3/library/stdtypes.html#str.format)?

src/citrine/waiting/waiting.py
```diff
@@ -0,0 +1,77 @@
...
+    status_text = status.status
+    output_text = status_text if status_text != "InProgress" else "In progress"
+    return output_text
+
+
+def _print_execution_status(
+    status: WorkflowExecutionStatus, start_time: float, line_end: str = "\r"
+):
+    print(
+        f"Status = {_pretty_execution_status(status):<25}Elapsed time"
```
maxhutch, line 39: f-strings aren't included in python 3.5, which is a version that we do current support.  Can you replace these with calls to [format](https://docs.python.org/3/library/stdtypes.html#str.format)?

src/citrine/waiting/waiting.py
```diff
@@ -0,0 +1,77 @@
...
+        " = {str(int(time() - start_time)).rjust(3)}s",
+        end=line_end,
+    )
+
+
+def wait_until_validated(
```
maxhutch, line 45: This method name implies to me that the module should always be VALID after it completes successfully.  Maybe something like `wait_while_validating` would imply less about the final state.

src/citrine/waiting/waiting.py
```diff
@@ -0,0 +1,77 @@
...
+
+    _wait_until(is_validated, timeout=timeout, interval=interval)
+
+    if print_status_info:
+        print("\nStatus info:")
+        status_info = collection.get(module.uid).status_info
+        pprint(status_info)
+
+
+def wait_until_execution_is_finished(execution: WorkflowExecution):
```
maxhutch, line 68: How about "wait_while_executing"

maxhutch: Why no timeout or interval?

maxhutch (commented):
I made several comments inline.

Big picture, I think it would be really valuable to take this opportunity to unify the interface.  That could mean:
 * A single `wait` method that accepts either an execution or a module
 * Getting updates to the module from the attached session rather than passing in an explicit collection to the wait method.

If that's too big of a bite, we can leave it at separate methods like it is now, but I think we should have that unified interface in mind.

maxhutch:
> 1. Should we use mocks or end-to-end tests to test this?

We should use mocks to achieve code coverage, but migrating the end-to-end tests to use this method will give us broad behavioral coverage.

> 2. Should we create a directory for modules that contain small convenience functions like this? e.g. things like find_or_create would go there rather than being in seeding.

I have a slight preference to avoid catch-all `util` modules.  How about `citrine.async.waiting`.  I could imagine, for example, adding `citrine.async.manager` to include a CLI task manager or `citrine.async.progress` to include progress bars.

sdedkins:
@maxhutch async is a protected keyword in python >=3.7 -> this is why PR checks are currently failing for those versions. How about using sync, which is not protected?

sdedkins:
> * Getting updates to the module from the attached session rather than passing in an explicit collection to the wait method.

@maxhutch I cannot see how to get the session and the collection to which the module belongs from an object of type Module. Can you point me in the right direction?

maxhutch:
> @maxhutch async is a protected keyword in python >=3.7 -> this is why PR checks are currently failing for those versions. How about using sync, which is not protected?

`sync` is the opposite of `async`, so we definitely don't want that.  We could do full-on `asynchronous`, or maybe `jobs`.  My preference against `utils` or `helpers` is slight -- I think we can do better but those are acceptable if we can't come up with anything.

maxhutch:
> @maxhutch I cannot see how to get the session and the collection to which the module belongs from an object of type Module. Can you point me in the right direction?

This is [the pattern](https://github.com/CitrineInformatics/citrine-python/blob/master/src/citrine/resources/workflow_executions.py#L62) that you'd be trying to follow, but outside of the object rather than inside of it.  You can see that [workflow executions](https://github.com/CitrineInformatics/citrine-python/blob/master/src/citrine/resources/workflow_executions.py#L47), and [modules](https://github.com/CitrineInformatics/citrine-python/blob/master/src/citrine/informatics/predictors.py#L163) both have a Session in them that is set if the module is returned by `get`.  You could use that.  If implementing that gets too complicated, then I think its ok to just use the collection.

sdedkins:
> > @maxhutch async is a protected keyword in python >=3.7 -> this is why PR checks are currently failing for those versions. How about using sync, which is not protected?
> 
> `sync` is the opposite of `async`, so we definitely don't want that. We could do full-on `asynchronous`, or maybe `jobs`. My preference against `utils` or `helpers` is slight -- I think we can do better but those are acceptable if we can't come up with anything.

Moved to jobs

sdedkins:
@maxhutch Thanks for your answers. This is ready for review.

src/citrine/jobs/waiting.py
```diff
@@ -0,0 +1,133 @@
...
+    )
+
+
+def wait_while_validating(
+    collection: Collection[Module],
+    module: Module,
+    print_status_info: bool = False,
+    timeout: float = 1800.0,
+    interval: float = 3.0,
+):
```
maxhutch, line 49:
```suggestion
) -> Module:
```

src/citrine/jobs/waiting.py
```diff
@@ -0,0 +1,133 @@
...
+        Whether to print status info, by default False
+    timeout : float, optional
+        Maximum time spent inquiring in seconds, by default 1800.0
+    interval : float, optional
+        Inquiry interval in seconds, by default 3.0
+
+    Returns
+    -------
+    Module
+        Module in Collection
```
maxhutch, line 69:
```suggestion
        Module in Collection after validation
```

src/citrine/jobs/waiting.py
```diff
@@ -0,0 +1,133 @@
...
+    Raises
+    ------
+    ConditionTimeoutError
+        If fails to validate within timeout
+
+    """
+    def is_validated():
+        status = collection.get(module.uid).status
+        _print_validation_status(status, start)
+        return status != "VALIDATING"
```
maxhutch, line 80:
```suggestion
        return status != "VALIDATING" and status != "INPROGRESS"
```
In the new status system, we use "INPROGRESS" consistently across all jobs.  This change would make it forward compatible.

src/citrine/jobs/waiting.py
```diff
@@ -0,0 +1,133 @@
...
+    return collection.get(module.uid)
+
+
+def wait_while_executing(
+    execution: WorkflowExecution,
+    timeout: float = 1800.0,
+    interval: float = 3.0,
+):
+    """
+    Wait until execution is finished.
```
maxhutch, line 104:
```suggestion
    [ALPHA] Wait until execution is finished.
```

All executions are currently alpha and I know this signature will have to change, so its worth noting.

src/citrine/jobs/waiting.py
```diff
@@ -0,0 +1,133 @@
...
+
+    Raises
+    ------
+    ConditionTimeoutError
+        If fails to validate within timeout
+
+    """
+    def is_validated():
+        status = collection.get(module.uid).status
+        _print_validation_status(status, start)
```
maxhutch, line 79: I think this should only be called if `print_status_info` is true.

src/citrine/jobs/waiting.py
```diff
@@ -0,0 +1,133 @@
...
+        pprint(status_info)
+
+    return collection.get(module.uid)
+
+
+def wait_while_executing(
+    execution: WorkflowExecution,
+    timeout: float = 1800.0,
+    interval: float = 3.0,
+):
```
maxhutch, line 102: I suggest going ahead and returning the execution here too, so that the pattern is more consistent with module validation.
```suggestion
) -> WorkflowExecution:
```

src/citrine/jobs/waiting.py
```diff
@@ -0,0 +1,133 @@
...
+        _print_execution_status(status, start)
+        return not execution.status().in_progress
+
+    start = time()
+
+    while not execution_is_finished() and (time() - start < timeout):
+        sleep(interval)
+    if not execution_is_finished():
+        msg = "Timeout reached, but condition is still {}".format(execution_is_finished())
+        raise ConditionTimeoutError(msg)
```
maxhutch, line 133:
```suggestion
        raise ConditionTimeoutError(msg)
    return execution
```
See comment on the signature.

src/citrine/jobs/waiting.py
```diff
@@ -0,0 +1,133 @@
...
+
+    Raises
+    ------
+    ConditionTimeoutError
+        If fails to finish execution within timeout
+
+    """
+    def execution_is_finished():
+        status = execution.status()
+        _print_execution_status(status, start)
```
maxhutch, line 124: Can we add `print_status_info` to this method too?

maxhutch (changes requested):
The test coverage looks good!  I have a few minor comments on the behavior, mostly around consistency.

The branch currently conflicts with master -- I think it just needs a version bump.

sdedkins:
@maxhutch I think I have addressed all of your requested changes. Thanks for all of your feedback.

src/citrine/jobs/waiting.py
```diff
@@ -0,0 +1,144 @@
...
+    Raises
+    ------
+    ConditionTimeoutError
+        If fails to finish execution within timeout
+
+    """
+    def execution_is_finished():
+        status = execution.status()
+        if print_status_info:
+            _print_execution_status(status, start)
```
maxhutch, line 134: `start` seems to be a forward reference: it isn't actually defined until line 137.  The fact that this works is a testament to the loose nature of python, but its still worth fixing to be less confusing to future developers (and possibly static analysis tools).

src/citrine/jobs/waiting.py
```diff
@@ -0,0 +1,144 @@
...
+    Raises
+    ------
+    ConditionTimeoutError
+        If fails to validate within timeout
+
+    """
+    def is_validated():
+        status = collection.get(module.uid).status
+        if print_status_info:
+            _print_validation_status(status, start)
```
maxhutch, line 80: `start` seems to be a forward reference: it isn't actually defined until line 83.  The fact that this works is a testament to the loose nature of python, but its still worth fixing to be less confusing to future developers (and possibly static analysis tools).

src/citrine/jobs/waiting.py
```diff
@@ -0,0 +1,144 @@
...
+    ------
+    ConditionTimeoutError
+        If fails to finish execution within timeout
+
+    """
+    def execution_is_finished():
+        status = execution.status()
+        if print_status_info:
+            _print_execution_status(status, start)
+        return not execution.status().in_progress
```
maxhutch, line 135:
```suggestion
        return status.in_progress
```
The status was already fetched on line 132.

maxhutch (commented):
A couple minor comments; last round of them, I bet.

sdedkins:
@maxhutch Thanks for catching those. I think I have made the necessary changes now. Definitely happy to make further changes if necessary.

maxhutch (approved):
LGTM!

PR #441, opened 2020-09-10
PredictorEvaluationWorkflow -- top to bottom

maxhutch:
This PR adds the `PredictorEvaluationWorkflow` add associated objects. This workflow accepts a list of `PredictorEvaluator`s and runs each evaluation on a predictor when triggered. Unlike other workflow executions, a `PredictorEvaluationExecution ` supports `status_info` and results are lazily loaded by evaluator name. Results for each response are Python objects (instead of a raw `dict` like other execution results) and keyed by metric.

### PR Type:


latture:
@maxhutch This is still in progress, but it's ready for a high-level design review. I made some changes from you original PR. I'll call out questions I had when I was going through it with inline comments.

src/citrine/informatics/workflows.py
```diff
@@ -162,3 +163,67 @@ def executions(self) -> WorkflowExecutionCollection:
         if getattr(self, 'project_id', None) is None:
             raise AttributeError('Cannot initialize execution without project reference!')
         return WorkflowExecutionCollection(self.project_id, self.uid, self.session)
+
+
+class PredictorEvaluationWorkflow(Resource['PredictorEvaluationWorkflow'], Workflow):
```
latture, line 168: @maxhutch I relocated this workflow to the informatics package. Is there reason you didn't include this class with the other workflows in your original PR?

src/citrine/resources/predictor_evaluation_workflow.py
```diff
@@ -1,83 +1,21 @@
...
-    name: str
-        name of the predictor evaluation workflow
-    description: str
-        the description of the predictor evaluation workflow
-    evaluators: List[PredictorEvaluator]
-        the list of evaluators to apply to the predictor
-
-    """
-
-    _response_key = 'WorkflowExecutions'
```
latture: @maxhutch Was this a copy-paste error or intentional?

src/citrine/resources/predictor_evaluation_workflow.py
```diff
@@ -1,83 +1,21 @@
...
-        return '<PredictorEvaluationWorkflow {!r}>'.format(str(self.uid))
-
-    def _path(self):
-        return '/projects/{project_id}/workflows/{workflow_id}/executions/{execution_id}'.format(
-            **{
-                "project_id": self.project_id,
-                "workflow_id": self.workflow_id,
-                "execution_id": self.uid
-            }
-        )
```
latture: @maxhutch Why was `_path` specified here?

src/citrine/informatics/workflows.py
maxhutch: We just have two patterns -- one where the resources are in `resources` and another where they are in `informatics`.  I'm not opposed to putting it in `informatics` but it is different enough from the other workflows that I'd suggest keeping it in a separate file.

src/citrine/resources/predictor_evaluation_workflow.py
maxhutch: Copy-paste.  I didn't change much plumbing.

src/citrine/resources/predictor_evaluation_workflow.py
maxhutch: I think it was just a convenience method for the piping.

src/citrine/resources/predictor_evaluation_execution.py
```diff
@@ -0,0 +1,118 @@
...
+        The first execution has version_number = 1.
+
+    """
+
+    _response_key = 'WorkflowExecutions'
+
+    uid = properties.UUID('id')
+    project_id = properties.UUID('project_id', deserializable=False)
+    workflow_id = properties.UUID('workflow_id', deserializable=False)
+    version_number = properties.Integer("version_number")
```
maxhutch, line 35: Does `version_number` make sense here in the way that it is used for design workflows?  I think we might want to ax it.

src/citrine/informatics/predictor_evaluation_result.py
```diff
@@ -0,0 +1,245 @@
...
+    _response_results = properties.Mapping(properties.String, properties.Object(ResponseMetrics), "response_results")
+    typ = properties.String('type', default='CrossValidationResult', deserializable=False)
+
+    def __init__(self, *,
+                 evaluator: PredictorEvaluator,
+                 response_results: Mapping[str, ResponseMetrics]):
+        self._evaluator = evaluator
+        self._response_results = response_results
+
+    def __getitem__(self, item):
```
maxhutch, line 235: This behavior should probably be documented in the class docstring.

maxhutch (commented):
Overall, the user-facing classes here look right to me.  We'll need to review the parts of the impl that make the actual API requests when the API is setup.

src/citrine/resources/predictor_evaluation_execution.py
latture: I'll remove it.

src/citrine/informatics/predictor_evaluation_result.py
latture: 👍

src/citrine/informatics/predictor_evaluation_result.py
```diff
@@ -0,0 +1,245 @@
...
+    """
+
+    @classmethod
+    def get_type(cls, data) -> Type[Serializable]:
+        """Return the subtype."""
+        return {
+            "CrossValidationResult": CrossValidationResult,
+        }[data["type"]]
+
+    def evaluator(self) -> PredictorEvaluator:
```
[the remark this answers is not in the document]
latture: Unless we inherit from `ABC` using `@abstractmethod` won't provide any utility. If we inherit from `ABC`, however, we can't instantiate the parent class which would break our serialization model. We could create an abstract protocol class that defines these abstract methods, but that seems like unnecessary overhead. I'm actually in favor of leaving the `NotImplementedError`.

src/citrine/resources/predictor_evaluation_execution.py
```diff
@@ -0,0 +1,118 @@
...
+    status = properties.Optional(properties.String(), 'status', serializable=False)
+    status_info = properties.Optional(
+        properties.List(properties.String()),
+        'status_info',
+        serializable=False
+    )
+
```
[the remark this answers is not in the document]
latture: 👍

src/citrine/informatics/workflows.py
```diff
@@ -162,3 +163,67 @@ def executions(self) -> WorkflowExecutionCollection:
...
+    experimental_reasons = properties.Optional(
+        properties.List(properties.String()),
+        'experimental_reasons',
+        serializable=False
+    )
+    active = properties.Boolean('active', default=True)
+    archived = properties.Boolean('archived', default=False)
+    created_by = properties.Optional(properties.UUID, 'created_by', serializable=False)
+    create_time = properties.Optional(properties.Datetime, 'create_time', serializable=False)
+    module_type = properties.String('module_type', default='PREDICTOR_EVALUATION_WORKFLOW')
```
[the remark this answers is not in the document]
latture: We currently use `module_type` to deserialize workflows. If we delete this, we'll need to come up with another mechanism.

src/citrine/informatics/predictor_evaluation_result.py
```diff
@@ -0,0 +1,257 @@
...
+
+    @classmethod
+    def get_type(cls, data) -> Type[Serializable]:
+        """Return the subtype."""
+        return {
+            "CrossValidationResult": CrossValidationResult,
+        }[data["type"]]
+
+    @property
+    def evaluator(self) -> PredictorEvaluator:
```
latture, line 202: How do we feel about converting this and the other abstract methods to properties? (We currently use `@property` to access workflow executions, so a similar pattern is used elsewhere in the SDK.)

src/citrine/informatics/predictor_evaluation_result.py
```diff
@@ -0,0 +1,245 @@
...
+    typ = properties.String('type', default='RealMetricValue', deserializable=False)
+
+    def __init__(self, *,
+                 mean: float,
+                 standard_error: Optional[float]):
+        self.mean = mean
+        self.standard_error = standard_error
+
+
+class PredictedVsActual(PolymorphicSerializable["PredictedVsActual"]):
```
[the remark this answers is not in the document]
latture: 👍 
I'm not sure either. We'll see once we start writing tests.

src/citrine/informatics/workflows/__init__.py
```diff
@@ -0,0 +1,4 @@
+from .design_workflow import *
+from .predictor_evaluation_workflow import *
+from .performance_workflow import *
+from .workflow import *
```
latture, line 4: I deleted `workflows.py` and moved existing workflows to their own package. Importing the classes here will persevere the existing API and allow us to add new workflows in new files over time instead of accruing code into a single monolithic file (e.g. `predictors.py`).

src/citrine/informatics/predictor_evaluation_result.py
maxhutch: Sure

src/citrine/informatics/predictor_evaluation_metrics.py
```diff
@@ -0,0 +1,138 @@
...
+class CoverageProbability(Serializable["CoverageProbability"], PredictorEvaluationMetric):
+    """[ALPHA] Fraction of observations for which the magnitude of the error is within a
+    confidence interval of a given coverage level.
+
+    Parameters
+    ----------
+    coverage_level: Union[str, float]
+        Confidence-interval coverage level.
+        The coverage level must both be between 0 and 1.0 (non-inclusive) and specified
+        to 3 significant figures.
```
bfolie, line 118: Doesn't it only need to be specified to 3 sig figs if it is passed as a string?

src/citrine/informatics/predictor_evaluation_result.py
```diff
@@ -0,0 +1,257 @@
...
+        Standard error of the mean
+
+    """
+    mean = properties.Float("mean")
+    standard_error = properties.Optional(properties.Float(), "standard_error")
+    typ = properties.String('type', default='RealMetricValue', deserializable=False)
+
+    def __init__(self, *,
+                 mean: float,
+                 standard_error: Optional[float]):
```
bfolie, line 43: Do we distinguish between `None` and 0? That is, if the argument is None then should we set `self.standard_error` to 0?

src/citrine/informatics/predictor_evaluation_result.py
```diff
@@ -0,0 +1,257 @@
...
+    ----------
+    uuid: UUID
+        Unique Citrine id given to the candidate
+    identifiers: Set[str]
+        Set of globally unique identfiers given to the candidate
+    trial: int
+        1-based index of the trial this candidate belonged to
+    fold: int
+        1-based index of the fold this candidate belonged to
+    predicted: RealMetricValue
```
bfolie, line 117: copy paste error

src/citrine/informatics/predictor_evaluation_result.py
```diff
@@ -0,0 +1,257 @@
...
+        Predicted value
+    actual: RealMetricValue
```
bfolie, line 119: copy paste error

src/citrine/informatics/workflows/__init__.py
bfolie: Sounds like we should do the same with predictors. I'll open an issue

src/citrine/informatics/predictor_evaluation_result.py
```diff
@@ -0,0 +1,257 @@
...
+        raise NotImplementedError
+
+
+class CrossValidationResult(Serializable["CrossValidationResult"], PredictorEvaluationResult):
+    """[ALPHA] Result of performing a cross-validation evaluation on a predictor.
+
+    Results for a cross-validated response can be accessed via ``cvResult['response_name']``,
+    where ``cvResult`` is a
+    :class:`citrine.informatics.predictor_evaluation_result.CrossValidationResult`
+    and ``'response_key'`` is a response analyzed by a
```
bfolie, line 223:
```suggestion
    and ``'response_name'`` is a response analyzed by a
```

src/citrine/informatics/predictor_evaluator.py
```diff
@@ -0,0 +1,89 @@
+from typing import Optional, Set, Type
+
+from citrine._serialization import properties
+from citrine._serialization.polymorphic_serializable import PolymorphicSerializable
+from citrine._serialization.serializable import Serializable
+from citrine.informatics.predictor_evaluation_metrics import PredictorEvaluationMetric
+
+
+class PredictorEvaluator(PolymorphicSerializable["PredictorEvaluator"]):
```
bfolie, line 9: Is there a way to enforce that all sub-classes have a `name` field? I ask because that `name` will be used to get the execution's responses

bfolie:
I want to make sure I understand the proposed user workflow. Given a `PredictorEvaluationExecution`, `execution`, the user might call `cv_result = execution.results("5-fold CV temperature grouped together")`, where the string is the name of the `PredictorEvaluator` that was included when creating the `PredictorEvaluationWorkflow`. They then call `cv_foo = cv_sult.results("foo")` to get the results for a specific response and get a `PredictorEvaluationResult`, which right now is always a `CrossValidationResult`. They then call `cv_foo.response_results['ndme']` to get the results for ndme. Is that correct?

src/citrine/informatics/predictor_evaluation_metrics.py
latture: Previously, yes. I just pushed a batch of changes that allows the user to specify the string with < 3 significant figures. If they specify > 3 significant figures (as a float or a string) we will round to 3 decimal places and issue a warning.

src/citrine/informatics/predictor_evaluation_result.py
latture: > None would mean it wasn't computed because there aren't enough samples and 0 would mean that all trials produce the same value

That's correct. Note, we currently set `None` to 0 when we format results on the backend. This is meant to allow us to distinguish between the two scenarios.

src/citrine/informatics/predictor_evaluation_result.py
latture: 🙇 
Good catch.

src/citrine/informatics/predictor_evaluator.py
latture: I'll add an `@property name -> NotImplementedError` method to `PredictorEvaluator`. I think that combined with a descriptive docstring might be close to what you're after. Ideally we'd use an abstract base class or define a Protocol. However the former breaks our serialization logic, and the latter isn't available in all the versions of Python we need to support.

src/citrine/informatics/predictor_evaluation_result.py
```diff
@@ -114,10 +114,12 @@ class CategoricalPredictedVsActual(Serializable["CategoricalPredictedVsActual"],
         1-based index of the trial this candidate belonged to
     fold: int
         1-based index of the fold this candidate belonged to
-    predicted: RealMetricValue
-        Predicted value
-    actual: RealMetricValue
-        Actual value
+    predicted: Mapping[str, float]
+        Predicted class probabilities defined as a map from each class name
+        to its relative probability
```
gregor-robinson, line 119:
```suggestion
        to its relative frequency
```

src/citrine/informatics/predictor_evaluation_result.py
```diff
@@ -114,10 +114,12 @@ class CategoricalPredictedVsActual(Serializable["CategoricalPredictedVsActual"],
...
+    actual: Mapping[str, float]
+        Actual class probabilities defined as a map from each class name
+        to its relative probability
```
gregor-robinson, line 122:
```suggestion
        to its relative frequency
```

maxhutch:
> I want to make sure I understand the proposed user workflow. Given a `PredictorEvaluationExecution`, `execution`, the user might call `cv_result = execution.results("5-fold CV temperature grouped together")`, where the string is the name of the `PredictorEvaluator` that was included when creating the `PredictorEvaluationWorkflow`. They then call `cv_foo = cv_sult.results("foo")` to get the results for a specific response and get a `PredictorEvaluationResult`, which right now is always a `CrossValidationResult`. They then call `cv_foo.response_results['ndme']` to get the results for ndme. Is that correct? Will there always be a `response_results` dictionary?

There are `__getitem__` methods on a few of those classes, so you'd be able to do:
```
execution.results("5-fold CV temperature grouped together")["foo"]["ndme"]
```
I'm going to add a getitem to the execution so we can do:
```
execution.results["5-fold CV temperature grouped together"]["foo"]["ndme"]
```

The interface for `PredictorEvaluationResult` is currently pretty bare, but there is only one concrete class.  When we add additional PredictorEvaluators we can decide what to add to it. Given that we're working in python, I don't think the softer typing is a big deal.

src/citrine/informatics/predictor_evaluation_metrics.py
```diff
@@ -0,0 +1,189 @@
...
+                raise ValueError(
+                    "Invalid coverage level string '{requested_level}'. "
+                    "Coverage level must represent a floating point number between "
+                    "0 and 1 (non-inclusive).".format(
+                        requested_level=coverage_level
+                    ))
+            if _level_float >= 1.0 or _level_float <= 0.0:
+                raise ValueError("Coverage level must be between 0 and 1 (non-inclusive).")
+            _level_float = round(_level_float, 3)
+            if len(coverage_level) > 5:
```
talltea, line 155: This is very minor, but there maybe strings that have more than 3 sig figs that don't trigger this warning, for example: `".1111"`. Perhaps we can compare the pre rounded value to the the post rounded value instead?

src/citrine/informatics/predictor_evaluation_metrics.py
```diff
@@ -0,0 +1,189 @@
...
+                warn(
+                    "Coverage level can only be specified to 3 decimal places."
+                    "Requested level '{requested_level}' will be rounded "
+                    "to {rounded_level}.".format(
+                        requested_level=coverage_level,
+                        rounded_level=_level_float
+                    ))
+        elif isinstance(coverage_level, float):
```
talltea, line 163: It would be nice to collapse this logic with the above where there is overlap (ei, first convert the str to float, and then collapse the float logic into one code block)

src/citrine/informatics/predictor_evaluation_result.py
```diff
@@ -0,0 +1,286 @@
...
+
+    uuid = properties.UUID("uuid")
+    identifiers = properties.Set(properties.String, "identifiers")
+    trial = properties.Integer("trial")
+    fold = properties.Integer("fold")
+    predicted = properties.Object(RealMetricValue, "predicted")
+    actual = properties.Object(RealMetricValue, "actual")
+    typ = properties.String('type', default='RealPredictedVsActual', deserializable=False)
+
+    def __init__(self, *,
```
talltea, line 94: Is the keyword-only arguments pattern something we're going to be applying to all of citrine?

maxhutch: I'd really like to.  I tend to start with it because its easier to remove than to add.

src/citrine/informatics/predictor_evaluation_metrics.py
maxhutch: Adding a test for the warning too.

src/citrine/resources/predictor_evaluation_execution.py
```diff
@@ -0,0 +1,168 @@
...
+                 ):
+        self.project_id: UUID = project_id
+        self.session: Session = session
+        self.workflow_id: Optional[UUID] = workflow_id
+
+    def build(self, data: dict) -> PredictorEvaluationExecution:
+        """Build an individual PredictorEvaluationExecution."""
+        execution = PredictorEvaluationExecution.build(data)
+        execution.session = self.session
+        execution.project_id = self.project_id
```
talltea, line 112: does this need to set the `workflow_id` too?

talltea (approved):
This generally looks good to be!
I'm particularly excited about not using ModuleRef in the trigger method, forcing of the named arguments, and moving parameters that the user isn't meant to set out of the __init__ methods

src/citrine/resources/predictor_evaluation_execution.py
maxhutch: No, since that's returned by the API.

src/citrine/resources/predictor_evaluation_workflow.py
```diff
@@ -0,0 +1,25 @@
...
+    def __init__(self, project_id: UUID, session: Session):
+        self.project_id: UUID = project_id
+        self.session: Session = session
+
+    def build(self, data: dict) -> PredictorEvaluationWorkflow:
+        """Build an individual PredictorEvaluationExecution."""
+        execution = PredictorEvaluationWorkflow.build(data)
+        execution.session = self.session
+        execution.project_id = self.project_id
+        return execution
```
latture, line 25:
```suggestion
        workflow = PredictorEvaluationWorkflow.build(data)
        workflow.session = self.session
        workflow.project_id = self.project_id
        return workflow
```

src/citrine/informatics/workflows/performance_workflow.py
```diff
@@ -0,0 +1,66 @@
...
+from citrine._serialization import properties
+from citrine._session import Session
+from citrine.informatics.analysis_configuration import CrossValidationAnalysisConfiguration
+from citrine.informatics.workflows.workflow import Workflow
+from citrine.resources.workflow_executions import WorkflowExecutionCollection
+
+__all__ = ['PerformanceWorkflow']
+
+
+class PerformanceWorkflow(Resource['PerformanceWorkflow'], Workflow):
```
bfolie, line 14: Should we deprecate this?

tests/informatics/test_predictor_evaluation_metrics.py
```diff
@@ -0,0 +1,60 @@
...
+    (F1(), "f1", "F1 Score"),
+    (AreaUnderROC(), "area_under_roc", "Area Under the ROC"),
+    (CoverageProbability(0.123), "coverage_probability_0.123", "Coverage Probability (0.123)")
+])
+def metric(request):
+    return request.param
+
+
+def test_deser_from_parent(metric):
+    # Serialize and deserialize the descriptors, making sure they are round-trip serializable
```
bfolie, line 21:
```suggestion
    # Serialize and deserialize the metrics, making sure they are round-trip serializable
```

tests/informatics/test_predictor_evaluation_metrics.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def test_string_rep(metric):
+    """String representations of metrics should match our expectation"""
+    assert str(metric[0]) == metric[2]
+    assert repr(metric[0]) == metric[1]
+
+
+def test_to_json(metric):
+    """Make sure we can dump the descriptors to json"""
```
bfolie, line 35:
```suggestion
    """Make sure we can dump the metrics to json"""
```

tests/informatics/workflows/test_predictor_evaluation_workflow.py
```diff
@@ -0,0 +1,31 @@
...
+    pew = PredictorEvaluationWorkflow(
+        name="Test",
+        description="TestWorkflow",
+        evaluators=[evaluator]
+    )
+    return pew
+
+
+def test_dump(pew):
+    assert pew.dump()["name"] == "Test"
```
bfolie, line 19: Wouldn't it be more thorough (and just as easy) to test round-trip de/ser?

src/citrine/informatics/predictor_evaluation_result.py
```diff
@@ -0,0 +1,260 @@
...
+
+    def __iter__(self):
+        return iter(self.value)
+
+    def __getitem__(self, item: int):
+        return self.value[item]
+
+
+class RealPredictedVsActual(Serializable["RealPredictedVsActual"], MetricValue):
+    """[ALPHA] List of predicted vs. actual data points.
```
bfolie, line 136:
```suggestion
    """[ALPHA] List of predicted vs. actual data points for a real value.
```

src/citrine/informatics/predictor_evaluation_result.py
```diff
@@ -0,0 +1,260 @@
...
+    fold = properties.Integer("fold")
+    predicted = properties.Mapping(properties.String, properties.Float, "predicted")
+    actual = properties.Mapping(properties.String, properties.Float, "actual")
+
+    def __init__(self):
+        pass  # pragma: no cover
+
+
+class CategoricalPredictedVsActual(Serializable["CategoricalPredictedVsActual"], MetricValue):
+    """[ALPHA] List of predicted vs. actual data points.
```
bfolie, line 115:
```suggestion
    """[ALPHA] List of predicted vs. actual data points for a categorical value.
```

bfolie (commented):
This is a lot, so I can't be sure I caught everything, but we've been discussing the design for a while and I feel comfortable with that. And the test + code coverage gives me a high degree of confidence.

The only comment remaining that I feel strongly about it deprecating the old performance workflow and related analyses.

Otherwise I could approve this now. It would be preferable to demonstrate a successful build with this branch _and_ the new nextgen tests, but if you're confident that the problems are with the tests themselves and that this won't have to change, then I wouldn't mind approving.

tests/informatics/workflows/test_predictor_evaluation_workflow.py
maxhutch: No, since there are fields that are required when deserializing (like status) that aren't serializable.  Deserialization is covered in `resources/test_predictor_evaulation_workflow.py`.

This testing pattern definitely needs some work, but I think this PR sticks to the style that is in place.

bfolie (approved):

maxhutch:
This is working for me against development.

PR #442, opened 2020-09-12
add create_or_update

jkoeller:
# Citrine Python PR

## Description 

This PR adds the create_or_update functionality for modules and workflows, as described here: https://citrine.atlassian.net/l/c/c6MQ24HD. 

Note that I don't know if this is the right file to put it in.

Thanks to @sdedkins for the initial version of this code implemented in cp-tools here: https://github.com/CitrineInformatics/cp-tools-internal/blob/84908e2b8982787e541146c5752eea8f2485191a/cp_tools/utils.py#L121

Question: wouldn't it be more natural to implement this functionality as a new method in the `Collection` class https://github.com/CitrineInformatics/citrine-python/blob/581d2e2df7815c3d293757cf6f204423b5dfb2ad/src/citrine/_rest/collection.py#L18? We already have a register method and an update method. This new function just calls one of those, and seems natural to be a method of that class rather than its own standalone function.

cc @IAmGrootel @ventura-rivera 

### PR Type:


src/citrine/seeding/find_or_create.py
```diff
@@ -91,3 +93,32 @@ def default_provider():
...
+        Resource that you want to create or update. If collection already contains
+        a resource with the same name it will be updated. Otherwise it will be created.
+    Returns
+    -------
+    ResourceType
+        Registered updated or created resource.
+    """    
+    try:
+        old_resource = get_by_name_or_raise_error(collection, resource.name)
+        print(f"Updating module: {resource.name}")
```
maxhutch, line 116: f-strings aren't supported in python 3.5, which we do support.  I suggest using the `format` method instead.

src/citrine/seeding/find_or_create.py
```diff
@@ -91,3 +93,32 @@ def default_provider():
...
+        # Copy so that passed-in resource is unaffected
+        new_resource = deepcopy(resource)
+        new_resource.uid = old_resource.uid
+        return collection.update(new_resource)
+    except ValueError:
```
maxhutch, line 121: We should narrow this exception: subclass `ValueError`, throw that subclass in `get_by_name_or_raise_error`, and catch only that exception here.

maxhutch: An alternative would be to call the simpler [find collection](https://github.com/CitrineInformatics/citrine-python/blob/master/src/citrine/seeding/find_or_create.py#L6) method and create when it returns `None`.  That's probably better than exception handling (since this isn't an "exceptional" case).

maxhutch (changes requested):
Overall, the interface looks good.  I made a few inline comments about the impl.  The tests are failing on style checks (lots of missing whitespace) and coverage.  You should be able to extend the [fake collection](https://github.com/CitrineInformatics/citrine-python/blob/master/tests/seeding/test_find_or_create.py#L22) with an update method to test.

src/citrine/seeding/find_or_create.py
IAmGrootel: I changed this in the newest commits.

src/citrine/seeding/find_or_create.py
IAmGrootel: Good suggestion with the find_collection. We've adjusted things as you recommended.

IAmGrootel:
@maxhutch this is ready for review!
cc-ing @sdedkins

src/citrine/seeding/find_or_create.py
```diff
@@ -91,3 +97,37 @@ def default_provider():
...
+    Parameters
+    ----------
+    collection : Union[DesignSpaceCollection, PredictorCollection, WorkflowCollection]
+        Collection within which you want to update or create a resource
+    resource : CreationType
+        Resource that you want to create or update.
+
+    Returns
+    -------
+    ResourceType
```
maxhutch, line 120:
```suggestion
    CreationType
```

tests/seeding/test_find_or_create.py
```diff
@@ -280,3 +319,35 @@ def test_find_or_create_dataset_raise_error_exist_multiple(dataset_collection):
...
+                            inputs = [],
+                            outputs = [],
+                            latent_variables = [])
+    create_or_update(predictor_collection, pred)
+    updated_pred = [r for r in list(predictor_collection.list()) if r.name == "resource 4"][0]
+    assert updated_pred.description == "I am updated!"
+
+def test_create_or_update_raise_error_multiple_found(predictor_collection):
+    # test when there are multiple resources that exists with the same listed name and raise error
+    pred = SimpleMLPredictor(name="resource 1",
```
maxhutch, line 347: It took me a second to recognize that `resource 1` was created twice.

tests/seeding/test_find_or_create.py
```diff
@@ -127,6 +130,42 @@ def list(self, page: Optional[int] = None, per_page: int = 100):
...
+        def list(self, page: Optional[int] = None, per_page: int = 100):
+            if page is None:
+                return self.predictors
+            else:
+                return self.predictors[(page - 1)*per_page:page*per_page]
+
+    predictors = SeedingTestPredictorCollection(UUID('6b608f78-e341-422c-8076-35adc8828545'),
+                                session)
+
+    for i in range(0, 5):
```
maxhutch, line 155: The pre-registration logic here could use a comment -- its no obvious what its trying to accomplish or how the test will use it.

maxhutch (changes requested):
The signature and implementation look great.  The fake collection looks good, but it could use a comment or two about the pre-registration logic.

The only blocking issue is that the return value is never checked.  We should check that it matches what you get back when you list.

tests/seeding/test_find_or_create.py
IAmGrootel: added a comment for this for clarity

tests/seeding/test_find_or_create.py
IAmGrootel: @maxhutch Thanks for the speedy review and specific pointers! 
the latest commits should address all your requests.

maxhutch (approved):
LGTM!

PR #448, opened 2020-09-22
Increased the default per_page from 100 to 1000 for project listing and search

crleblanc:
# Citrine Python PR

## Description 

This is a simple PR that increases the default per_page value from 100 to 1000.  Load testing the project listing and search endpoints has shown that a value of 100 yields poor performance when listing a large number of projects (> 40K projects).  Using a value of 1000 does not significantly increase query latency and reduces the number of API calls by a factor of 10.  This should result in much better performance and less load on the accounts DB, roughly 10X faster when paginating a large number of projects.

This PR compliments an optimization to the accounts service that increases the performance by roughly 2X, based on local load testing.

This PR leaves the default collection per_page value to avoid unintended impacts on other services.

I updated the version of Pandas used in test_requirements.txt because version 0.25 was not using binary wheel files and the Docker build was failing after more than 10 minutes.  The current version installs in seconds with a binary wheel.

### PR Type:

### DIFF
--- a/scripts/Dockerfile.pytest
+++ b/scripts/Dockerfile.pytest
@@ -11,5 +11,5 @@ RUN pip install -U -r test_requirements.txt
 COPY . .
 RUN pip install --no-deps -e .
 ENTRYPOINT ["pytest", "--cov=src/", "--cov-report", "term-missing", "--cov-report", "term:skip-covered", \
-    "--cov-fail-under=100", "-s"]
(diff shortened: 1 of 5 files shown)


maxhutch (approved):
LGTM.  Thanks!

maxhutch (approved):

jfriend-citrine (approved):
LGTM

PR #451, opened 2020-09-25
Added an optional Unit type to IngredientQuantityByProcessAndName

dleblanc:
This adds the optional 'unit' to IngredientQuantityByProcessAndName, necessary for a proper filter using absolute quantities.

### PR Type:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,7 @@ class PreBuildCommand(build_py):
 
 
 setup(name='citrine',
-      version='0.71.1',
+      version='0.72.0',
       url='http://github.com/CitrineInformatics/citrine-python',
       description='Python library for the Citrine Platform',
       author='Citrine Informatics',
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,7 @@ class PreBuildCommand(build_py):
 
 
 setup(name='citrine',
-      version='0.71.0',
+      version='0.71.1',
       url='http://github.com/CitrineInformatics/citrine-python',
       description='Python library for the Citrine Platform',
       author='Citrine Informatics',
--- a/src/citrine/gemtables/variables.py
+++ b/src/citrine/gemtables/variables.py
@@ -410,6 +410,7 @@ class IngredientQuantityByProcessAndName(
     type_selector = properties.Enumeration(DataObjectTypeSelector, "type_selector")
     typ = properties.String('type', default="ing_quantity_by_process_and_name",
                             deserializable=False)
+    unit = properties.Optional(properties.String, "unit")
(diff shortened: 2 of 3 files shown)


dleblanc:
Looks like this is necessary to get in so the devkit e2e's can supply a now-mandatory unit to the backend when searching for absolute quantities.

src/citrine/gemtables/variables.py
```diff
@@ -421,13 +422,15 @@ def __init__(self, *,
                  process_template: LinkByUID,
                  ingredient_name: str,
                  quantity_dimension: IngredientQuantityDimension,
-                 type_selector: DataObjectTypeSelector = DataObjectTypeSelector.PREFER_RUN):
+                 type_selector: DataObjectTypeSelector = DataObjectTypeSelector.PREFER_RUN,
+                 unit: str = None):
```
maxhutch, line 426:
```suggestion
                 unit: Optional[str] = None):
```

maxhutch (changes requested):
`unit` should be added to the docstring, with a description of when it needs to be set.  We _could_ even raise a ValueError here in citrine-python if the unit is `None` and the quantity_dimension is ABSOLUTE.

I do not think there is anywhere in the user guide that needs to be update.

I think we should be adding this option to `IngredientQuantityInOutput` too (cc @kroenlein ).

It would be good to set `unit` in one of the unit test cases, probably [this one](https://github.com/CitrineInformatics/citrine-python/blob/master/tests/gemtable/test_variables.py#L20)

setup.py
```diff
@@ -57,7 +57,7 @@ def run(self):
 
 
 setup(name='citrine',
-      version='0.71.0',
+      version='0.71.1',
```
maxhutch, line 60: I think a new option should be a minor version bump.

dleblanc: Yeah good point.

src/citrine/gemtables/variables.py
dleblanc: Updated the docstring, and added the unit to the test case.

With regards to adding the unit to IngredientQuantityInOutput, maybe that should be a separate story, as this is focused on unblocking default table generation.

kroenlein (commented):
It seems a bit inconsistent to implement for IngredientQuantityByProcessAndName but not IngredientQuantityInOutput.

As per our conversation in stand-up, we're carding this up as a second ticket to prioritize, since there's no obvious synergy between the implementations.

https://citrine.atlassian.net/browse/PLA-5736

maxhutch (approved):

PR #453, opened 2020-09-30
Make serde base class behavior explicit with "override"

maxhutch:
# Citrine Python PR

## Description 

The behavior for serializing and deserializing properties that were defined in parent classes was tailor made to support overriding members from `gemd` in `DataConcepts` classes.  This behavior is preserved but put behind a flag, which has been applied to all of those cases.

The new default behavior will serialize and deserialize properties that exist in a parent class the same way as in a concrete class, by default.  This allows us to define the serde properties in parent classes, as demonstrated on predictors.  It also allows for the definition of interfaces with the same name as serialized members (to be demonstrated in #441 ).

### PR Type:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,7 @@ class PreBuildCommand(build_py):
 
 
 setup(name='citrine',
-      version='0.71.0',
+      version='0.71.1',
       url='http://github.com/CitrineInformatics/citrine-python',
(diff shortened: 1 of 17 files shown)


maxhutch:
Resolves #166

latture (approved):

PR #457, opened 2020-10-07
PLA-5803 optional map serialization as list of pairs

talltea:
# Citrine Python PR

## Description 
A quick sketch of what it might look like to optionally serialize maps as lists of pairs

### PR Type:


src/citrine/_serialization/properties.py
```diff
@@ -703,32 +703,53 @@ def __init__(self,
...
+                deserialized[deserialized_key] = deserialized_value
+            return deserialized
+
         for key, value in value.items():
             deserialized_key = self.keys_type.deserialize(key)
             deserialized_value = self.values_type.deserialize(value)
             deserialized[deserialized_key] = deserialized_value
         return deserialized
 
     def _serialize(self, value: dict) -> dict:
```
maxhutch, line 744: The return type here isn't right if `list_of_pairs` is true.  I'm not sure what the right way to fix it is -- probably either a Union or just leaving the type off.

maxhutch (commented):
Looks right.  One minor comment about type hints.

tests/_serialization/test_container_properties.py
```diff
@@ -42,3 +42,19 @@ def test_mapping_property(key_type, value_type, key_value, value_value, key_seri
...
+@pytest.mark.parametrize('key_type,key_value,key_serialized', VALID_SERIALIZATIONS)
+@pytest.mark.parametrize('value_type,value_value,value_serialized', VALID_SERIALIZATIONS)
+def test_mapping_property_list_of_pairs(key_type, value_type, key_value, value_value, key_serialized, value_serialized):
+    prop = properties.Mapping(key_type, value_type, list_of_pairs = True)
+    value = {key_value: value_value}
+    serialized = [(key_serialized, value_serialized),]
+    assert prop.deserialize(serialized) == value
+    assert prop.serialize(value) == serialized
+
+def test_mapping_property_list_of_pairs_multiple():
```
talltea, line 55: going to actually merge these two tests with something like `{key_value: value_value for _ in range(5)}` instead

src/citrine/_serialization/properties.py
bfolie: 👍 for `Union`

tests/_serialization/test_container_properties.py
```diff
@@ -42,3 +42,19 @@ def test_mapping_property(key_type, value_type, key_value, value_value, key_seri
     serialized = {key_serialized: value_serialized}
     assert prop.deserialize(serialized) == value
     assert prop.serialize(value) == serialized
+
+@pytest.mark.parametrize('key_type,key_value,key_serialized', VALID_SERIALIZATIONS)
+@pytest.mark.parametrize('value_type,value_value,value_serialized', VALID_SERIALIZATIONS)
+def test_mapping_property_list_of_pairs(key_type, value_type, key_value, value_value, key_serialized, value_serialized):
+    prop = properties.Mapping(key_type, value_type, list_of_pairs = True)
```
talltea, line 49: woops, forgot to rename this

tests/_serialization/test_container_properties.py
talltea: jk, not that, because keys need to be unique in dictionaries :derp:

setup.py
```diff
@@ -57,7 +57,7 @@ def run(self):
 
 
 setup(name='citrine',
-      version='0.72.3',
+      version='0.73.1',
```
bfolie, line 60: Why the minor/patch version bump?

src/citrine/_serialization/properties.py
```diff
@@ -703,32 +703,53 @@ def __init__(self,
...
+                 override: bool = False,
+                 ser_as_list_of_pairs: bool = False):
         super().__init__(serialization_path,
                          serializable,
                          deserializable,
                          default,
                          override)
         self.keys_type = keys_type if isinstance(keys_type, Property) else keys_type()
         self.values_type = values_type if isinstance(values_type, Property) else values_type()
+        self.ser_as_list_of_pairs = ser_as_list_of_pairs
```
bfolie, line 715: This should be explained somewhere

setup.py
talltea: uhhh... good question... that was a merge failure on my part. I had some code for a different ticket hanging around that shouldn't have gotten in, and got confused when it was in my merge conflict. This should be `0.72.4`

talltea: Fixed, ended up with a minor patch since this introduces a backwards compatible new feature

src/citrine/_serialization/properties.py
talltea: Added as part of the class docstring

src/citrine/_serialization/properties.py
```diff
@@ -696,39 +696,71 @@ def __str__(self):
...
+        if self.ser_as_list_of_pairs:
+            return list
+        else:
+            return dict
 
-    def _deserialize(self, value: dict) -> dict:
+    def _deserialize(self, value: typing.Union[dict, list]) -> dict:
         deserialized = dict()
+
+        if self.ser_as_list_of_pairs:
```
maxhutch, line 742: This logic would result in an error if `self.ser_as_list_of_pairs` and the `value` is a `dict`.  I'm not sure that's the behavior we want: it might be preferable to always be able to deserialize either format and have `ser_as_list_of_pairs` only affect serialization logic.  That would make forward and backward compatibility eaiser.

Also, since list and dict are distinct, we can just do `if type(value) == list`

talltea: yeah, that's what I had originally. I changed it to this to constrain it more specifically so that any errors in serialization were caught earlier. Mostly, I was thinking that if the python client is sending the data serialized as a list and the backend is correctly deserializing that, then the backend should also be sending the data serialized as a list. And if it is not, then something is wrong.

But I think that you are right that it will make backwards compatible changes somewhat easier if we can handle both in this client, and will change it back.

maxhutch: The behavior could also be parameterized, if we want to open the backward compatibility window and close it later to be more strict.

bfolie (approved):

bfolie (approved):

PR #459, opened 2020-10-14
Increasing per_page default for DatasetCollection.list

crleblanc:
# Citrine Python PR

## Description 
This PR modifies the default value of per_page for the list method on DatasetCollection.  The previous default was resulting in a large number of API calls and a heavy load on the accounts DB.  Increasing this to 1000 increases the payload size per API call, greatly decreasing the number of API calls and db queries.  This should result in much better performance and less load on the accounts db.

This change only impacts dataset collections and can be overridden by specifying the per_page value when listing datasets.  Given that these list methods return an iterable, this should be a transparent change for all clients that either iterate over the results or convert this to a Python list.

The same change was made to project collections and has worked well.

### PR Type:

### DIFF
--- a/src/citrine/resources/dataset.py
+++ b/src/citrine/resources/dataset.py
@@ -349,3 +349,30 @@ class DatasetCollection(Collection[Dataset]):
         full_model = self.build(data)
         full_model.project_id = self.project_id
         return full_model
+
+    def list(self,
+             page: Optional[int] = None,
+             per_page: int = 1000) -> Iterable[Dataset]:
+        """
+        List datasets using pagination.
+
+        Leaving page and per_page as default values will yield all elements in the
+        collection, paginating over all available pages.
+
+        Parameters
+        ---------
+        page: int, optional
+            The "page" of results to list. Default is to read all pages and yield
+            all results.  This option is deprecated.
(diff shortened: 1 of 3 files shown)


src/citrine/resources/dataset.py
```diff
@@ -349,3 +349,30 @@ def register(self, model: Dataset) -> Dataset:
         full_model = self.build(data)
         full_model.project_id = self.project_id
         return full_model
+
+    def list(self,
+             page: Optional[int] = None,
+             per_page: int = 1000) -> Iterable[Dataset]:
```
dleblanc, line 355: This is an overriden method?

And we're changing the default per_page to 1000 for everyone? (not just in the tests)?

crleblanc: Yes it's an overridden method and this default applies to everyone using citrine-python.

Applying this to the e2es only would require the sort of solution I mentioned to you earlier, e.g. a keyword arg for find_collection() that would let us pass this value to the various collections.

maxhutch:
The pattern looks fine. I defer technical details to @dleblanc

dleblanc (approved):

dleblanc (approved):

PR #461, opened 2020-10-16
Pla 5804 recursive setter serde

talltea:
# Citrine Python PR

## Description 
WIP fix for Collection properties not correctly handling their member elements
(still missing test coverage)

### PR Type:


src/citrine/informatics/reports.py
```diff
@@ -96,8 +96,8 @@ class ModelSummary(Serializable['ModelSummary']):
 
     name = properties.String('name')
     type_ = properties.String('type')
-    inputs = properties.List(properties.String(), 'inputs')
-    outputs = properties.List(properties.String(), 'outputs')
+    inputs = properties.List(properties.Union([properties.Object(Descriptor), properties.String()]), 'inputs')
```
talltea, line 99: I'd like to make this just `inputs = properties.List(properties.Object(Descriptor), 'inputs')` instead of this union, but doing so breaks a bunch of tests and backwards compatibility.

src/citrine/_serialization/properties.py
```diff
@@ -340,6 +340,26 @@ def _serialize(self, value: list) -> list:
             serialized.append(self.element_type.serialize(element))
         return serialized
 
+    def __set__(self, obj, value: typing.Union[SerializedType, DeserializedType]):
```
talltea, line 343: I added this just for the List Property for now, but this should be generalized for the other collections

tests/_util/test_functions.py
```diff
@@ -31,10 +31,9 @@ def test_get_object_id_from_data_concepts_id_is_none():
         name='test',
         bounds=RealBounds(0.0, 1.0, '')
     )
-    template.uids = {'id': None}
 
     with pytest.raises(ValueError):
-        get_object_id(template)
+        template.uids = {'id': None}
```
talltea, line 36: It makes more sense to me that there would be a ValueError on attempting to set this value to something invalid than waiting till the user attempts to get that value

src/citrine/_serialization/properties.py
```diff
@@ -340,8 +357,17 @@ def _serialize(self, value: list) -> list:
             serialized.append(self.element_type.serialize(element))
         return serialized
 
+    def _set_elements(self, value):
+        elems = []
+        for sub_val in value:
+            if issubclass(type(sub_val), self.element_type.underlying_types):
```
talltea, line 363: I'm not so sure about this if statement

talltea:
@maxhutch is this approximately what you had in mind when we talked about this earlier? And if it is, do you have suggestions on how to get that sweet 100% test coverage?

maxhutch:
@talltea I think so.  Are you having a hard time hitting a specific line?  I'd just create a bunch of test classes and test setter validations and dump/build round-robins.

src/citrine/_serialization/properties.py
```diff
@@ -134,6 +134,23 @@ def __str__(self):
         return '<Property {!r}>'.format(self.serialization_path)
 
 
+class PropertyCollection(Property[DeserializedType, SerializedType]):
+
+    def __set__(self, obj, value: typing.Union[SerializedType, DeserializedType]):
+        """Property setter, deferring to the setter of the parent class, if applicable."""
+        base_class = _get_base_class(obj, self.serialization_path)
+        if issubclass(type(value), self.underlying_types):
+            value_to_set = self._set_elements(value)
```
maxhutch, line 143: This is a forward reference to `_set_elements`.  Can you make `PropertyCollection` abstract (or throw NotImplementedErrors) and define an abstract (or NotImplemented) `_set_elements` with a docstring?

src/citrine/_serialization/properties.py
maxhutch: This seems reasonable, but I'm also a bit worried about the `element_type: typing.Union[Property, typing.Type[Property]]`.  Do we have test coverage of both parts of that union?

src/citrine/_serialization/properties.py
```diff
@@ -340,8 +360,17 @@ def _serialize(self, value: list) -> list:
             serialized.append(self.element_type.serialize(element))
         return serialized
 
+    def _set_elements(self, value):
+        elems = []
+        for sub_val in value:
+            if issubclass(type(sub_val), self.element_type.underlying_types):
+                elems.append(sub_val)
+            else:
+                elems.append(self.element_type.deserialize(sub_val))
```
maxhutch, line 369: This is actually the line I'm more worried about.  I don't think `typing.Type[Property]` has a `deserialize` method.

src/citrine/_serialization/properties.py
maxhutch: I just tried:
```
class DummyDescriptor(Serializable):
    """docstring for dummyDescriptor"""
    dummy_map = properties.Mapping(properties.Float, properties.String, "foo")
    dummy_set = properties.Set(type(properties.Float), "bar")
```
which led to errors.  Maybe the typing bit of those unions is cruft that should be removed.

tests/_serialization/test_container_properties.py
```diff
@@ -53,9 +54,24 @@ def test_mapping_property_list_of_pairs(key_type, value_type, key_value, value_v
...
 
+
 def test_mapping_property_list_of_pairs_multiple():
     prop = properties.Mapping(properties.String, properties.Integer, ser_as_list_of_pairs = True)
     value = {'foo': 1, 'bar': 2}
     serialized = [('foo', 1), ('bar', 2)]
     assert prop.deserialize(serialized) == value
     unittest.TestCase().assertCountEqual(prop.serialize(value), serialized)
+
+class dummyDescriptor(object):
```
maxhutch, line 65:
```suggestion

class DummyDescriptor(object):
```
nit: pep spacing and camel case.

maxhutch (changes requested):
Logic looks good and test coverage is ok.  I'm a bit worried about the typing side of a few unions, but that might be a legacy issue (I'd be willing to merge through it if we'll do it in a follow-up).  The only blocking feedback is to define `_set_elements` abstractly.

src/citrine/_serialization/properties.py
talltea: I don't think we're expecting to handle `properties.Set(type(properties.Float), "bar")`. The three things I can think of us expecting to handle is: 
`properties.Set(type(properties.Float()), "bar")`
`properties.Set(properties.Float, "bar")`
`properties.Set(properties.Float(), "bar")`
And the last two are in use in the code base

I'll extend the test to handle all three, but the code as is handles it, in particular because of `self.element_type = element_type if isinstance(element_type, Property) else element_type()`

src/citrine/_serialization/properties.py
talltea: I went with NotImplemented, but we could switch to abstract if you think it's worth adding that in.

I tend to prefer not using it unless I have to, because it adds some complexity and "dark magic" that I've seen cause issues in the past (I don't think it would cause issues here, but it doesn't seem necessary either)

talltea: woops, looks like python 3.5 and 3.6 didn't like that, gonna fix it and make them happy

talltea: Went with @abstractmethod to match existing implementations

maxhutch (approved):
LGTM

PR #468, opened 2020-10-28
PLA-5878 fix wait_while_executing bug

talltea:
# Citrine Python PR

## Description 
This fixes `wait_while_executing` so that it works both with the old style executions and the new ones. I've tested this against development as seen in the doc update.
I also added mocks for the sleep and time calls so that the unit tests run a little faster, and had to change the imports a little to get that to work.

### PR Type:

### DIFF
--- a/src/citrine/jobs/waiting.py
+++ b/src/citrine/jobs/waiting.py
@@ -1,4 +1,4 @@
-from time import time, sleep
+import time
 from pprint import pprint
 from citrine._rest.collection import Collection
 from citrine.informatics.modules import Module
@@ -6,6 +6,10 @@ from citrine.resources.workflow_executions import (
     WorkflowExecution,
     WorkflowExecutionStatus,
 )
+from citrine.resources.predictor_evaluation_execution import (
+    PredictorEvaluationExecution
+)
+from typing import Union
 
 
 class ConditionTimeoutError(RuntimeError):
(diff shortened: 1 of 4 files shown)


src/citrine/jobs/waiting.py
```diff
@@ -128,16 +133,24 @@ def wait_while_executing(
...
-    while not execution_is_finished() and (time() - start < timeout):
-        sleep(interval)
+        if isinstance(execution, PredictorEvaluationExecution):
+            if collection is None:
+                raise ValueError("Must provide collection")
+            status = collection.get(execution.uid).status
+            if print_status_info:
+                _print_string_status(status, start)
+            return status != "INPROGRESS"
+        else:
```
bfolie, line 146: We have two different patterns for accessing execution status, one for predictor evaluations and one for the other workflows. Are we planning on transitioning everything over to the former type?

talltea: yes, we're hoping to switch over for consistency.
I think the doc that we'd like to more towards is here: https://github.com/CitrineInformatics/ai-engine-documentation/blob/master/docs/status.md, although it looks like that doc has camel case strings, and PEW has all caps, so there's still work to be done.

latture (approved):

bfolie (approved):

PR #473, opened 2020-10-30
Formulation design space and constraints

latture:
This PR adds a `FormulationDesignSpace` and associated ingredient and label constraints:

- `IngredientCountConstraint`
- `IngredientFractionConstraint`
- `LabelFractionConstraint`

I also created submodules for constraints and design spaces (using the pattern we introduced for workflows) to avoid accumulating additional classes into a single file. 

### PR Type:


src/citrine/informatics/design_spaces/formulation_design_space.py
```diff
@@ -0,0 +1,103 @@
...
+    ----------
+    name: str
+        the name of the design space
+    description: str
+        the description of the design space
+    formulation_descriptor: FormulationDescriptor
+        descriptor used to store formulations sampled from the design space
+    ingredients: Set[str]
+        set of ingredient names that can be used in a formulation
+    constraints: Set[IngredientConstraint]
```
maxhutch, line 27: This is where I was thinking of elevating a "max_ingredient_count" member, which would result in an automatic IngredientCountConstraint.

src/citrine/informatics/design_spaces/formulation_design_space.py
```diff
@@ -0,0 +1,103 @@
...
+    experimental_reasons = properties.Optional(
+        properties.List(properties.String()),
+        'experimental_reasons',
+        serializable=False
+    )
+    module_type = properties.String('module_type', default='DESIGN_SPACE', deserializable=False)
+    schema_id = properties.UUID('schema_id', default=UUID('f3907a58-aa46-462c-8837-a5aa9605e79e'),
+                                deserializable=False)
+
+    def __init__(self,
```
maxhutch, line 80:
```suggestion
    def __init__(self, *
```
This API is surely in flux, so let's protect ourselves a bit with changes.

maxhutch (commented):
I think the tests could be far more compact (you can light-up most of the code through a single ser/de loop).  You could look at the classes in PEW.  However, the coverage is there and the design makes sense.

I strongly suggest adding "*" to most of these classes to make the API more flexible (not relying on argument position).

I think adding a "max_ingredient_count" (or similar) option to the FormulationDesignSpace will improve the UX, but that's more of a judgement call.

latture:
> I strongly suggest adding "*" to most of these classes to make the API more flexible (not relying on argument position).

I'll add it to the design space and constraints.

> I think adding a "max_ingredient_count" (or similar) option to the FormulationDesignSpace will improve the UX, but that's more of a judgement call.

We can easily add that in a follow up PR. I think we'll want to inject the constraint on the backend and rearrange some of the validation logic.

latture:
> I think the tests could be far more compact (you can light-up most of the code through a single ser/de loop). You could look at the classes in PEW. However, the coverage is there and the design makes sense.

I removed the FDS init test. We still have 100% coverage without it. I agree could probably get these even more compact, but at least this trims the low hanging fruit.

maxhutch (approved):

src/citrine/informatics/constraints/ingredient_count_constraint.py
```diff
@@ -0,0 +1,49 @@
...
+    formulation_descriptor = properties.Object(FormulationDescriptor, 'formulation_descriptor')
+    min = properties.Integer('min')
+    max = properties.Integer('max')
+    label = properties.Optional(properties.String, 'label')
+    typ = properties.String('type', default='IngredientCountConstraint')
+
+    def __init__(self, *,
+                 formulation_descriptor: FormulationDescriptor,
+                 min: float,
+                 max: float,
```
bfolie, line 39: I just noticed this because I'm using the constraint, but these should be `int`

PR #476, opened 2020-11-06
Raise NotImplementedError when deleting PEWs

maxhutch:
# Citrine Python PR

## Description 
Predictor evaluation workflows and executions cannot be deleted;
rather they are archived (and unarchived).  This raises a
NotImplementedError instead of trying to delete them and getting
a 404 since there is no such route in the API.

### PR Type:

### DIFF
--- a/src/citrine/resources/predictor_evaluation_execution.py
+++ b/src/citrine/resources/predictor_evaluation_execution.py
@@ -1,6 +1,6 @@
 """Resources that represent both individual and collections of workflow executions."""
 from functools import lru_cache, partial
-from typing import Optional, Iterable
+from typing import Optional, Iterable, Union
 from uuid import UUID
 
 from citrine._rest.collection import Collection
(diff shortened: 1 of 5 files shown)


maxhutch:
another option would be to remove the `delete` method from the base `Collection`, create a new class called `Deleteable` that contains that method, and extend that method in a bunch of collections instead.  An approach like that feels preferable but it might make more sense to refactor our Collection base class holistically.

src/citrine/resources/predictor_evaluation_execution.py
```diff
@@ -195,3 +196,8 @@ def list(self,
                                         collection_builder=self._build_collection_elements,
                                         page=page,
                                         per_page=per_page)
+
+    def delete(self, uid: Union[UUID, str]) -> Response:
+        """Predictor Evaluation Workflows cannot be deleted; they can be archived instead."""
```
latture, line 201:
```suggestion
        """Predictor Evaluation Executions cannot be deleted; they can be archived instead."""
```

src/citrine/resources/predictor_evaluation_execution.py
```diff
@@ -195,3 +196,8 @@ def list(self,
...
+        raise NotImplementedError(
+            "Predictor Evaluation Workflows cannot be deleted; they can be archived instead.")
```
latture, line 203:
```suggestion
            "Predictor Evaluation Executions cannot be deleted; they can be archived instead.")
```

latture (dismissed):

latture (approved):

PR #479, opened 2020-11-10
Feature/pla 4229 narrow template bounds

dleblanc:
This adds the support for updating with data validation to data objects (primarily templates).

### PR Type:

### DIFF
--- a/src/citrine/_rest/collection.py
+++ b/src/citrine/_rest/collection.py
@@ -7,7 +7,8 @@ from uuid import UUID
 from time import time, sleep
 
 from citrine._rest.paginator import Paginator
-from citrine.exceptions import ModuleRegistrationFailedException, NonRetryableException
+from citrine.exceptions import ModuleRegistrationFailedException, NonRetryableException, \
+    PollingTimeoutError, JobFailureError
 from citrine.resources.job import JobSubmissionResponse, JobStatusResponse
 from citrine.resources.response import Response
 
@@ -275,13 +276,17 @@ class Collection(Generic[ResourceType]):
             else:
                 logger.error('Job exceeded user timeout of {} seconds.'.format(timeout))
                 logger.debug('Last status: {}'.format(status.dump()))
-                raise TimeoutError('Job {} timed out.'.format(job_id))
+                raise PollingTimeoutError('Job {} timed out.'.format(job_id))
         if status.status == 'Failure':
             logger.debug('Job terminated with Failure status: {}'.format(status.dump()))
+            failure_reasons = []
             for task in status.tasks:
                 if task.status == 'Failure':
                     logger.error('Task {} failed with reason "{}"'.format(
                         task.id, task.failure_reason))
-            raise RuntimeError('Job {} terminated with Failure status.'.format(job_id))
+                    failure_reasons.append(task.failure_reason)
+            raise JobFailureError(
+                'Job {} terminated with Failure status. Failure reasons: {}'.format(
+                    job_id, failure_reasons), job_id, failure_reasons)
 
         return status
--- a/src/citrine/_rest/collection.py
+++ b/src/citrine/_rest/collection.py
@@ -248,6 +248,8 @@ class Collection(Generic[ResourceType]):
             Amount of time to wait on the job (in seconds) before giving up. Defaults
             to 2 minutes. Note that this number has no effect on the underlying job
             itself, which can also time out server-side.
+        polling_delay:
+            How long to delay between each polling retry attempt.
 
         Returns
         -------
--- a/src/citrine/_rest/collection.py
+++ b/src/citrine/_rest/collection.py
@@ -288,7 +288,8 @@ class Collection(Generic[ResourceType]):
                         task.id, task.failure_reason))
                     failure_reasons.append(task.failure_reason)
             raise JobFailureError(
-                'Job {} terminated with Failure status. Failure reasons: {}'.format(
-                    job_id, failure_reasons), job_id, failure_reasons)
+                message='Job {} terminated with Failure status. Failure reasons: {}'.format(
+                    job_id, failure_reasons), job_id=job_id,
+                failure_reasons=failure_reasons)
 
         return status
--- a/src/citrine/_rest/collection.py
+++ b/src/citrine/_rest/collection.py
@@ -229,8 +229,8 @@ class Collection(Generic[ResourceType]):
             )
             warnings.warn(msg)
 
-    def _poll_for_job_completion(self, project_id: UUID,
-                                 job: Union[JobSubmissionResponse, UUID], *,
+    def _poll_for_job_completion(self, project_id: Union[UUID, str],
+                                 job: Union[JobSubmissionResponse, UUID, str], *,
                                  timeout: float = 2 * 60,
                                  polling_delay: float = 2.0) -> JobStatusResponse:
         """
--- a/src/citrine/_rest/collection.py
+++ b/src/citrine/_rest/collection.py
@@ -1,11 +1,17 @@
 import warnings
 from abc import abstractmethod
+from logging import getLogger
 from typing import Optional, Union, Generic, TypeVar, Iterable, Dict, Tuple, Callable
 from uuid import UUID
 
+from time import time, sleep
+
 from citrine._rest.paginator import Paginator
 from citrine.exceptions import ModuleRegistrationFailedException, NonRetryableException
+from citrine.resources.job import JobSubmissionResponse, JobStatusResponse
 from citrine.resources.response import Response
+
+logger = getLogger(__name__)
 
 ResourceType = TypeVar('ResourceType', bound='Resource')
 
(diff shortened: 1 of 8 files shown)


src/citrine/_rest/collection.py
```diff
@@ -222,3 +228,60 @@ def _check_experimental(self, data):
...
+                break
+            elif time() - start_time < timeout:
+                logger.info('Job still in progress, polling status again in {:.2f} seconds.'
+                            .format(polling_delay))
+
+                sleep(polling_delay)
+            else:
+                logger.error('Job exceeded user timeout of {} seconds.'.format(timeout))
+                logger.debug('Last status: {}'.format(status.dump()))
+                raise TimeoutError('Job {} timed out.'.format(job_id))
```
maxhutch, line 278: Could we raise a custom exception here (derived from `TimeoutError`), e.g. [these](https://github.com/CitrineInformatics/citrine-python/blob/master/src/citrine/exceptions.py).

src/citrine/_rest/collection.py
```diff
@@ -222,3 +228,60 @@ def _check_experimental(self, data):
...
+        if status.status == 'Failure':
+            logger.debug('Job terminated with Failure status: {}'.format(status.dump()))
+            for task in status.tasks:
+                if task.status == 'Failure':
+                    logger.error('Task {} failed with reason "{}"'.format(
+                        task.id, task.failure_reason))
+            raise RuntimeError('Job {} terminated with Failure status.'.format(job_id))
```
maxhutch, line 285: Could we raise a custom exception here (derived from `RuntimeError`), e.g. [these](https://github.com/CitrineInformatics/citrine-python/blob/master/src/citrine/exceptions.py).

maxhutch: Should this message include the `task.failure_reason`?

src/citrine/resources/data_concepts.py
```diff
@@ -388,6 +388,68 @@ def register_all(self, models: List[ResourceType], dry_run=False) -> List[Resour
         )
         return [self.build(obj) for obj in response_data['objects']]
 
+    def update_with_data_validation(self, model: ResourceType,
+                                    dry_run: Optional[bool] = False,
+                                    wait_for_response: Optional[bool] = True,
+                                    timeout: float = 2 * 60,
+                                    polling_delay: float = 1.0) -> Union[None, UUID]:
```
maxhutch, line 395: `Union[None, UUID]` is the same as `Optional[UUID]`, right?

src/citrine/resources/data_concepts.py
```diff
@@ -388,6 +388,68 @@ def register_all(self, models: List[ResourceType], dry_run=False) -> List[Resour
...
+        ResourceType
+            If wait_for_response if True, then this call will poll the backend, waiting
+            for the eventual job result. In the case of successful validation/update,
+            a return value of None is provided which indicates success. In the case of
+            a failure validating or processing the update, an exception (RuntimeError)
+            is raised and an error message is logged with the underlying reason of the
+            failure.
+
+            If wait_for_response if False, A job ID (of type UUID) is returned that one
+            can use to poll for the job completion and result.
```
maxhutch, line 423: What public method can be used to poll the job?

maxhutch (commented):
Some comments inline.

src/citrine/_rest/collection.py
```diff
@@ -222,3 +228,60 @@ def _check_experimental(self, data):
                 "\n  ".join(data.get("experimental_reasons") or ["Unknown reason"])
             )
             warnings.warn(msg)
+
+    def _poll_for_job_completion(self, project_id: UUID,
+                                 job: Union[JobSubmissionResponse, UUID], *,
```
kroenlein, line 233:
```suggestion
    def _poll_for_job_completion(self, project_id: Union[UUID, str],
                                 job: Union[JobSubmissionResponse, UUID, str], *,
```
Matching other signatures around UUIDs in file.

src/citrine/resources/data_concepts.py
dleblanc: Oh I didn't realize that, I guess that's simpler.

src/citrine/_rest/collection.py
dleblanc: A custom exception can work, and I think including the task failure reason is appropriate.

src/citrine/_rest/collection.py
dleblanc: Done - we now raise a custom PollingTimeoutError.

src/citrine/_rest/collection.py
dleblanc: I've also included that info (and the job ID).

src/citrine/resources/data_concepts.py
dleblanc: Fixed

src/citrine/resources/data_concepts.py
dleblanc: I've introduced a "poll_update_with_data_validation_job" method - which exposes separately the logic I had buried under the 'wait' flag.

src/citrine/exceptions.py
```diff
@@ -84,6 +85,21 @@ class WorkflowNotReadyException(RetryableException):
     pass
 
 
+class PollingTimeoutError(RetryableException):
```
maxhutch, line 88: Is this retryable?  Seems like if the polling times out the user should not simply repeat the request.

src/citrine/exceptions.py
```diff
@@ -84,6 +85,21 @@ class WorkflowNotReadyException(RetryableException):
...
+    """Polling for an asynchronous result has exceeded the timeout."""
+
+    pass
+
+
+class JobFailureError(RetryableException):
```
maxhutch, line 94: This definitely isn't retryable; once the job failed it will stay failed.

src/citrine/resources/data_concepts.py
```diff
@@ -388,6 +388,102 @@ def register_all(self, models: List[ResourceType], dry_run=False) -> List[Resour
         )
         return [self.build(obj) for obj in response_data['objects']]
 
+    def async_update(self, model: ResourceType,
+                     dry_run: Optional[bool] = False,
```
maxhutch, line 392:
```suggestion
                     dry_run: bool = False,
```
This should never actually be None.

src/citrine/resources/data_concepts.py
```diff
@@ -388,6 +388,102 @@ def register_all(self, models: List[ResourceType], dry_run=False) -> List[Resour
...
+                     wait_for_response: Optional[bool] = True,
```
maxhutch, line 393:
```suggestion
                     wait_for_response: bool = True,
```
This should never actually be None.

src/citrine/resources/data_concepts.py
```diff
@@ -388,6 +388,102 @@ def register_all(self, models: List[ResourceType], dry_run=False) -> List[Resour
...
+            Whether to actually update the item or run a dry run of the update operation.
+            Dry run is intended to be used for validation. Default: false
+
+        Returns
+        -------
+        Optional[UUID]
+            If wait_for_response if True, then this call will poll the backend, waiting
+            for the eventual job result. In the case of successful validation/update,
+            a return value of None is provided which indicates success. In the case of
+            a failure validating or processing the update, an exception (RuntimeError)
```
maxhutch, line 418:
```suggestion
            a failure validating or processing the update, an exception (JobFailureError)
```

src/citrine/resources/data_concepts.py
```diff
@@ -388,6 +388,102 @@ def register_all(self, models: List[ResourceType], dry_run=False) -> List[Resour
...
+        to templates.
+
+        Parameters
+        ----------
+        model: ResourceType
+            The DataConcepts object.
+        dry_run: bool
+            Whether to actually update the item or run a dry run of the update operation.
+            Dry run is intended to be used for validation. Default: false
+
```
maxhutch, line 411: wait_for_response, timeout, and polling_delay need parameter docs

src/citrine/resources/data_concepts.py
```diff
@@ -388,6 +388,102 @@ def register_all(self, models: List[ResourceType], dry_run=False) -> List[Resour
...
+        waiting for the eventual job result. In the case of successful validation/update,
+        a return value of None is provided which indicates success. In the case of
+        a failure validating or processing the update, an exception (RuntimeError)
+        is raised and an error message is logged with the underlying reason of the
+        failure.
+
+        Parameters
+        ----------
+        job_id: UUID
+           The job ID for the asynchronous update job we wish to poll.
```
maxhutch, line 471: this needs parameter docstrings for timeout and polling delay.

src/citrine/resources/data_concepts.py
```diff
@@ -388,6 +388,102 @@ def register_all(self, models: List[ResourceType], dry_run=False) -> List[Resour
...
+    def poll_async_update_job(self, job_id: UUID, timeout: float = 2 * 60,
+                              polling_delay: float = 1.0) -> None:
+        """
+        [ALPHA] Poll for the result of the async_update call.
+
+        This call will poll the backend given the Job ID that came from a call to
+        :func:`~citrine.resources.DataConceptsCollection.async_update`,
+        waiting for the eventual job result. In the case of successful validation/update,
+        a return value of None is provided which indicates success. In the case of
+        a failure validating or processing the update, an exception (RuntimeError)
```
maxhutch, line 464:
```suggestion
        a failure validating or processing the update, an exception (JobFailureError)
```

maxhutch (changes requested):
A couple small requests, but generally looking good.

src/citrine/resources/data_concepts.py
```diff
@@ -388,6 +388,102 @@ def register_all(self, models: List[ResourceType], dry_run=False) -> List[Resour
         )
         return [self.build(obj) for obj in response_data['objects']]
 
+    def async_update(self, model: ResourceType,
```
kroenlein, line 391: Max 2 positional arguments.  Perhaps:
```suggestion
    def async_update(self, model: ResourceType, *, 
```

src/citrine/resources/data_concepts.py
```diff
@@ -388,6 +388,102 @@ def register_all(self, models: List[ResourceType], dry_run=False) -> List[Resour
...
+        if wait_for_response:
+            self.poll_async_update_job(job_id, timeout=timeout,
+                                       polling_delay=polling_delay)
+
+            # That worked, nothing returned in this case
+            return None
+        else:
+            return job_id
+
+    def poll_async_update_job(self, job_id: UUID, timeout: float = 2 * 60,
```
kroenlein, line 455: Max 2 positional arguments.  Perhaps:
```suggestion
    def poll_async_update_job(self, job_id: UUID, *, timeout: float = 2 * 60,
```

src/citrine/exceptions.py
dleblanc: True on that one.

src/citrine/exceptions.py
dleblanc: I suppose one could retry the poll if they wanted to wait longer. I'm happy to make it a non-retryable exception.

src/citrine/resources/data_concepts.py
dleblanc: done

src/citrine/resources/data_concepts.py
dleblanc: done

src/citrine/exceptions.py
dleblanc: Done.

src/citrine/exceptions.py
dleblanc: Done.

src/citrine/exceptions.py
```diff
@@ -85,16 +85,16 @@ class WorkflowNotReadyException(RetryableException):
...
 
     pass
 
 
-class JobFailureError(RetryableException):
+class JobFailureError(NonRetryableException):
     """The asynchronous job completed with the given failure message."""
 
-    def __init__(self, message: str, job_id: UUID, failure_reasons: List[str]):
+    def __init__(self, message: str, *, job_id: UUID, failure_reasons: List[str]):
```
kroenlein, line 97: Change of suggestion following a conversation with @maxhutch 
```suggestion
    def __init__(self, *, message: str, job_id: UUID, failure_reasons: List[str]):
```

src/citrine/exceptions.py
```diff
@@ -84,6 +85,21 @@ class WorkflowNotReadyException(RetryableException):
...
+class PollingTimeoutError(NonRetryableException):
+    """Polling for an asynchronous result has exceeded the timeout."""
+
+    pass
+
+
+class JobFailureError(NonRetryableException):
+    """The asynchronous job completed with the given failure message."""
+
+    def __init__(self, message: str, *, job_id: UUID, failure_reasons: List[str]):
```
maxhutch, line 97: This doesn't make a ton of sense.  If we have 3 required arguments, then I think they should all be keyword only.

Also, I think this is the source of the test failures.  job_id and failure_reasons are keyword only but the constructor invocation [here](https://github.com/CitrineInformatics/citrine-python/pull/479/files#diff-f6170f32137e1e86f98504d21a90b363c2e46a166001311cdd69a497d54610beR290) isn't using keywords

dleblanc: Agreed. Remove that.

kroenlein (approved):

maxhutch (approved):

PR #486, opened 2020-12-02
Don't serialize label as list of pairs.

latture:
Removed flag that causes labels to be serialized as a list of pairs.


bfolie (dismissed):

latture:
@bfolie The version number got bumped on master while the PR tests were running. Re-requesting review.

bfolie (approved):

PR #487, opened 2020-12-02
Migration utilities for enumerated -> data source design spaces

maxhutch:
# Citrine Python PR

## Description 

There are two utilities here:
 - enumerated_to_data_source to convert a local copy of an enumerated
   design space to a data source design space, including the file upload
 - migrate_enumerated_design_space migrates an on-platform enumerated
   design space to an on-platform data source design space

### PR Type:

### DIFF
--- a/src/citrine/builders/design_spaces.py
+++ b/src/citrine/builders/design_spaces.py
@@ -290,7 +290,7 @@ def migrate_enumerated_design_space(*,
                                     cleanup: bool = True
                                     ) -> DataSourceDesignSpace:
     """
-    Migrate an EnumeratedDesignSpace on the Citrine Platform to a DataSourceDesignSpace.
+    [ALPHA] Migrate an EnumeratedDesignSpace on the Citrine Platform to a DataSourceDesignSpace.
 
     Parameters
     ----------
--- a/src/citrine/builders/design_spaces.py
+++ b/src/citrine/builders/design_spaces.py
@@ -307,13 +307,13 @@ def migrate_enumerated_design_space(*,
         Optional string name to specify for the data CSV file.
         Defaults to the design space name + "source data"
     cleanup: bool
-        Whether or not to archive the migrated dataset if the migration is successful
+        Whether or not to archive the migrated design space if the migration is successful
         Default: true
 
     Returns
     -------
     DataSourceDesignSpace
-        The resulting dataset, which should have functional parity with the source design space.
+        The resulting design space, which should have functional parity with the source design space.
 
     """
     enumerated_ds = project.design_spaces.get(uid)
--- a/src/citrine/builders/design_spaces.py
+++ b/src/citrine/builders/design_spaces.py
@@ -1,3 +1,12 @@
+import csv
+from uuid import UUID
+
+from citrine.informatics.data_sources import CSVDataSource
+from citrine.resources.dataset import Dataset
+from citrine.resources.project import Project
+from citrine.seeding import find_or_create
+from citrine.seeding.find_or_create import find_or_create_dataset
+
 try:
     import pandas as pd
 except ImportError:  # pragma: no cover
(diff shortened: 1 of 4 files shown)


src/citrine/builders/design_spaces.py
```diff
@@ -235,3 +244,93 @@ def cartesian_join_design_spaces(
...
+    uid: UUID
+        Unique identifier of the EnumeratedDesignSpace to migrate
+    dataset: Union[str, Dataset]
+        Dataset or name of dataset into which to write the data as a CSV.
+        If given as a string, will find_or_create using that name.
+    filename: Optional[str]
+        Optional string name to specify for the data CSV file.
+        Defaults to the design space name + "source data"
+    cleanup: bool
+        Whether or not to archive the migrated dataset if the migration is successful
```
bfolie, line 310:
```suggestion
        Whether or not to archive the migrated design space if the migration is successful
```

src/citrine/builders/design_spaces.py
```diff
@@ -235,3 +244,93 @@ def cartesian_join_design_spaces(
...
+        Default: true
+
+    Returns
+    -------
+    DataSourceDesignSpace
+        The resulting dataset, which should have functional parity with the source design space.
```
bfolie, line 316:
```suggestion
        The resulting design space, which should have functional parity with the source design space.
```

src/citrine/builders/design_spaces.py
```diff
@@ -235,3 +244,93 @@ def cartesian_join_design_spaces(
...
+
+    """
+    enumerated_ds = project.design_spaces.get(uid)
+    if not isinstance(enumerated_ds, EnumeratedDesignSpace):
+        msg = "Trying to migrate an enumerated design space but this is a {}".format(type(enumerated_ds))
+        raise ValueError(msg)
+
+    if isinstance(dataset, str):
+        dataset = find_or_create_dataset(project.datasets, dataset)
```
bfolie, line 325: Why is this "create dataset" functionality included? Would we not expect the user to have a ready dataset in the common case?

src/citrine/builders/design_spaces.py
```diff
@@ -235,3 +244,93 @@ def cartesian_join_design_spaces(
...
+
+    # create the new data source design space
+    data_source_ds = enumerated_to_data_source(enumerated_ds, dataset, filename)
+    data_source_ds = project.design_spaces.register(data_source_ds)
+
+    if cleanup:
+        # archive the old enumerated design space
```
bfolie, line 332: This looks like it archives the new design space

src/citrine/builders/design_spaces.py
```diff
@@ -235,3 +244,93 @@ def cartesian_join_design_spaces(
...
+    enumerated_ds = project.design_spaces.get(uid)
+    if not isinstance(enumerated_ds, EnumeratedDesignSpace):
+        msg = "Trying to migrate an enumerated design space but this is a {}".format(type(enumerated_ds))
+        raise ValueError(msg)
+
+    if isinstance(dataset, str):
+        dataset = find_or_create_dataset(project.datasets, dataset)
+
+    # create the new data source design space
+    data_source_ds = enumerated_to_data_source(enumerated_ds, dataset, filename)
```
talltea, line 328: This should be `enumerated_to_data_source(enumerated_ds=enumerated_ds, dataset=dataset, filename=filename)`

src/citrine/builders/design_spaces.py
talltea: oh, that would explain why it wasn't showing up in the ui list for me ➕

talltea:
I tested this out (locally against development) and was able to use the helper functions to migrate a cake design space except for the inline comments.

src/citrine/builders/design_spaces.py
maxhutch: To make it even easier to use.  I agree that creating datasets just for this purpose is probably not the right behavior, though.  How about leaving the string option in and looking up the dataset by name if the string is given, but not creating the dataset if it doesn't already exist?

src/citrine/builders/design_spaces.py
maxhutch: nice catch

src/citrine/builders/design_spaces.py
bfolie: These are utilities that are meant to benefit the user, so my inclination is to do the minimal thing that we know is useful. In this case, that would be to make the user specify the dataset itself, since the utility here is coming from the migration. If we find out that users want to supply a string and have it look up or create a dataset of that name, then we can add that in later.

bfolie (approved):

PR #490, opened 2020-12-04
Modified unit to be required, consistent with backend

kroenlein:
# Citrine Python PR

## Description 
PLA-6124

DS ran into an issue because backend now requires a unit argument for IngredientQuantityDimension.ABSOLUTE.  Based on user feedback, this migrated from previous behavior, where unit was not required to build a table.

The original platform behavior was likely informed by not being concerned about unit consistency from a display perspective.  However, in order to train a predictor, original unit is required.

IngredientQuantityInOutput does not have a unit requirement.  We therefore are now in a state where two very similar methods have inconsistent signatures and possibly inconsistent behavior on the back end.

This PR is intended to quickly address the present pitfall, with documentation for the broader issue logged in the ticket.

### PR Type:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -62,7 +62,7 @@ class PreBuildCommand(build_py):
 
 
 setup(name='citrine',
-      version='0.84.0',
+      version='0.85.0',
       url='http://github.com/CitrineInformatics/citrine-python',
       description='Python library for the Citrine Platform',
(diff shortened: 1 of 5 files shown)


maxhutch (approved):
LGTM

maxhutch (approved):

PR #495, opened 2020-12-15
Made version number a module variable.

latture:
This PR moves the version number to `src/citrine/__version__.py`, so it can be shared with user docs. This pattern was modeled after a similar solution used by the `requests` package.

### PR Type:


src/citrine/__version__.py
```diff
@@ -0,0 +1 @@
+__version__ = '0.85.3'
```
latture, line 1: Link to equivalent file in `requests`: https://github.com/psf/requests/blob/master/requests/__version__.py

They put more information in that file that just the version number. We could do something similar in the future, but this provides a pattern we could expand on in the future.

setup.py
```diff
@@ -60,9 +60,13 @@ def run(self):
         strip_all_hints(STRIP_DIRS)
         build_py.run(self)
 
+about = {}
+with open(join(this_directory, 'src', 'citrine', '__version__.py'), 'r', 'utf-8') as f:
+    exec(f.read(), about)
```
latture, line 65: Very similar pattern to https://github.com/psf/requests/blob/master/setup.py#L60

docs/source/conf.py
```diff
@@ -14,16 +14,19 @@
...
 copyright = '2019, Citrine Informatics'
 author = 'Citrine Informatics'
 
-# The full version, including alpha/beta/rc tags
-release = '0.0.1'
-
+# The short X.Y version.
+version = citrine.__version__
+# The full version, including alpha/beta/rc tags.
+release = citrine.__version__
```
latture, line 29: https://github.com/psf/requests/blob/master/docs/conf.py#L68

setup.py
```diff
@@ -60,9 +60,13 @@ def run(self):
         strip_all_hints(STRIP_DIRS)
         build_py.run(self)
 
+about = {}
+with open(join(this_directory, 'src', 'citrine', '__version__.py'), 'r') as f:
+    exec(f.read(), about)
```
latture, line 65: Very similar pattern to https://github.com/psf/requests/blob/master/setup.py#L60

latture:
@maxhutch This isn't something we have to do, but this is one way we could provide a shared version number.

src/citrine/__init__.py
```diff
@@ -1,5 +1,6 @@
 # TODO: Add a docstring here
 from citrine.citrine import Citrine  # noqa: F401
 import logging
+from .__version__ import __version__
```
talltea, line 4: Build failures are coming from this import being unused

talltea (commented):
This looks good to me. I think we also need to change the script that checks version bumps (https://github.com/CitrineInformatics/citrine-python/blob/master/scripts/validate-version-bump.sh) to point to `src/citrine/__version__.py`

src/citrine/__init__.py
latture: 👍

latture:
> This looks good to me. I think we also need to change the script that checks version bumps (https://github.com/CitrineInformatics/citrine-python/blob/master/scripts/validate-version-bump.sh) to point to `src/citrine/__version__.py`

You're right. (As someone who's terrible at writing regular expressions) do you know what that regex should be?

talltea (dismissed):
This looks good to me 👍  (as long as we can merge in the follow up pr changing the version bump script to point to the correct file in master)

maxhutch (dismissed):
LGTM

latture:
This needed a version bump. Re-requesting review. Also, the follow up PR is queued up here: #496

maxhutch (approved):

PR #496, opened 2020-12-16
Check __version__.py for previous version number

latture:
This PR points our `validate-version-bump.py` at `__version__.py` to check the previous version number instead of `setup.py`.


talltea (approved):
👍

PR #497, opened 2020-12-17
Product design space with subspaces

bfolie:
# Citrine Python PR

## Description 
PLA-6027
Adds `subspaces` argument to `ProductDesignSpace` and updates the type hint from `Univariate` to `ProductDesignSpace`. `subspaces` is a mixed list of design spaces (embedded) and uuids (referring to on-platform design spaces). I have verified the following expected behavior against a live environment:
* Old product design spaces (no subspaces, `type=Univariate`) are cleanly deserialized into the new format with the new type hint and `subspaces=[]`. If the design space is updated and pushed, it will overwrite the original (including changing its type hint from `Univariate` to `ProductDesignSpace`)
* I can register a product design space with dimensions, an on-platform subspace, and an embedded subspace
* When getting this product design space, the subspaces both come back hydrated, but one has a uid and one does not
* When serializing this hydrated product design space, the subspace with a uid gets serialized as its uid while the embedded subspace gets serialized in its entirety. This means that if I modify both subspaces and try to update, only the changes to the embedded subspace will stick.

### PR Type:


src/citrine/informatics/design_spaces/design_space.py
```diff
@@ -12,11 +13,28 @@ class DesignSpace(Module):
 
     Abstract type that returns the proper type given a serialized dict.
 
-
     """
 
     _response_key = None
 
+    uid = properties.Optional(properties.UUID, 'id', serializable=False)
```
bfolie, line 20: consolidating some deser logic

src/citrine/informatics/design_spaces/product_design_space.py
```diff
@@ -26,41 +27,55 @@ class ProductDesignSpace(Resource['ProductDesignSpace'], DesignSpace):
...
+                 dimensions: Optional[List[Dimension]] = None,
                  session: Session = Session()):
         self.name: str = name
         self.description: str = description
-        self.dimensions: List[Dimension] = dimensions
+        self.subspaces: List[Union[UUID, DesignSpace]] = subspaces or []
+        self.dimensions: List[Dimension] = dimensions or []
         self.session: Session = session
 
+    def dump(self) -> dict:
```
bfolie, line 56: There's essentially a pre-dump step here. It might make sense to add `pre_dump` as a standard method, rather than completely overriding `dump()` as I do here.

tests/conftest.py
```diff
@@ -15,9 +17,45 @@ def valid_product_design_space_data():
...
             name='my design space',
             description='does some things',
+            subspaces=[
+                dict(
+                    module_type='DESIGN_SPACE',
+                    status='READY',
+                    id=str(uuid.uuid4()),
+                    archived=False,
+                    name='first subspace',
+                    instance=dict(
```
bfolie, line 30: The subspace modules come back inside an `instance` field instead of a `config` field. I don't know why.

src/citrine/informatics/design_spaces/product_design_space.py
```diff
@@ -11,56 +12,75 @@
...
-        'experimental_reasons',
-        serializable=False
+    subspaces = properties.List(properties.Union(
+        [properties.UUID, properties.Object(DesignSpace)]
+    ), 'config.subspaces', default=[])
+    dimensions = properties.Optional(
+        properties.List(properties.Object(Dimension)), 'config.dimensions'
     )
+    typ = properties.String('config.type', default='ProductDesignSpace', deserializable=False)
+    # Product design spaces may not be embedded in other subspaces, hence status is required
```
maxhutch, line 42: Is this enforced by the backend?

src/citrine/informatics/design_spaces/product_design_space.py
```diff
@@ -11,56 +12,75 @@
...
+    status = properties.String('status', serializable=False)
 
     # NOTE: These could go here or in _post_dump - it's unclear which is better right now
     module_type = properties.String('module_type', default='DESIGN_SPACE')
     schema_id = properties.UUID('schema_id', default=UUID('6c16d694-d015-42a7-b462-8ef299473c9a'))
 
-    def __init__(self,
+    def __init__(self, *,
```
maxhutch, line 49: Note that this is a breaking change.  I like it (and this is an alpha route), but we'll want to be sure to message it to users.

src/citrine/informatics/design_spaces/product_design_space.py
maxhutch: We could do that.  Non-blocking, though.

src/citrine/informatics/design_spaces/product_design_space.py
```diff
@@ -11,56 +12,75 @@
...
+        for i, subspace in enumerate(data['config']['subspaces']):
+            if isinstance(subspace, dict):
+                # embedded design spaces are not modules, so only serialize their config
+                data['config']['subspaces'][i] = subspace['config']
+        return data
+
+    @classmethod
+    def _pre_build(cls, data: dict) -> dict:
+        subspaces = data['config'].get('subspaces', [])
+        for i, _ in enumerate(subspaces):
```
maxhutch, line 81: This is just propagating the instance -> config rename deeper into the tree, but not changing the shape, right?  If so, that's probably worth a comment.

maxhutch (commented):
This looks great!  A few minor comments, and there's a merge conflict (I removed schema_id last night).  If this is working with our e2es, then I think we're good.

src/citrine/informatics/design_spaces/product_design_space.py
bfolie: No I don't think so. I also don't think we strictly enforce "no graph predictors inside graph predictors." But in both cases we don't support it, it might not work, and it provides no additional functionality. So I changed the comment to "should not."

bfolie:
I resolved the merge conflicts (happy to be rid of schema ids), made some small comment changes, and added docs.

e2es passed against development using preload. I'm re-running [against dev without preload](https://jenkins.corp.citrine.io/job/nextgen-ITs/11475/) and [against stage with preload](https://jenkins.corp.citrine.io/job/nextgen-ITs/11476/).

bfolie:
Update: all e2es passed

maxhutch (approved):
:100:

PR #500, opened 2020-12-17
Rearrange evaluation metrics docs.

gregor-robinson:
# Citrine Python PR

## Description 
This rearranges the exposition of available predictor evaluation metrics, so as to avoid leaving the reader with the misleading first impression that PVA is unavailable for numeric and categorical types. Describing PVA first gets that out of the way so they don't need to go back and reinterpret the preceding text.

### PR Type:


latture (changes requested):
Needs a version bump, but looks good otherwise.

latture (approved):

PR #503, opened 2020-12-18
Feature/pla 6165 exact ranged bounds

dleblanc:
Add exact ranges to pair with the backend changes in 6165.

These are mandatory here, and I've bumped the minor version.

### PR Type:


tests/resources/test_file_link.py
```diff
@@ -407,6 +407,11 @@ def test_process_file(collection, session):
                                 'type': 'integer_bounds',
                                 'lower_bound': 0,
                                 'upper_bound': 10
+                            },
+                            'exact_range_bounds': {
+                                'type': 'integer_bounds',
+                                'lower_bound': 0,
+                                'upper_bound': 10
```
kroenlein, line 414: If I'm following the intent here, shouldn't we have the exact and recommended values be different to make sure things are plumbed correctly?

kroenlein (approved):
Quick, non-blocking question.

tests/resources/test_file_link.py
dleblanc: Well we'e relying directly on the serialization stuff, I'm not concerned about a mismatch since we don't otherwise interact with it. We're well tested in the backend (and e2e's), so I think this is fine.

kroenlein: I'll trust your instinct.

kroenlein (approved):

PR #504, opened 2020-12-21
Break out predictors

bfolie:
# Citrine Python PR

## Description 
One file per predictor

### PR Type:


src/citrine/informatics/predictors/expression_predictor.py
```diff
@@ -0,0 +1,194 @@
...
+
+    def _post_dump(self, data: dict) -> dict:
+        data['display_name'] = data['config']['name']
+        return data
+
+    def __str__(self):
+        return '<ExpressionPredictor {!r}>'.format(self.name)
+
+
+class DeprecatedExpressionPredictor(Serializable['DeprecatedExpressionPredictor'], Predictor):
```
bfolie, line 73: @latture how do we decide when we can remove this?

src/citrine/informatics/predictors/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+from .predictor import *
+from .expression_predictor import *
+from .simple_ml_predictor import *
+from .generalized_mean_property_predictor import *
+from .graph_predictor import *
+from .ingredient_fractions_predictor import *
+from .ingredients_to_simple_mixture_predictor import *
+from .label_fractions_predictor import *
+from .molecular_structure_featurizer import *
+from .simple_mixture_predictor import SimpleMixturePredictor
```
talltea, line 11: why import these explicitly and not with `*`? (not blocking, just curious)

talltea (dismissed):
👍

src/citrine/informatics/predictors/expression_predictor.py
latture: We should follow up with product/DS to see if it's still being used. My guess is that it's probably safe, but we should check and potentially make an announce that it's being removed in the next version first.

src/citrine/informatics/predictors/__init__.py
bfolie: That's an oversight. Fixed

latture:
@bfolie One thing I've noticed by breaking out the other modules into separate files is that this breaks class references in our user docs. We want to check if that's the case here, but for workflows we had to use the full path, e.g. change :class:`~citrine.informatics.workflows.DesignWorkflow` to :class`~citrine.informatics.workflows.design_workflow.DesignWorkflow`

bfolie:
> @bfolie One thing I've noticed by breaking out the other modules into separate files is that this breaks class references in our user docs. We want to check if that's the case here, but for workflows we had to use the full path, e.g. change :class:`~citrine.informatics.workflows.DesignWorkflow` to :class`~citrine.informatics.workflows.design_workflow.DesignWorkflow`

Good catch, thanks. I'll check and update the docs.

talltea (approved):

PR #507, opened 2020-12-24
Doc suggestion - preview_roots parameter in table_config.preview

xperrylinn:
# Citrine Python PR

## Description 
Suggesting to update the preview method preview_roots parameter docstring to say "material run" instead of "material history". It seems ambiguous in whether it could be a run, spec, or template. I was personally confused because the only available row definition takes in templates, not runs. Just a suggestion! :-)

### PR Type:


src/citrine/resources/table_config.py
```diff
@@ -383,7 +383,7 @@ def preview(self, table_config: TableConfig, preview_roots: List[LinkByUID]) ->
         table_config: TableConfig
             Table Config to preview
         preview_roots: List[LinkByUID]
-            List of links to the material history roots to use in the preview
+            List of links to the material runs to use in the preview
```
kroenlein, line 386:
```suggestion
            List of links to the material runs to use as terminal materials in the preview
```

kroenlein (approved):
I think the suggestion makes it a little clearer what role the runs play, but maybe it's just adding unnecessary verbiage.

You need to bump the patch version in order to merge back to main branch -- this is true for every commit to the library, even if all edits are to documentation.

src/citrine/resources/table_config.py
xperrylinn: I dig it

xperrylinn:
Bumped the version number and synced with the master branch.

kroenlein (approved):

PR #509, opened 2020-12-29
Remove bounds from min/max objectives

maxhutch:
# Citrine Python PR

## Description 
These can be inferred from the descriptor; we don't need them to
be provided.  These classes are alpha.

### PR Type:


latture:
When will we infer bounds if we remove them here? Maybe when a design workflow is run or will objectives require descriptors instead of keys?

maxhutch:
> When will we infer bounds if we remove them here? Maybe when a design workflow is run or will objectives require descriptors instead of keys?

I was thinking in the design workflow run, where we can grab the descriptors from the predictor.

latture:
> I was thinking in the design workflow run, where we can grab the descriptors from the predictor.

Did you want to do that before making this change or remove objective bounds entirely?

maxhutch:
> Did you want to do that before making this change or remove objective bounds entirely?

No, since I think the overwhelming majority of users (if not the entirety) don't set these bounds, so applying them automatically would be a change to default behavior.  I brought it up more to indicate that we don't need to collect the information here **even if** want to set these bounds in the default case.

latture (dismissed):

latture (approved):

PR #513, opened 2021-01-08
attribute constraints type hint typo?

xperrylinn:
I get the following error if I try to use a list of tuples in specifying attribute constraints, which says that my attribute constraints need to be a list of lists. I think the correct type hint should say list[list[LinkByUID, Bounds]] instead of list[(LinkByUID, Bounds)]

ValueError: ({'scope': 'hello world', 'id': 'Viscosity Temperature'}, {'lower_bound': 129, 'upper_bound': 131, 'default_units': 'degree_Celsius', 'type': 'real_bounds'}) is not one of valid types: <class 'list'> 

I haven't confirmed the ValueError is raised for Variable types other than AttributeByTemplate, but I'm assuming the same is true for all other Variable types.

# Citrine Python PR

## Description 

### PR Type:


maxhutch (commented):
Nice catch -- I checked it with the usages in the tests here and in our e2es and it is indeed used as a list.  

I'm sorry that I didn't get to this earlier and the version bump has fallen out of date.  I'll approve when that's updated.

maxhutch (dismissed):

maxhutch (approved):

PR #519, opened 2021-01-15
Added more detailed http and api_error info to the NonNonRetryableHttpException

dleblanc:
Instead of only showing the path that failed, we can now see:
- The path, status code, reason, and api error info (with a fallback to
textual response).

This would have really helped with the test_material_history debugging.

Please let me know if this level of detail is desired by default (I would think it's useful).

### PR Type:


dleblanc:
An example test run with a forced validation failure looks like:
```
    raise BadRequest(path, response)
E   citrine.exceptions.BadRequest: Bad Request (code: 400) returned from POST request to path: 'projects/7b4cb782-83a5-4299-9c4e-d72b53ddb420/datasets/0b2bb39e-1a59-464e-a9cf-f8718033f138/process-specs'
E   	Error validating create the process spec
E   Multiple condition with named condition_name
E   	Multiple parameter with named parameter_name
```

src/citrine/exceptions.py
```diff
@@ -33,27 +33,46 @@ class NonRetryableHttpException(NonRetryableException):
...
             except (TypeError, ValueError):
                 self.api_error = None
+                self.detailed_error_info = "\n\t{}".format(response.text)
         else:
+            self.http_err = path
             self.response_text = None
             self.code = None
             self.api_error = None
 
+        super().__init__("{}\n{}".format(self.http_err, self.detailed_error_info))
```
kroenlein, line 74: I very much approve of using newlines and tabs to provide a more legible return, but if possible we should be collecting the errors in a list and assembling all the formatting logic in one place.

djack (approved):
I agree with Ken, but this is an unscheduled improvement in the state of the world.

src/citrine/exceptions.py
dleblanc: I've done that in a new commit -fixed up the formatting nicely.

dleblanc:
The formatted errors now look like:

```
    raise BadRequest(path, response)
E   citrine.exceptions.BadRequest: Bad Request (code: 400) returned from POST request to path: 'projects/e8c338aa-bc33-4ad0-a8cd-35e6b628ae26/datasets/34e82dfd-adc1-4f4f-a131-16b34056e582/process-specs'
E   	Multiple condition with named condition_name (attribute.duplicate)
E   	Multiple parameter with named parameter_name (attribute.duplicate)
```

src/citrine/exceptions.py
```diff
@@ -33,27 +33,48 @@ class NonRetryableHttpException(NonRetryableException):
...
             self.response_text = response.text
             self.code = response.status_code
+
+            method = "unknown"
+            if response.request is not None:
+                method = response.request.method
+
+            self.detailed_error_info = ["{} (code: {}) returned from {} request to "
+                                        "path: '{}'".format(response.reason, self.code,
+                                                            method, path)]
```
kroenlein, line 48: Functionally equivalent, but future-proofs a bit
```suggestion
            self.detailed_error_info.append(
                "{} (code: {}) returned from {} request to path: '{}'".format(
                    response.reason, self.code, method, path
                )
            )
```

kroenlein (approved):
Minor reformat

dleblanc:
Added a couple exception fixes I didn't notice previously.

bfolie (approved):

PR #520, opened 2021-01-19
[PLA-5845] Add method that constructs default PEW from predictor id.

latture:
This PR adds a method to `PredictorEvaluationWorkflowCollection` that constructs a default workflow from a predictor id.


src/citrine/resources/predictor_evaluation_workflow.py
```diff
@@ -53,3 +54,23 @@ def delete(self, uid: Union[UUID, str]) -> Response:
         """Predictor Evaluation Workflows cannot be deleted; they can be archived instead."""
         raise NotImplementedError(
             "Predictor Evaluation Workflows cannot be deleted; they can be archived instead.")
+
+    def create_default(self, predictor_id: UUID) -> PredictorEvaluationWorkflow:
+        """Create a default predictor evaluation workflow for a predictor.
```
maxhutch, line 59:
```suggestion
        """[ALPHA] Create a default predictor evaluation workflow for a predictor and execute it.
```

maxhutch (commented):
Looks good.  One doc nit.  Maybe its worth describing how it forms the default?

latture:
> Maybe its worth describing how it forms the default?
 
That's a good idea. I'll add some documentation.

src/citrine/resources/predictor_evaluation_workflow.py
```diff
@@ -53,3 +54,36 @@ def delete(self, uid: Union[UUID, str]) -> Response:
         """Predictor Evaluation Workflows cannot be deleted; they can be archived instead."""
         raise NotImplementedError(
             "Predictor Evaluation Workflows cannot be deleted; they can be archived instead.")
+
+    def create_default(self, predictor_id: UUID) -> PredictorEvaluationWorkflow:
```
latture, line 58: @maxhutch What about something like what I've written below. I'm worried these docs might get out of date, so I'm open to suggestions on what you think could go here instead.

src/citrine/resources/predictor_evaluation_workflow.py
```diff
@@ -53,3 +54,36 @@ def delete(self, uid: Union[UUID, str]) -> Response:
...
+            Unique identifier of the predictor used to create a default workflow
+
+        Returns
+        -------
+        PredictorEvaluationWorkflow
+            Default workflow
+
+        """  # noqa: E501
+        url = self._get_path('default')
+        ref = PredictorRef(str(predictor_id))
```
talltea, line 86: Why do we want to wrap this up in a PredictorRef rather than using UUID directly?

latture: I didn't know we could do that. If the backend expects a `PredictorRef` and I use the UUID directly with `self.session.post_resource(url, predictor_id)`, will the backend handle it correctly?

latture: That results in `TypeError: Object of type UUID is not JSON serializable`, but maybe if I convert to a string first?

latture: Nope, casting to a `str` returns a 400. I'm not sure what you mean by use it directly. If you're asking why the backed expects a `PredictorRef`, it was just for consistency with the `ModuleRef` that we use elsewhere.

talltea: I don't think the backend distinguishes between uid types, but I see that this backend endpoint is expecting a post body of the form `{"predictor_id": "c7e50770-c22a-4065-9578-9761ebdfc77b"}`. Which you can get by doing `ref.dump()` if ref is a PredictorRef, or by being more explicit and saying something like `data = self.session.post_resource(url, {"predictor_id": predictor_id})` if `predictor_id` is the relevant id (or something similar, you might need to dump the `UUID` anyway?).

I would prefer not creating a new reference class for this, since I think the analogous ModuleRef has caused some confusion, and it doesn't seem to add anything other than this dict wrapping, but I might be missing something!

talltea: oh, i was thinking something like `self.session.post_resource(url, {"predictor_id": predictor_id.dump()})` (commented before i saw the rest of your replies)

latture: That's fine. I don't have a strong preference either way (especially since this is the only place it's used). I deleted it in the latest commit, and now I create the payload manually like you suggested.

src/citrine/resources/predictor_evaluation_workflow.py
maxhutch: Looks good!

talltea (approved):
lgtm!

PR #525, opened 2021-01-26
Added thin wrappers for asset ownership lists

kroenlein:
# Citrine Python PR

## Description 
This PR exposes endpoints for determining which assets are owned by particular projects.

I feel like these methods are a little too thin but fleshing out the return values would require additional service calls, which also doesn't match other behaviors.

### PR Type:


src/citrine/resources/project.py
```diff
@@ -315,6 +315,71 @@ def make_private(self,
...
+        Returns
+        -------
+        List[str]
+            The ids of the table configs owned by current project
+
+        """
+        result = self.session.get_resource(self._path() + "/table_definition_ids")
+        return result["table_definition_ids"]
+
+    def owned_module_ids(self) -> List[str]:
```
maxhutch, line 370: I don't think we need to expose this since modules cannot currently be shared, and I don't think we want to since we're generally moving to more strongly typed modules (e.g. differentiating between predictors and design spaces).

maxhutch (changes requested):
The implementation looks sensible.  @jfriend-citrine : are you OK with these methods being bound in citrine-python (it'll make it a little harder to update the APIs since we'll need to have this client in mind).  

On the AI engine side, I don't think we actually need or want to expose the list of owned module ids.

src/citrine/resources/project.py
kroenlein: Makes sense to me, and they'll get pulled once it's clear what ESI's opinions are.

Presumably this means you agree with just listing UIDs instead of building objects.

jfriend-citrine: @maxhutch LGTM. Trusting test coverage as our safety net.

jfriend-citrine (approved):
LGTM

maxhutch (approved):
LGTM

PR #526, opened 2021-01-28
Feature/pla 6500 data object batch delete

dleblanc:
This exposes some support for the batch delete data-obj endpoint.

### PR Type:


src/citrine/resources/project.py
```diff
@@ -430,6 +432,45 @@ def remove_user(self, user_uid: Union[str, UUID]) -> bool:
         )
         return True
 
+    def gemd_batch_delete(self, id_list: Union[List[Tuple[str, str]], List[UUID]]) -> \
+            List[Tuple[str, str, ApiError]]:
+        """
```
kroenlein, line 437: Please describe the arguments to the method

src/citrine/resources/project.py
```diff
@@ -430,6 +432,45 @@ def remove_user(self, user_uid: Union[str, UUID]) -> bool:
         )
         return True
 
+    def gemd_batch_delete(self, id_list: Union[List[Tuple[str, str]], List[UUID]]) -> \
+            List[Tuple[str, str, ApiError]]:
```
kroenlein, line 436: For all the confusion and inconsistency in these signature, you should probably copy the sig from `data_concepts.delete` to maintain some consistency.  That, unfortunately, would remove support for external IDs, but is futureproof to enabling them later.

kroenlein (changes requested):

src/citrine/resources/project.py
dleblanc: What do you mean, 'futureproof to enabling them later'. Why would we purposefully exclude the ability to delete via external ID when that works now?
I've looked at the signature for .delete(), and it doesn't seem to really apply for a bulk scenario:
 - You would have to supply a single scope for all IDs, that seems suboptimal
 - We don't support dry_run in a batch delete scenario, and that would be challenging due to the inbound links that are progressively resolved (and depend on the deleted items)
 - .delete() blindly returns a successful Response object, which doesn't provide nearly the detail on failed items that this does

kroenlein: This difference of opinion is driven, in part, by the lack of actual customer signal.  To my understanding, the users who have need for this feature at present are 1) DE for cleaning up after a failed load and 2) engineering for speeding up tests.  Neither of those scenarios is substantially benefited by allowing a wider range of IDs.  I was not suggesting that we also include `dry_run` though I understand I was ambiguous on that point.

Right now, call signatures in citrine-python are very heterogenous.  I don't think we have sufficient justification at present to refactor to address this, but this particular signature choice (a scope-UID tuple) is introducing yet another pattern.  

If @djack concurs with your analysis, I'll swap to an approve.

src/citrine/resources/project.py
```diff
@@ -2,44 +2,46 @@
...
-from citrine.resources.gemtables import GemTableCollection
+from citrine.resources.process_template import ProcessTemplateCollection
+from citrine.resources.processor import ProcessorCollection
+from citrine.resources.project_member import ProjectMember
+from citrine.resources.project_roles import MEMBER, ROLES, ACTIONS
+from citrine.resources.property_template import PropertyTemplateCollection
+from citrine.resources.response import Response
+from citrine.resources.table_config import TableConfigCollection
 from citrine.resources.user import User
+from citrine.resources.workflow import WorkflowCollection
```
djack, line 44: That's quite a diff.

src/citrine/resources/project.py
djack: Ken's point is not unreasonable - the users of this endpoint are (for now) going to want to be able to copy/paste and have things just work. This signature forces them to think a bit more if they want to make use of this endpoint. 

However, my instinct here is that Dave's signature is pretty natural for a batch endpoint on gemd objects than the `.delete` and `.get` signatures we have today. It's not so much a brand new pattern as it is a version of the same pattern that is list valued, as opposed to individually valued. I do think we could paper over a lot of these issues by (also) accepting a https://github.com/CitrineInformatics/gemd-python/blob/master/gemd/entity/link_by_uid.py in the union type. I have a suspicion that regardless of whether we follow @dleblanc or @kroenlein's path, there are going to be a lot of helper functions that transform / split linkbyuids.

I think any frustrations about API consistency could likely be papered over by being consistent when we add batch-get support to citrine-python https://citrine.atlassian.net/browse/PLA-6574

kroenlein: Nice call on LinkByUID as the identifier wrapper.  That also matches well with endpoints that take objects directly.

src/citrine/resources/project.py
dleblanc: IntelliJ optimize imports. Mistakes were made. Regrets were had.

src/citrine/resources/project.py
kroenlein: W.r.t. the tuple, I can't find an instance in either gemd-python or citrine-python where we have a scope-uid tuple.  If someone can find an instance of that, I'll approve.  Otherwise, I think we should remove the tuple from the argument list.

kroenlein (approved):

PR #531, opened 2021-02-01
Make graph predictor deserialization more permissive

maxhutch:
# Citrine Python PR

## Description 

Relax the expectations on GraphPredictors coming back from the API so they are valid even when not yet registered.

### PR Type:


src/citrine/informatics/predictors/graph_predictor.py
```diff
@@ -35,11 +35,9 @@ class GraphPredictor(Serializable['GraphPredictor'], Predictor):
 
     predictors = _properties.List(_properties.Union(
         [_properties.UUID, _properties.Object(Predictor)]), 'config.predictors')
-    training_data = _properties.List(_properties.Object(DataSource), 'config.training_data')
+    training_data = _properties.List(
+        _properties.Object(DataSource), 'config.training_data', default=[])
     typ = _properties.String('config.type', default='Graph', deserializable=False)
-    # Graph predictors may not be embedded in other predictors, hence while status is optional
-    # for deserializing most predictors, it is required for deserializing a graph
-    status = _properties.String('status', serializable=False)
```
maxhutch, line 42: When returned via check-for-updates, the predictor isn't actually registered so it has no status.  Removing this line causes the `status` definition in `Predictor` to kick in, which lists this as optional.

src/citrine/informatics/predictors/graph_predictor.py
```diff
@@ -35,11 +35,9 @@ class GraphPredictor(Serializable['GraphPredictor'], Predictor):
 
     predictors = _properties.List(_properties.Union(
         [_properties.UUID, _properties.Object(Predictor)]), 'config.predictors')
-    training_data = _properties.List(_properties.Object(DataSource), 'config.training_data')
+    training_data = _properties.List(
+        _properties.Object(DataSource), 'config.training_data', default=[])
```
maxhutch, line 39: The backend has a `None` default, so this turns it back into an empty list.

latture: Using a mutable default value could lead to strange behavior. Is there a way to make the default immutable (maybe a using tuple but I'm not sure how this would affect our typed serialization) or translate `None` to an empty list in the constructor?

latture: But this is on the class level. Even translating it inside the constructor would lead to the same list getting used for every deserialization.

latture: I think we'll need to translate `None` to a `[]` at some point during deserialization, but looking at the code I'm not sure where the best place to do that is.

latture: The values are read into a new list each time an object is deserialized here: https://github.com/CitrineInformatics/citrine-python/blob/master/src/citrine/_serialization/properties.py#L370

So this shouldn't be an issue. Thanks to @maxhutch for adding a test to check.

latture (approved):

PR #532, opened 2021-02-01
Deprecate Score's name and description for future removal

maxhutch:
# Citrine Python PR

## Description 
The name and description fields on scores are confusing since
they are not modules (i.e. assets or workflows).  We're removing
them, but want to do so with a deprecation cycle.  This deprecates
setting and getting them.

Note that there is a breaking change here: because name and
description were the first two arguments in the constructor,
they can't simply be given defaults in the first two positions.
To resolve this, the PR adds the keyword-only setting to the
constructors, which is a best practice anyways.

### PR Type:


src/citrine/__version__.py
```diff
@@ -1 +1 @@
-__version__ = '0.99.3'
+__version__ = '0.100.0'
```
talltea, line 1: 💯 🎉

talltea (approved):
Do we need to fix up e2es before merging this? (Both removing name/description and using named params)

maxhutch:
> Do we need to fix up e2es before merging this? (Both removing name/description and using named params)

already checked

PR #533, opened 2021-02-01
Calling a PUT dataset if it already has an ID assigned

dleblanc:
Prior to this, .register on a dataset would only ever do a POST on a resource. POST will discard any id present and assign a new one. PUT is used to update resources.

So now we perform a .PUT if the dataset being handed to .register has an ID assigned. This should allow us to re-register datasets without issue.

### PR Type:


src/citrine/resources/delete.py
```diff
@@ -0,0 +1,71 @@
...
+        id_list: List[Union[LinkByUID, UUID]],
+        project_id: UUID,
+        session: Session,
+        dataset_id: Optional[UUID] = None
+) -> List[Tuple[LinkByUID, ApiError]]:
+    """
+    Shared implementation of GEMD Batch deletion.
+
+    You may provide GEMD objects that reference each other, and the objects
+    will be removed in the appripriate order.
```
djack, line 20:
```suggestion
    will be removed in the appropriate order.
```

src/citrine/__version__.py
```diff
@@ -1 +1 @@
-__version__ = '0.98.3'
+__version__ = '0.98.4'
```
djack, line 1:
```suggestion
__version__ = '0.98.5'
```

djack (approved):

PR #534, opened 2021-02-03
Feature/pla 6572 per dataset batch delete

dleblanc:
Added a per-dataset batch delete, sharing code with the project level batch delete.

This only had to fix the UUID -> string conversion of the Dataset ID.

### PR Type:


src/citrine/resources/dataset.py
```diff
@@ -321,7 +321,7 @@ def gemd_batch_delete(self, id_list: List[Union[LinkByUID, UUID]]) -> \
 
         """
         from citrine.resources.delete import _gemd_batch_delete
-        return _gemd_batch_delete(id_list, self.project_id, self.session, self.uid)
+        return _gemd_batch_delete(id_list, self.project_id, self.session, str(self.uid))
```
kroenlein, line 324: This actually violates the declared signature: 
```
def _gemd_batch_delete(
        id_list: List[Union[LinkByUID, UUID]],
        project_id: UUID,
        session: Session,
        dataset_id: Optional[UUID] = None
) -> List[Tuple[LinkByUID, ApiError]]:
```
Either update the sig to expect a string, or do the stringification in the `_gemd_batch_delete` method body (preferred)

kroenlein (changes requested):

src/citrine/resources/dataset.py
dleblanc: <omit python typing rant> ... yeah the latter sounds better.

kroenlein (approved):

PR #546, opened 2021-02-23
Added pagination and sorting to batch delete

kroenlein:
# Citrine Python PR

## Description 
This PR makes two changes:

1. It modifies the signature for `gemd_batch_delete` to allow for UUID-like strings and Base Entities (Objects & Templates).  
2. If it receives a list of more than 50 elements, it attempts to sort them into delete order and then paginate them accordingly.

1 was necessary for 2 because no sorting is possible with just the UUIDs unless we literally fetch all types from the server first.  I opted not to do this because this would require implementing the storables route in this library, which I think should happen eventually but would have substantially reduced speed on responding to this need.

Non-breaking but adds new elements to the API, thus a minor version bump.

### PR Type:


src/citrine/resources/dataset.py
```diff
@@ -289,28 +289,33 @@ def delete(self, data_concepts_resource: ResourceType, dry_run=False) -> Resourc
...
 
         A failure will be returned if the object cannot be deleted due to an external
         reference.
 
         All data objects must be associated with this dataset resource. You must also
         have write access on this dataset.
 
+        If you wish to delete more than 50 objects, queuing of deletes requires that
+        the types of objects be known, and thus you _must_ provide ids in the form
+        of BaseEntities.
```
dleblanc, line 308: Long-term I suppose we could also query that in a batch-get first? For now this seems expedient.

src/citrine/resources/delete.py
```diff
@@ -48,23 +57,45 @@ def _gemd_batch_delete(
         deleted.
 
     """
+    if len(id_list) > DELETE_SERVICE_MAX:  # we need to sort it
+        if any([not isinstance(x, BaseEntity) for x in id_list]):
+            raise TypeError(
+                "If more than {} deletes are requested, id_list must contain "
+                "only BaseEntities (objects & templates)".format(DELETE_SERVICE_MAX)
+            )
+        id_list = sorted(id_list, key=lambda x: writable_sort_order(x), reverse=True)
```
dleblanc, line 66: I admit I am concerned about duplication of logic between here and the backend, but not sure we have much of a choice at present.

dleblanc (approved):
LGTM

PR #547, opened 2021-02-24
PLA-6781 missing fields

talltea:
# Citrine Python PR

## Description 
These fields were missing for the Design Workflow.

### PR Type:


src/citrine/informatics/workflows/design_workflow.py
```diff
@@ -47,12 +47,17 @@ class DesignWorkflow(Resource['DesignWorkflow'], Workflow):
...
+    archived_by = properties.Optional(properties.UUID, 'archived_by', serializable=False)
     create_time = properties.Optional(properties.Datetime, 'create_time', serializable=False)
+    update_time = properties.Optional(properties.Datetime, 'update_time', serializable=False)
+    archive_time = properties.Optional(properties.Datetime, 'archive_time', serializable=False)
     module_type = properties.String('module_type', default='DESIGN_WORKFLOW')
     typ = properties.String('type', default='DesignWorkflow', deserializable=False)
 
     status_description = properties.String('status_description', serializable=False)
     name = properties.String('name')
+    description = properties.Optional(properties.String, 'description')
```
talltea, line 60: ah, i should probably add this to the init too

maxhutch: This is probably due to a bad review on my part :-(

maxhutch (commented):
Other than description, these fields aren't currently exposed in citrine-python for any module/workflow (afaik).  Is there a particular use case for adding them (e.g. testing, or a user request)?

If we do want to add them, then we should add them to predictors, design spaces, processors, etc too.  Might make sense to define a mix in that defines them all like how the base classes define `status` and `status_info` today ([design space example](https://github.com/CitrineInformatics/citrine-python/blob/master/src/citrine/informatics/design_spaces/design_space.py#L24)).

tests/conftest.py
```diff
@@ -798,6 +798,7 @@ def design_workflow_dict(generic_entity):
     ret = generic_entity.copy()
     ret.update({
         "name": "Example Design Workflow",
+        "description": "A description! Of the Design Workflow! So you know what it's for!",
```
latture, line 801: Did you mean to include `description` twice?

talltea: noooope 🤣

src/citrine/_rest/mixin.py
```diff
@@ -0,0 +1,18 @@
+from typing import TypeVar, Optional  # noqa: F401
+from citrine._serialization.serializable import Serializable
+from citrine._serialization import properties
+
+Self = TypeVar('Self', bound='Mixin')
+
+
+class Mixin():
```
talltea, line 8: Please to help with name 🙏

bfolie: suggestion: `ResourceMetadata`

bfolie:
We can pull other fields into this abstraction: `experimental`, `experimental_reasons`, `archived`, `status`, `status_info`, and eventually `status_description` (though that one's not general yet)

tests/conftest.py
bfolie: this still needs to be removed

src/citrine/informatics/workflows/design_workflow.py
```diff
@@ -33,27 +34,16 @@ class DesignWorkflow(Resource['DesignWorkflow'], Workflow):
 
     uid = properties.Optional(properties.UUID, 'id', serializable=False)
     name = properties.String('display_name')
```
bfolie, line 36: `name` is implemented twice

talltea: the two different `name`s are for the old style and new style workflows. `display_name` was used by the module (old) and `name` is used by the DW (new). The overwriting is weird, but not a change introduced in this pr

src/citrine/informatics/design_spaces/design_space.py
```diff
@@ -5,6 +5,7 @@
 from citrine._serialization.serializable import Serializable
 from citrine.informatics.modules import Module
 
+
 __all__ = ['DesignSpace']
```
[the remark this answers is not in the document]
talltea: I was trying to make this work, but wasn't able to. I couldn't get the ai_resource_metadata fields to serialize, I think because of how the build methods are set up in PolymorphicSerializable and Serializable

I agree that it would be much nicer to pull the metadata into the Module level :(

bfolie (dismissed):

maxhutch (approved):
@bfolie approved last week, so I'm just checking the version bump

PR #551, opened 2021-03-10
[PLA-6545] construct default design space from predictor

latture:
This PR adds a method `create_default` method to the `DesignSpaceCollection` which constructs a design space from a predictor. The implementation mirrors that of [`PredictorEvaluationWorkflowCollection.create_default`](https://github.com/CitrineInformatics/citrine-python/blob/master/src/citrine/resources/predictor_evaluation_workflow.py#L57).

### PR Type:


src/citrine/resources/design_space.py
```diff
@@ -62,3 +62,34 @@ def update(self, model: DesignSpace) -> DesignSpace:
...
+        Returns
+        -------
+        DesignSpace
+            Default design space
+
+        """
+        path = f'projects/{self.project_id}/predictors/{predictor_id}/default-design-space'
+        data = self.session.get_resource(path)
+        if 'instance' in data:
+            data['config'] = data.pop('instance')
```
talltea, line 94: Is this because of the strange wrapping we needed to do in the backend?

latture: Yeah, it's similar to what we do for subspaces in the pre-build method. However, I can't do the rename there because we assume the instance key is present when we get the class.

maxhutch (approved):

talltea (approved):

PR #553, opened 2021-03-12
PLA-6368 switch from responses to output for autoMLPredictor

talltea:
# Citrine Python PR

## Description 
This changes the AutoML interface from accepting `responses` (list of descriptors) to `output` (single descriptor). This is **different** than the backend API, which accepts `outputs` as a list of descriptors.

This change is backwards compatible so that the old backend logic will work with this python release. Once we've released the new backend logic to production, we can remove the backwards compatibility.

I also updated the docs to have the AutoML predictor show up first. The `Predictors` documentation page now looks like:
Screen Shot 2021-03-18 at 9 25 10 AM

### PR Type:


bfolie (dismissed):

maxhutch (approved):
LGTM

PR #556, opened 2021-03-15
[PLA-6448] Documented that generalized power must be an integer. 

latture:
This adds docs to the `GeneralizedMeanPropertyPredictor` and changes the type of `p` from a `float` to an `int`.


src/citrine/informatics/predictors/generalized_mean_property_predictor.py
```diff
@@ -55,7 +55,7 @@ class GeneralizedMeanPropertyPredictor(
 
     input_descriptor = _properties.Object(FormulationDescriptor, 'config.input')
     properties = _properties.List(_properties.String, 'config.properties')
-    p = _properties.Float('config.p')
+    p = _properties.Integer('config.p')
```
latture, line 58: Should we change this on the backend also, or leave is as a float in case we want to allow non-integer powers in the future?

docs/source/workflows/predictors.rst
```diff
@@ -462,7 +463,7 @@ This predictor will compute a real descriptor with a key ``mean of property dens
         inputs=[formulation_descriptor]
     )
 
-If ``p`` is given a value other than ``1.0``, that value will be included in the key for the feature, e.g. ``2.0-mean of property viscosity``.
+If ``p`` is given a value other than ``1``, that value will be included in the key for the feature, e.g. ``2.0-mean of property viscosity``.
```
latture, line 466: If we only support integers we may want to change the string representation so decimals aren't included, e.g. `2-mean of property viscosity` instead of `2.0-mean`.

bfolie: I think it is "2-mean", actually: https://github.com/CitrineInformatics/mithril/blob/master/src/main/scala/io/citrine/nucleus/ml/relations/library/FormulaExpander.scala#L204

src/citrine/informatics/predictors/generalized_mean_property_predictor.py
bfolie: I left a comment on the ticket: https://citrine.atlassian.net/browse/PLA-6448. I think that we should leave it as a float and not modify the backend.

latture: How do you feel about changing it here? It seems strange to me that we would type is as a `float` in Python but throw a validation error if it's anything but an `int`. However the mismatch between Python and the backend is weird too. I _think_ this is clearer from a user's perspective, but I could see the argument for leaving the type as-is.

bfolie: Yes, sorry, I think you're right that we should change it to an Integer here. But we should leave it as a Double in the backend. I don't think that will cause deser issues, since everything that has been persisted so far has been an integer.

bfolie (approved):

PR #560, opened 2021-03-16
Issue deprecation warning if p is a float.

latture:
Our serialization logic doesn't allow floats to be specified for integer fields (even if `float.is_integer() == True`). Instead of breaking existing code, this PR issues a deprecation warning if `p` is a `float`.


bfolie (approved):
How hard would it be to change the serialization logic so that it is tolerant of floats that are equivalent to an integer?

latture:
I tried that first, but I couldn't get it to work. It's non-trivial (at least for me), but someone with more experience with our serialization logic could probably do it.

PR #561, opened 2021-03-18
Performing an async batch deletion for project/dataset

dleblanc:
This means we:
 - can now remove attribute templates
- no long have a max size limit (though we can time out from a client
perspective - the async process will continue on the server side). The
client timeout can be arbitrarily large.

Removed the sync batch delete, as this does everything that did with
fewer limitations.

Sharing the job polling in a better place.

### PR Type:


src/citrine/__version__.py
```diff
@@ -1 +1 @@
-__version__ = '0.107.1'
+__version__ = '0.108.1'
```
kroenlein, line 1: Weird version bump
```suggestion
__version__ = '0.108.0'
```

src/citrine/resources/dataset.py
```diff
@@ -3,6 +3,18 @@
...
+from gemd.entity.base_entity import BaseEntity
+from gemd.entity.link_by_uid import LinkByUID
+from gemd.entity.object import MeasurementSpec, MeasurementRun, MaterialSpec, MaterialRun, \
+    ProcessSpec, ProcessRun, IngredientSpec, IngredientRun
+from gemd.entity.template.condition_template import ConditionTemplate
+from gemd.entity.template.material_template import MaterialTemplate
+from gemd.entity.template.measurement_template import MeasurementTemplate
+from gemd.entity.template.parameter_template import ParameterTemplate
+from gemd.entity.template.process_template import ProcessTemplate
+from gemd.entity.template.property_template import PropertyTemplate
```
kroenlein, line 15:
```suggestion
from gemd.entity.template import PropertyTemplate, MaterialTemplate, MeasurementTemplate, \
    ParameterTemplate, ProcessTemplate, ConditionTemplate
```

src/citrine/resources/delete.py
```diff
@@ -1,54 +1,53 @@
...
     ----------
-    id_list: List[Union[LinkByUID, UUID, str, BaseEntity]]
-        A list of the IDs of data objects to be removed. They can be passed
-        as a LinkByUID tuple, a UUID, a string, or the object itself. A UUID
-        or string is assumed to be a Citrine ID, whereas a LinkByUID or
-        BaseEntity can also be used to provide an external ID.
+    id_list: List[Union[LinkByUID, UUID]]
+        A list of the IDs of data objects to be removed. They can be passed either
+        as a LinkByUID tuple, or as a UUID. The latter is assumed to be a Citrine
+        ID, whereas the former can also be used to provide an external ID.
```
kroenlein, line 34: Why the change in documentation?  The sig matches the older version
```suggestion
    id_list: List[Union[LinkByUID, UUID, str, BaseEntity]]
        A list of the IDs of data objects to be removed. They can be passed
        as a LinkByUID tuple, a UUID, a string, or the object itself. A UUID
        or string is assumed to be a Citrine ID, whereas a LinkByUID or
        BaseEntity can also be used to provide an external ID.

File: tests/resources/test_dataset.py [hunk 4]
```diff
@@ -587,23 +590,32 @@ def test_batch_delete(dataset):
...
+    assert 2 == session.num_calls
     expect_call = FakeCall(
         method="POST",
-        path="/projects/{}/gemd/batch-delete".format(dataset.project_id)
+        path="/projects/{}/gemd/async-batch-delete".format(dataset.project_id)
     )
-    assert expect_call.method == session.last_call.method
-    assert expect_call.path == session.last_call.path
+    # assert expect_call.method == session.last_call.method
+    # assert expect_call.path == session.last_call.path
```
kroenlein, line 618: Why are these commented out?

kroenlein (dismissed):
This raises a question in my mind, which is did we ever establish what correct behavior for files is a dataset we wanted to wipe was?  I'm guessing that would be a different ticket if we're gonna attack it.

src/citrine/resources/delete.py
dleblanc: That wasn't intentional - good catch.

tests/resources/test_dataset.py
dleblanc: Since there's two calls, the 'last' portion isn't as useful. Generally speaking, our response should be enough here to satisfy the test - I don't think it's particularly important to assert on the call paths here.

dleblanc: I'll remove those lines entirely.

src/citrine/resources/dataset.py
dleblanc: The former is what my IDE auto-cleans up the imports as. I agree with the suggestion, but I don't think it's worth the effort to spend time being concerned about how imports are formatted.

dleblanc:
Converted to draft until the required functionality is merged to dev and promoted to stage.

dleblanc:
> This raises a question in my mind, which is did we ever establish what correct behavior for files is a dataset we wanted to wipe was? I'm guessing that would be a different ticket if we're gonna attack it.

You mean - what do we do when we call DELETE on a dataset, but there's still files associated? Technically we COULD leave the files there with just a delete marker as the current version (keeping the data), OR - we could wipe all the versions (removing it).

Delete markers combined with a suitable archival policy would mean we could safely keep old data on cold storage pretty cheaply.

I would imagine we don't want to do either without some kind of explicit opt-in (query param?).

dleblanc:
The backend functionality has been merged to develop and promoted to stage. Ready for review.

kroenlein:
> > This raises a question in my mind, which is did we ever establish what correct behavior for files is a dataset we wanted to wipe was? I'm guessing that would be a different ticket if we're gonna attack it.
> 
> You mean - what do we do when we call DELETE on a dataset, but there's still files associated? Technically we COULD leave the files there with just a delete marker as the current version (keeping the data), OR - we could wipe all the versions (removing it).
> 
> Delete markers combined with a suitable archival policy would mean we could safely keep old data on cold storage pretty cheaply.
> 
> I would imagine we don't want to do either without some kind of explicit opt-in (query param?).

It's just an inconsistent experience, and it can only be reconciled if we create a hard-delete on files.  One could nuke the content on a dataset, but be unable to delete the dataset.  For now, non-blocking.

kroenlein (dismissed):
LGTM

dleblanc:
> > > This raises a question in my mind, which is did we ever establish what correct behavior for files is a dataset we wanted to wipe was? I'm guessing that would be a different ticket if we're gonna attack it.
> > 
> > 
> > You mean - what do we do when we call DELETE on a dataset, but there's still files associated? Technically we COULD leave the files there with just a delete marker as the current version (keeping the data), OR - we could wipe all the versions (removing it).
> > Delete markers combined with a suitable archival policy would mean we could safely keep old data on cold storage pretty cheaply.
> > I would imagine we don't want to do either without some kind of explicit opt-in (query param?).
> 
> It's just an inconsistent experience, and it can only be reconciled if we create a hard-delete on files. One could nuke the content on a dataset, but be unable to delete the dataset. For now, non-blocking.

I think we can hard delete files as well. By default we don't really delete the file, but we insert a 'delete marker' as the latest version. But if we're actually clearing out the dataset (with the knowledge that it permanently removes data), then we can remove all underlying versions of a file as well.

This is a good point though, and something to probably add to the 'wipe dataset' story/stories.

kroenlein (approved):

PR #562, opened 2021-03-19
Add mean property predictor

latture:
This PR adds a `MeanPropertyPredictor` and deprecates the `GeneralizedMeanPropertyPredictor`. The former requires that properties are specified as list of `RealDescriptors` instead of a list of descriptor keys. (We use a list of real descriptors and not a set because real descriptors are not hashable, at least not currently.)


src/citrine/informatics/predictors/mean_property_predictor.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+class MeanPropertyPredictor(
+        Serializable['MeanPropertyPredictor'], Predictor):
+    """[ALPHA] A predictor interface that computes mean component properties.
+
+    .. seealso::
+       If you are using a deprecated generalized mean property predictor please see
+       :class:`~citrine.informatics.predictors.GeneralizedMeanPropertyPredictor`
+       for details on how to migrate to the new format.
```
bfolie, line 21: I don't see these details in `GeneralizedMeanPropertyPredictor`

bfolie:
>We use a list of real descriptors and not a set because real descriptors are not washable

Washable?

latture:
> > We use a list of real descriptors and not a set because real descriptors are not washable
> 
> Washable?

Sorry, autocorrect.. that should be hashable.

src/citrine/informatics/predictors/mean_property_predictor.py
latture: I don't know where those changes went. Hold on while I rewrite it...

latture: Just pushed the docs.

bfolie (approved):

PR #567, opened 2021-03-30
Added dataset.delete_contents

dleblanc:
This currently does a batch deletion of all GEMD objects inside a
dataset (async, supports all GEMD object types including templates).

### PR Type:


src/citrine/resources/delete.py
```diff
@@ -85,6 +88,46 @@ def _async_gemd_batch_delete(
...
 
+    return _poll_for_async_batch_delete_result(project_id, session, job_id, timeout, polling_delay)
+
+
+def _poll_for_async_batch_delete_result(
+        project_id: UUID,
+        session: Session,
+        job_id: str,
+        timeout: float,
+        polling_delay: float
```
kroenlein, line 99: This isn't a public method, so I don't know if the required-keyword criterion is relevant, but the argument list is quite long.

kroenlein (changes requested):

src/citrine/resources/delete.py
dleblanc: Do you have a suggestion on how to avoid that?

kroenlein: My inclination is to 
```suggestion
def _poll_for_async_batch_delete_result(
        *,
        project_id: UUID,
        session: Session,
        job_id: str,
        timeout: float,
        polling_delay: float
```
and force the private invocations to include keyword arguments, since you aren't actually inconveniencing a user by having the restriction.  But I'm not going to block on that one.

kroenlein (approved):

src/citrine/resources/delete.py
dleblanc: For private stuff, I just don't see the point - this isn't widely used, and we have unit tests.

PR #571, opened 2021-04-05
Make predictor report fetching lazy

maxhutch:
# Citrine Python PR

## Description

This removes the session and report from predictor constructors
and puts them in the predictor base class.  When a predictor
is gotten, the project_id and session from the collection is inserted
so that subsequent calls to report can lazily load the report contents.

Tests were updated.  Since this is in the base class, a lot
of test content was able to be removed.

This should resolve a race condition where the predictor can be returned from the API before the report was created.  It also just makes more sense...

### PR Type:


robbtraister (approved):
LGTM. Delete all the things!

talltea (approved):
🎉 nice!

PR #574, opened 2021-04-07
Add training_data_count to ModelSummary

maxhutch:
# Citrine Python PR

## Description 
We added this optional field in the model report coming from the
backend.  This models it in the client.

### PR Type:


src/citrine/informatics/reports.py
```diff
@@ -110,6 +110,9 @@ class ModelSummary(Serializable['ModelSummary']):
     predictor_name = properties.String('predictor_configuration_name', default='')
     predictor_uid = properties.Optional(properties.UUID(), 'predictor_configuration_uid')
 
+    training_data_count = properties.Optional(properties.Integer, "training_data_count")
+    """Number of rows in the training data for the model, if applicable."""
```
bfolie, line 114: Shouldn't this be under "Parameters"

maxhutch: No, since the user can't set it.  By adding it here, we're telling our serialization logic how to read it from the wire but not allowing a user to set it in a constructor.  The post-attribute docstring does show up in the autodoc (I had to check myself to be sure).

Before we hit 1.0, we probably want to adopt this pattern for other read-only fields.  Sometimes [we do it right](https://github.com/CitrineInformatics/citrine-python/blob/master/src/citrine/informatics/design_candidate.py) but here in reports its mostly wrong.

bfolie: Yes, I agree with you. The inconsistency bothers me (I'd rather we change this entire file), but if that's too time-consuming to do right now I'll just make sure it gets fixed pre-v1.0

maxhutch: Changing those constructors is technically breaking, so I'd rather batch it up with the other pre-v1.0 rather than have user deal with a thousand papercuts.

bfolie (approved):

PR #575, opened 2021-04-07
Chemical formula featurizer

bfolie:
# Citrine Python PR

## Description 
PLA-7120: Expose a chemical formula featurizer.

This is a minimum version of the feature--it exposes the functionality and the docstring lists all of the available features. But those features aren't explained very well, the features are not matched with their relevant aliases, and there's no example in the documentation. We could add that information to this PR, or if we want to expose the predictor ASAP we can merge this and add the documentation as a follow-up.

### PR Type:


src/citrine/informatics/predictors/chemical_formula_featurizer.py
```diff
@@ -0,0 +1,162 @@
...
+
+class ChemicalFormulaFeaturizer(Resource['ChemicalFormulaFeaturizer'], Predictor):
+    """
+    A featurizer for chemical formulae. Inspired by Magpie.
+
+    The ChemicalFormulaFeaturizer computes a configurable set of features on chemical formula data.
+    The features are functions of element-level properties, which are inspired by
+    `Magpie <https://bitbucket.org/wolverton/magpie/src/master/>`_. The features are configured
+    using the ``features`` and ``excludes`` arguments, which accept either feature names or
+    predefined aliases. Many features are element-wise weighted means of element-level properties.
```
maxhutch, line 19:
```suggestion
    predefined aliases. Many features are stoichiometrically weighted generalized means of element-level properties.
```

src/citrine/informatics/predictors/chemical_formula_featurizer.py
```diff
@@ -0,0 +1,162 @@
...
+    How to compute the mean is configured using the ``powers`` argument.
+
+    The default is the "standard" alias, corresponding to features that are intuitive and
```
maxhutch, line 22: What is in standard?

src/citrine/informatics/predictors/chemical_formula_featurizer.py
```diff
@@ -0,0 +1,162 @@
...
+                 input_descriptor: ChemicalFormulaDescriptor,
+                 features: Optional[List[str]] = None,
+                 excludes: Optional[List[str]] = None,
+                 powers: Optional[List[int]] = None):
+        self.name = name
+        self.description = description
+        self.input_descriptor = input_descriptor
+        self.features = features if features is not None else ["standard"]
+        self.excludes = excludes if excludes is not None else []
+        self.powers = powers if powers is not None else [2]
```
maxhutch, line 155: I think we actually want the default to be 1; That's what the SimpleML predictor has been doing.
```suggestion
        self.powers = powers if powers is not None else [1]
```

maxhutch (commented):
If we aren't starting with alpha, then we should also include something in the user guide.  It could very closely copy the example from molecular structure features.

maxhutch (approved):

PR #585, opened 2021-04-22
PLA-7068 set list flexibility

talltea:
# Citrine Python PR

## Description 
* Modify Set and List so that they both accept set/list/tuple as the underlying type.
* Made the _set_elements() method recursive for `PropertyCollection`s
 * Changed some things from Set to List or vice-versa, but it’s non-breaking because the setter code is flexible
 * Removed some unnecessary warnings and related tests
 * Make `properties.Optional`, `SpecifiedMixedList`, `LinkOrElse`, and `Object` a PropertyCollection and have them set recursively
* Fixed various tests that were inconsistent with their required types

### PR Type:


maxhutch:
Just needs a version bump.

src/citrine/informatics/design_spaces/formulation_design_space.py
```diff
@@ -48,7 +48,7 @@ class FormulationDesignSpace(Resource['FormulationDesignSpace'], DesignSpace):
         properties.String,
         properties.Set(properties.String)
     ), 'config.labels')
-    constraints = properties.Set(properties.Object(Constraint), 'config.constraints')
+    constraints = properties.List(properties.Object(Constraint), 'config.constraints')
```
bfolie, line 51: This should be a `Set`, not a `List`. I was messing around and changed it back to the wrong thing.

tests/_serialization/test_taurus_interop.py
```diff
@@ -15,7 +15,7 @@ def test_flatten():
     bounds = CategoricalBounds(categories=["foo", "bar"])
     template = ProcessTemplate(
         "spam",
-        conditions=[(ConditionTemplate(name="eggs", bounds=bounds), bounds)]
+        conditions=[[ConditionTemplate(name="eggs", bounds=bounds), bounds]]
```
bfolie, line 18: What happens if this is a tuple instead of a list?

tests/serialization/test_object_template.py
```diff
@@ -20,20 +20,20 @@ def test_object_template_serde():
     color_template = PropertyTemplate("Color", bounds=CategoricalBounds(["red", "green", "blue"]))
     # Properties are a mixture of property templates and [template, bounds], pairs
     block_template = MaterialTemplate("Block", properties=[[length_template, sub_bounds],
-                                                           color_template])
+                                                           [color_template]])
```
bfolie, line 23: I don't think this is a desirable change--the user should be able to specify the template on its own, without wrapping it in brackets

tests/_serialization/test_taurus_interop.py
talltea: good catch, it's a leftover debug thing, putting back

src/citrine/informatics/design_spaces/formulation_design_space.py
```diff
@@ -63,15 +63,15 @@ def __init__(self, *,
                  description: str,
                  formulation_descriptor: FormulationDescriptor,
                  ingredients: Set[str],
-                 constraints: Set[Constraint],
+                 constraints: List[Constraint],
```
bfolie, line 66: This should still be a Set

src/citrine/informatics/design_spaces/formulation_design_space.py
```diff
@@ -63,15 +63,15 @@ def __init__(self, *,
...
                  labels: Optional[Mapping[str, Set[str]]] = None,
                  resolution: float = 0.0001,
                  session: Session = Session()):
         self.name: str = name
         self.description: str = description
         self.formulation_descriptor: FormulationDescriptor = formulation_descriptor
         self.ingredients: Set[str] = ingredients
-        self.constraints: Set[Constraint] = constraints
+        self.constraints: List[Constraint] = constraints
```
bfolie, line 74: This should still be a Set

bfolie (commented):
Except for a small `List -> Set` change, I think this is good. But since I was closely involved in the work, someone else should review as well.

src/citrine/informatics/design_spaces/formulation_design_space.py
maxhutch: Why, out of curiosity?

bfolie: Order doesn't matter. But also, it was a Set before, so we're not changing anything. I changed it to List as part of some messing around.

maxhutch (approved):
LGTM, and I confirmed this work with the e2e tests.

kroenlein (commented):
It's too late on Friday to be effectively comprehensive.  Sorry.

tests/resources/test_object_setters.py
```diff
@@ -59,7 +59,7 @@ def test_object_validation():
 
 def test_list_validation():
     """Test that lists are validated by gemd."""
-    with pytest.raises(TypeError):
+    with pytest.raises(ValueError):
```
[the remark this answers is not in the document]
talltea: I totally agree with you, this should be a TypeError.
However, because we're now going through every element in the list and trying to deserialize according to the underlying type, this error is now coming from here: https://github.com/CitrineInformatics/citrine-python/blob/PLA-7068/set-list-flexibility/src/citrine/_serialization/properties.py#L71
Which raises a ValueError with the message `'{} is not one of valid types:`. Which is wrong. I opted not to fix this in this PR cause it affects a bunch of other tests. We can fix it in a follow-up if that's okay?

src/citrine/_serialization/properties.py
```diff
@@ -360,7 +360,7 @@ def __init__(self,
 
     @property
     def underlying_types(self):
-        return list
+        return list, set, tuple
```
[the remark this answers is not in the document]
talltea: I'd be open to it, (and tests pass with it), but I think I'd lean towards being more explicit/constrictive? We can always expand later if we find we want more flexibility.

tests/resources/test_object_setters.py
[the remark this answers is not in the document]
kroenlein: v. reasonable

src/citrine/_serialization/properties.py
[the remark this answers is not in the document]
kroenlein: I'm having trouble tracking down the reference now, but I recall a discussion on the Python boards that made the point that duck-typing implies you should be testing for method compatibility (e.g., what `collections.abc` offers) rather than literal object inheritance.  Not blocking.

kroenlein (approved):

talltea:
tested against dev: https://jenkins.corp.citrine.io/job/nextgen-ITs/15795/

PR #590, opened 2021-04-24
Pla 7143/mark methods not implemented

bfolie:
# Citrine Python PR

## Description 
For collection methods get/update/register/list/delete, if the method does not work and is not overridden, override it with a NotImplementedError (if it might be implemented one day) or a RuntimeError (if it does not make sense to implement).

Implement `archive()` method for module collections. To simplify the code, a new `AbstractModuleCollection` is created.

### PR Type:


tests/resources/test_predictor.py
```diff
@@ -37,6 +37,29 @@ def test_build(valid_simple_ml_predictor_data, basic_predictor_report_data):
...
+        return data
+    session.put_resource.side_effect = _mock_put_resource
+    archived_predictor = pc.archive(uuid.uuid4())
+    assert archived_predictor.archived
+
+    session.get_resource.side_effect = NotFound("")
+    with pytest.raises(RuntimeError):
+        pc.archive(uuid.uuid4())
+
+
```
kroenlein, line 62: Only testing the predictor b/c the AbstractModuleCollection methods will get exercised through that pathway?  Is there a reason not to just make that module a parent class to all resources?

src/citrine/informatics/processors.py
```diff
@@ -68,7 +68,7 @@ class GridProcessor(Serializable['GridProcessor'], Processor):
         properties.Integer,
         'config.grid_dimensions'
     )
-    status = properties.String('status', serializable=False)
+    status = properties.Optional(properties.String(), 'status', serializable=False)
```
kroenlein, line 71: Is the None option tested anywhere?

kroenlein (approved):
Some non-blocking questions about approach.

kroenlein (approved):
Non-blocking questions still stand.

tests/resources/test_predictor.py
bfolie: > Only testing the predictor b/c the AbstractModuleCollection methods will get exercised through that pathway?

Yes, and since we require 100% test coverage, we know it's being exercised thoroughly

> Is there a reason not to just make that module a parent class to all resources?

You mean make `AbstractModuleCollection` a parent to all resources? It's not relevant to all resource collections. This delete/archive behavior is particular to collections of modules

src/citrine/informatics/processors.py
bfolie: I bumped into it when testing the new `archive` method because the module is serialized (removing the `status` field) and then deserialized.

Having `status` be optional is the behavior of the other modules, so this was a bug. It's no longer relevant because #547 fixes the same bug.

PR #597, opened 2021-04-30
Codify and unify statuses

bfolie:
# Citrine Python PR
PLA-7260 and PLA-7261

## Description 
Present a unified interface for querying the status of long-running objects. This pattern applies to module validation, workflow validation, module report generation, and workflow execution. Kicking off the job creates an object which is missing information and has some status. The user must repeatedly query the object, and eventually the status will change and, if successful, the desired information will all be present.

All such objects now must expose boolean-valued methods `in_progress()`, `succeeded()`, and `failed()`. This avoids string-matching, as was being done previously in utility code. It also papers over the fact that we don't have a common set of statuses.

The outlier is table creation, which produces a `JobSubmissionResponse` object that indicates when the table object can be fetched.. Migrating that to an asynchronous pattern would require backend work, so we can't do anything about it right now. `JobStatusResponse` codifies the available statuses of table building.

### PR Type:


tests/conftest.py
```diff
@@ -180,7 +180,7 @@ def valid_simple_ml_predictor_data(valid_gem_data_source_dict):
     z = RealDescriptor("z", 0, 100, "")
     return dict(
         module_type='PREDICTOR',
-        status='VALID',
+        status='READY',
```
maxhutch, line 183: Why are we changing these?  I thought they do come back from the backend with `status='VALID'`.

tests/jobs/test_waiting.py
```diff
@@ -55,62 +59,21 @@ def test_wait_while_executing(sleep_mock):
...
 def test_wait_while_executing_predictor_evaluation_execution_missing_collection():
     predictor_evaluation_execution = mock.Mock(spec=PredictorEvaluationExecution)
 
     with pytest.raises(ValueError):
         wait_while_executing(predictor_evaluation_execution, print_status_info=True)
-
-
-@mock.patch('time.time')
-@mock.patch('time.sleep', return_value=None)
-def test_while_executing_timeout(sleep_mock, time_mock):
```
maxhutch, line 98: Why the removal?

tests/jobs/test_waiting.py
```diff
@@ -55,62 +59,21 @@ def test_wait_while_executing(sleep_mock):
...
+    ])
+    workflow_execution.status = statuses
 
     wait_while_executing(workflow_execution, print_status_info=True)
 
-    assert("Finished" in captured_output.getvalue())
-
-
-@mock.patch('time.sleep', return_value=None)
-def test_wait_while_executing_predictor_evaluation_execution(sleep_mock):
```
maxhutch, line 72: Why the removal?

maxhutch (commented):
In the past, these methods actually queried the backend to update the status.  I think a user might reasonable expect that.  Do we want to make sure sessions are threaded through the objects to support that fetch?  If not, I think we should at least over-communicate it in the docs.

tests/conftest.py
bfolie: Nope, it's 'READY'. You can check the json directly, but the easiest way to see it is to go the Predictors list in the UI

tests/jobs/test_waiting.py
bfolie: Because the logic of `wait_while_executing` and `wait_while_validating` was collapsed into a single function, this now tests the exact same code as `test_wait_while_validating`

tests/jobs/test_waiting.py
bfolie: Same as above

bfolie:
> In the past, these methods actually queried the backend to update the status. I think a user might reasonable expect that. Do we want to make sure sessions are threaded through the objects to support that fetch? If not, I think we should at least over-communicate it in the docs.

The only method that queried the backend was in `WorkflowExecution`, which is deprecated. And in a sense, `PredictorReport` does too, since `Predictor.report` queries the backend to get a fresh report, which brings a fresh status. But that's unchanged.

For module/workflow validation and design execution/predictor evaluation execution, I clarified in the docstring that the methods do not query for status. Most of them already had access to `session` and `project_id`, and I added it for those that didn't, so making the status methods query in the future will be easy to do.

There is an inconsistency in how those attributes are defined. In modules, they are underscored: `_session` and `_project_id`. In workflows and workflow executions they are not. We'll have to address that later, when we make the breaking changes of taking non-user-settable attributes out of __init__methods.

maxhutch (dismissed):

maxhutch:
@talltea I reviewed everything prior to my changes.  There was one minor issues that the e2es surfaced that I addressed.  Could you review the balance?  e2es are passing.

talltea:
> @talltea I reviewed everything prior to my changes. There was one minor issues that the e2es surfaced that I addressed. Could you review the balance? e2es are passing.

We don't have old design workflows in e2es anymore, but I'm guessing that they would fail for similar reasons as the performance workflows

maxhutch:
> > @talltea I reviewed everything prior to my changes. There was one minor issues that the e2es surfaced that I addressed. Could you review the balance? e2es are passing.
> 
> We don't have old design workflows in e2es anymore, but I'm guessing that they would fail for similar reasons as the performance workflows

I'd be inclined to just roll forward on those (and I would have said the same of PerformanceWorkflows if they weren't still in our e2es).  The fix for design workflows would be a bit more complicated since we didn't create a new class for them (vs PerformanceWorkflow -> PredictorEvaluationWorkflow).  We'd be adding cruft to non-deprecated class.

talltea (approved):
lgtm

PR #599, opened 2021-05-03
enumerated resource type

bfolie:
# Citrine Python PR
PLA-7249

## Description 
Resources have a `resource_type` field which is a string that defaults to the uppercase class name. This is not correct in general*. Instead, define an Enum with the allowed types, make a protected attribute `_resource_type`, and define it for all relevant resources.

*The allowed values are set by the backend. This value is used during changes in resource access permissions.

### PR Type:


src/citrine/informatics/design_spaces/data_source_design_space.py
```diff
@@ -25,6 +25,7 @@ class DataSourceDesignSpace(Resource['DataSourceDesignSpace'], DesignSpace, AIRe
     """
 
     _response_key = None
+    _resource_type = ResourceTypeEnum.MODULE
```
bfolie, line 28: Because `Resource` is extended at the level of individual modules, this line needs to be duplicated for each one. Ugh

src/citrine/_rest/resource.py
```diff
@@ -9,14 +31,11 @@ class Resource(Serializable[Self]):
 
     _path_template: str = NotImplemented
     _response_key: Optional[str] = None
-
-    def resource_type(self) -> str:
```
bfolie, line 13: I feel OK deleting this method because it's on an internal class.

Not sure if this merits a minor version bump or if it's a patch because the change is entirely internal.

maxhutch: I'd go ahead and use a minor bump since it wasn't `_protected`.

maxhutch (approved):
LGTM.  Can we test this with our e2es before merging for extra coverage, though?

bfolie:
> LGTM. Can we test this with our e2es before merging for extra coverage, though?

Finally finished (and passed): https://jenkins.corp.citrine.io/job/Devkit-custom/441/

PR #601, opened 2021-05-17
Rename IngredientsToSimpleMixturePredictor to IngredientsToFormulationPredictor

bfolie:
# Citrine Python PR

## Description 
This is a misnomer, it creates a formulation that may or may not be a simple mixture. Mark the predictor as deprecated and create a new one with identical behavior. Also clean up some doc language, where "simple mixture" was often erroneously used where "formulation" was meant

Corresponding backend change: https://github.com/CitrineInformatics/mithril/pull/700

### PR Type:


src/citrine/informatics/predictors/ingredients_to_simple_mixture_predictor.py
```diff
@@ -11,7 +12,15 @@
 
 class IngredientsToSimpleMixturePredictor(
         Resource['IngredientsToSimpleMixturePredictor'], Predictor, AIResourceMetadata):
-    """[ALPHA] A predictor interface that constructs a simple mixture from ingredient quantities.
+    """[DEPRECATED] [ALPHA] Constructs a simple mixture from ingredient quantities.
```
latture, line 15:
```suggestion
    """[DEPRECATED] Constructs a simple mixture from ingredient quantities.
```

src/citrine/informatics/predictors/ingredients_to_simple_mixture_predictor.py
```diff
@@ -50,6 +59,11 @@ def __init__(self,
                  id_to_quantity: Mapping[str, RealDescriptor],
                  labels: Mapping[str, Set[str]],
                  archived: bool = False):
+        warn("{this_class} has been renamed. Please use {replacement} instead"
+             .format(this_class=self.__class__.__name__,
+                     replacement="Ingredients To Formulation Predictor"),
```
latture, line 64:
```suggestion
                     replacement=IngredientsToFormulationPredictor.__name__),
```
But you'll need to replace line 7 with
```
from citrine.informatics.predictors import Predictor, IngredientsToFormulationPredictor
```

latture (approved):
Only a few very minor suggestions, but those aren't blocking.

It's a bit strange that we'll deserialize deprecated predictors into the new type. There's a very small chance that could cause issues with client code (I think only if they're checking `isinstance`) or if we change the new predictor in the future such that we can't deserialize the old types. That risk is pretty small though.

src/citrine/informatics/predictors/ingredients_to_simple_mixture_predictor.py
bfolie: Importing `IngredientsToFormulationPredictor` causes a bunch of tests to fail. I don't know why

bfolie:
> It's a bit strange that we'll deserialize deprecated predictors into the new type. There's a very small chance that could cause issues with client code (I think only if they're checking `isinstance`) or if we change the new predictor in the future such that we can't deserialize the old types. That risk is pretty small though.

If we change the predictor itself, I think we'd have to go through the same hoops that we'd go through if we hadn't done this name change. But I agree that any client code which explicitly checks for the `IngredientsToSimpleMixturePredictor` type will break.

It seems unlikely enough that I feel comfortable making use of the fact that we're still in pre-release to make the breaking change. I'll send an announcement to #ds-de-prod

latture (approved):

PR #602, opened 2021-05-18
Tables-related clarifications

bfolie:
# Citrine Python PR

## Description 
PLA-7379. Several miscellaneous clarifications and improvements related to Gem Tables.

### PR Type:


setup.py
```diff
@@ -1,6 +1,4 @@
 from setuptools import setup, find_packages
-import sys
-from os.path import join
```
bfolie, line 3: Not related to tables, but I happened to notice the unnecessary imports

src/citrine/gemtables/columns.py
```diff
@@ -283,7 +283,7 @@ def _attrs(self) -> List[str]:
     def __init__(self, *,
                  data_source: str,
                  component_name: str,
-                 normalize: Optional[bool] = False):
+                 normalize: bool = False):
```
bfolie, line 286: This is technically a breaking change, since if someone has a script in which they define a `ComponentQuantityColumn` and set `normalize=None`, it will no longer serialize. But I strongly doubt anyone has done that.

src/citrine/resources/table_config.py
```diff
@@ -22,8 +22,6 @@
 from citrine.gemtables.variables import Variable, IngredientIdentifierByProcessTemplateAndName, \
     IngredientQuantityByProcessAndName, IngredientQuantityDimension
 
-logger = getLogger(__name__)
-
```
bfolie, line 26: The oft-quoted distinction between logger and warnings is:
> warnings.warn() in library code if the issue is avoidable and the client application should be modified to eliminate the warning
> logging.warning() if there is nothing the client application can do about the situation, but the event should still be noted

Since the consumer of citrine-python should modify their code to eliminate the warning, I think we want to emit warnings.warn()

src/citrine/gemtables/columns.py
kroenlein: Why are we getting rid of the Optional?  It matches the definition in [backend](https://github.com/CitrineInformatics/platform-backend/blob/590ea23c052ae2600aab43e39c7ea9edd4c49fdd/code/gemtables/model/src/main/scala/io/citrine/gemtables/model/column/CompositionColumnDefinition.scala#L126) and are keeping the default.  Am I missing the purpose of the Optional typehint?

src/citrine/resources/table_config.py
```diff
@@ -349,20 +347,22 @@ def default_for_material(
...
+            warn("\'scope\' is deprecated in default_for_material() and will soon be removed. "
+                 "To specify a custom scope, use a LinkByUID object.", DeprecationWarning)
         if isinstance(material, MaterialRun):
-            if scope is not None:
-                logger.warning(
-                    'Ignoring scope {} since material run object was specified.'.format(scope))
             uid_tup = next(iter(material.uids.items()), None)
             if uid_tup is None:
                 raise ValueError('Material must have a uid to build default table config.')
             scope, uid = uid_tup
```
kroenlein, line 357:
```suggestion
            link = LinkByUID.from_entity(material, CITRINE_SCOPE)
            scope, uid = material.scope, material.id
```
would make more sense, though that fails to catch the `there's no id` scenario.  Looping through internal structs always strikes me as not-great.

kroenlein (approved):
Minor things that seemed a little weird, but nothing I'm too bothered by.

src/citrine/gemtables/columns.py
bfolie: Optional means you can provide `None` and in doing so convey information. What does `None` mean in this context? Either you want to normalize or you don't. It's true that, strictly speaking, you can provide `None` and the backend fills in the default value of `False`. But that's not how `Optional` is generally used in the python client. For example, in listing calls, the `page` is optional. Not specifying a page means something different from specifying a page.

The definition on the backend isn't optional either. It's `boolean`, not `Some[boolean]`.  Having a default is not the same as being optional.

src/citrine/resources/table_config.py
bfolie: `LinkByUID.from_entity` has non-obvious behavior in that it fills in an ID if none exists with that scope (thereby missing the the scenario you mention here). We could define a different version of `from_entity` that behaves as we need it to, but we might as well create the desired behavior here.

> Looping through internal structs always strikes me as not-great.

I don't see what the issue is in this case. We just need scope-id pair, and any will work. It's not going to fail in unexpected ways. If there are performance reasons to prefer `CITRINE_SCOPE`,  we could check for that first.

kroenlein: > I don't see what the issue is in this case.

The pattern is an issue if we change the implementation of uids away from a simple dict.  Not blocking.

kroenlein (approved):

src/citrine/gemtables/columns.py
kroenlein: Ahh, yes, that tracks.  We don't need ternary logic.

PR #603, opened 2021-05-19
Sphinx docs are more legible and more complete

bfolie:
# Citrine Python PR

## Description 
PLA-7159, PLA-7158, PLA-6760

- Bumps sphinx to latest version and turns on the `inherited-members` option, which causes child classes to show methods and attributes inherited from parent classes
- As a proof of concept, adds docstrings to the attributes in `AIResourceMetadata`. These show up in the docs for all inheriting classes (modules and workflows)
- Change theme from "albaster" to "readthedocs", which looks way better.
- Create custom css file which formats method signatures so that each argument is on its own line.

Old docs:
image
 
New docs:
image
image
(and there's a bit more).

Drawbacks of this approach:

- For attributes specified through our serde code, the default value is filled in. For example, it says `experimental = True` because `True` is the default, but this is misleading.
- Method parameters appear twice, once as a parameter and once as a serde attribute. For example, `training_data` is in the list of parameters at the top, but there's also an entry `training_data: List[DataSource]= []`. This could be misleading
- The title can sometimes overfill the box
- The full class path is sometimes written out unnecessarily. For example, it would be easier to read if `get_type` was said to return `Type[Predictor]` instead of `Type[citrine.informatics.predictors.predictor.Predictor]`

### PR Type:


talltea:
I was having trouble making the docs locally, but with the change I pushed to the makefile, it works for me

talltea (dismissed):
this looks 💯 , thanks for doing this @bfolie!

talltea (approved):

PR #604, opened 2021-05-19
Adding missing numpy dependency

ventura-rivera:
# Citrine Python PR

## Description 
I am getting a `ModuleNotFoundError` because numpy doesn't exist as a dependency in the `setup.py` file. This has been overlooked in the tests because the `test_requirements.txt` file includes pandas (which will also install numpy), so you will not run into the issue in the automated testing. However installing citrine-python from pypi will not include numpy.

Numpy is used in [table configs here](https://github.com/CitrineInformatics/citrine-python/blob/59845fd967d9d5b2cbb8cd90a5738ca6cc634cdd/src/citrine/resources/table_config.py#L5) and [design spaces here](https://github.com/CitrineInformatics/citrine-python/blob/621cf4f8907f1edb7c523834759562b665bed868/src/citrine/builders/design_spaces.py#L14)

### PR Type:


setup.py
```diff
@@ -30,7 +30,8 @@
           "boto3>=1.9.226,<2",
           "botocore>=1.12.226,<2",
           "deprecation>=2.0.7,<3",
-          "urllib3>=1.25.7,<2"
+          "urllib3>=1.25.7,<2",
+          "numpy>=1.19.5,<2"
```
ventura-rivera, line 34: This might be tricky, current version of numpy requires using Python version >=3.7

bfolie: 1.19.5 is compatible with python 3.6, so I think we're ok there

bfolie (commented):
Could you also add numpy to `requirements.txt`?

ventura-rivera:
> Could you also add numpy to `requirements.txt`?

@bfolie actually, now that I've investigated a little more, I think I will need to use pandas as a dependency instead due to [this line](https://github.com/CitrineInformatics/citrine-python/blob/621cf4f8907f1edb7c523834759562b665bed868/src/citrine/builders/design_spaces.py#L10)

setup.py
```diff
@@ -30,7 +30,8 @@
           "boto3>=1.9.226,<2",
           "botocore>=1.12.226,<2",
           "deprecation>=2.0.7,<3",
-          "urllib3>=1.25.7,<2"
+          "urllib3>=1.25.7,<2",
+          "pandas>=1.0.3,<2"
```
ventura-rivera, line 34: This line now seems redundant with the `extra_requires` piece.

bfolie:
Builders is non-core, so we shouldn't use that as a justification to add dependencies to `requirements.txt` and `setup.py`.

The only non-builders code that uses pandas or numpy is https://github.com/CitrineInformatics/citrine-python/blob/59845fd967d9d5b2cbb8cd90a5738ca6cc634cdd/src/citrine/resources/table_config.py#L286 and that can be easily changed to not use numpy. I put it in out of laziness, but it turns out that python has a built-in argmax function (https://flothesof.github.io/python-missing-argmax.html). So I think the better change would be to remove our dependence on numpy there.

ventura-rivera:
> So I think the better change would be to remove our dependence on numpy there.

Totally agree. Can I close this?

ventura-rivera:
> Totally agree. Can I close this?

Actually I can revert my changes and implement native argmax-like functionality.

src/citrine/resources/table_config.py
```diff
@@ -283,7 +282,7 @@ def get(self, uid: Union[UUID, str], version: Optional[int] = None):
             path = self._get_path(uid)
             data = self.session.get_resource(path)
             version_numbers = [version_data['version_number'] for version_data in data['versions']]
-            index = np.argmax(version_numbers)
+            index = version_numbers.index(max(version_numbers))
```
ventura-rivera, line 285: I am assuming `version_numbers` is a list of unique integers here

bfolie: yes, that's correct

bfolie (approved):

ventura-rivera:
@bfolie due to the recent branch protection protocols enforced, I do not have permission to merge this branch with master. Are you able to do it?

PR #607, opened 2021-05-20
Add docstrings for all attributes

bfolie:
# Citrine Python PR
PLA-6762

## Description 
Place attribute docstrings directly below the attributes, instead of as part of the class docstring, for easier-to-read docs. Putting the information in the class docstring under Attributes caused several legibility problems. First, this information would not be iniherited (child classes wouldn't have an entry in their docs). Second, the attribute would appear twice in the docs, one time with an empty entry.

Some of the attributes simply were not documented, and now they are.

Finally, some of the attributes were mistakenly filed under parameters. In some cases, such as `ModelSummary`, there is an \_\_init\_\_ method, but it's never meant to be used and will be removed in a subsequent PR (the objects are deserialized from backend responses, not created by the user).

### PR Type:


src/citrine/informatics/design_candidate.py
```diff
@@ -45,7 +45,9 @@ class MeanAndStd(Serializable["MeanAndStd"], DesignVariable):
     """
 
     mean = properties.Float('m')
+    """:float: mean of the continuous distribution"""
```
talltea, line 48: in general, I am used to the comments going above the line they are commenting , is there a reason all of these are below them?

src/citrine/informatics/workflows/performance_workflow.py
```diff
@@ -25,7 +25,6 @@ class PerformanceWorkflow(Resource['PerformanceWorkflow'], Workflow):
 
     """
 
-    uid = properties.Optional(properties.UUID, 'id', serializable=False)
```
talltea, line 28: why can we remove this?

src/citrine/resources/gemtables.py
```diff
@@ -26,25 +26,19 @@ class GemTable(Resource['Table']):
...
-        Url pointing to the location of the GEM Table's contents.
-        This is an expiring download link and is not unique.
-
     """
 
     _response_key = 'table'
     _resource_type = ResourceTypeEnum.TABLE
 
     uid = properties.Optional(properties.UUID(), 'id')
+    """:Optional[str]: unique Citrine id of this GEM Table"""
```
talltea, line 35: should this be `Optional[UUID]`?

src/citrine/resources/job.py
```diff
@@ -77,25 +63,16 @@ class JobStatusResponse(Resource['JobStatusResponse']):
...
-        job output properties and results
-
     """
 
     job_type = properties.String("job_type")
+    """:str: the type of job for this status report"""
     status = properties.String("status")
+    """:str: The status of the job. One of "Running", "Success", or "Failure"."""
     tasks = properties.List(Object(TaskNode), "tasks")
+    """:str: all of the constituent task required to complete this job"""
```
talltea, line 73: `List[TaskNode]` here instead of `str`?

talltea (approved):
👍 Thanks for this! lgtm

PR #611, opened 2021-05-21
Data object methods have flexible signatures

bfolie:
# Citrine Python PR PLA-7406

## Description 
When providing the identifying information for a GEMD object, be consistent. The user can provide a UUID, a string-version of a UUID, a LinkByUID, or an object with at least one scope-id pair. Many of these methods have optional arguments to specify a custom scope. These arguments are deprecated and will be removed in the future.

### PR Type:


docs/source/formulations_example.rst
```diff
@@ -302,7 +302,7 @@ The code below defines the rows and defines one column that contains the identif
         datasets=[dataset.uid],
         rows=[
             MaterialRunByTemplate(
-                templates=[LinkByUID.from_entity(t) for t in material_templates_to_include]
+                templates=[LinkByUID.from_entity(t, scope) for t in material_templates_to_include]
```
kroenlein, line 305: This is a correct fix, but it raises the question of does it make sense for citrine-python to replace the default scope on import.  That might be more dark-art/monkey-patch than you feel comfortable with.

src/citrine/resources/data_concepts.py
```diff
@@ -496,15 +518,17 @@ def poll_async_update_job(self, job_id: UUID, *, timeout: float = 2 * 60,
         # That worked, nothing returned in this case
         return None
 
-    def get(self, uid: Union[UUID, str], scope: str = CITRINE_SCOPE) -> ResourceType:
+    def get(self, uid: Union[UUID, str, LinkByUID, BaseEntity],
```
kroenlein, line 521: We should probably be implementing this as a custom type so that we don't have this complex pattern all over the place.

src/citrine/resources/data_concepts.py
```diff
@@ -702,28 +728,32 @@ def list_by_tag(self, tag: str, per_page: int = 100) -> Iterator[ResourceType]:
...
+            [DEPRECATED] use a LinkByUID to specify a custom scope
             The scope of the uid, defaults to Citrine scope (CITRINE_SCOPE)
         dry_run: bool
             Whether to actually delete the item or run a dry run of the delete operation.
             Dry run is intended to be used for validation. Default: false
 
         """
-        path = self._get_path() + "/{}/{}".format(scope, uid)
+        link = _make_link_by_uid(uid)
+        path = self._get_path() + "/{}/{}".format(link.scope, link.id)
```
kroenlein, line 749: This code is subject to issues if the scope or id contain [URI Reserved Characters](https://en.wikipedia.org/wiki/Percent-encoding).  I don't know if the backend is doing it properly, but we should be sending percent-encoded strings down the wire.

src/citrine/resources/data_concepts.py
```diff
@@ -702,28 +728,32 @@ def list_by_tag(self, tag: str, per_page: int = 100) -> Iterator[ResourceType]:
...
         params = {'dry_run': dry_run}
         self.session.delete_resource(path, params=params)
         return Response(status_code=200)  # delete succeeded
 
-    def _get_relation(self, relation: str, uid: Union[UUID, str], scope: str,
-                      forward: bool = True, per_page: int = 100) -> Iterator[ResourceType]:
+    def _get_relation(self, relation: str, uid: Union[UUID, str, LinkByUID, BaseEntity],
+                      scope: Optional[str] = None, forward: bool = True, per_page: int = 100
+                      ) -> Iterator[ResourceType]:
```
kroenlein, line 756: Heck of a signature.  Implement the breaking change of invalidating positional arguments?

src/citrine/resources/data_concepts.py
```diff
@@ -754,10 +784,11 @@ def _get_relation(self, relation: str, uid: Union[UUID, str], scope: str,
         params = {}
         if self.dataset_id is not None:
             params['dataset_id'] = str(self.dataset_id)
+        link = _make_link_by_uid(uid, scope)
         raw_objects = self.session.cursor_paged_resource(
             self.session.get_resource,
-            'projects/{}/{}/{}/{}/{}'.format(
-                self.project_id, relation, scope, uid, self._collection_key.replace('_', '-')),
+            'projects/{}/{}/{}/{}/{}'.format(self.project_id, relation, link.scope, link.id,
+                                             self._collection_key.replace('_', '-')),
```
kroenlein, line 791: Same percent-encoding problem as above.  Has been hit in the wild.

src/citrine/resources/material_run.py
```diff
@@ -109,7 +110,8 @@ def get_type(cls) -> Type[MaterialRun]:
         """Return the resource type in the collection."""
         return MaterialRun
 
-    def get_history(self, scope: str, id: Union[str, UUID]) -> Type[MaterialRun]:
+    def get_history(self, scope: Optional[str],
+                    id: Union[str, UUID, LinkByUID, MaterialRun]) -> Type[MaterialRun]:
```
kroenlein, line 114: This should _really_ ban positional arguments, since with the new call structure, the natural form is actually wrong.

src/citrine/resources/material_run.py
```diff
@@ -133,8 +136,9 @@ def get_history(self, scope: str, id: Union[str, UUID]) -> Type[MaterialRun]:
             of the object.
 
         """
+        link = _make_link_by_uid(id, scope)
         base_path = os.path.dirname(self._get_path(ignore_dataset=True))
-        path = base_path + "/material-history/{}/{}".format(scope, id)
+        path = base_path + "/material-history/{}/{}".format(link.scope, link.id)
```
kroenlein, line 141: Percent encoding, again.

src/citrine/resources/material_run.py
```diff
@@ -217,43 +226,48 @@ def filter_by_template(self,
         return self.list_by_template(uid=template_id, scope=template_scope)
 
     def list_by_spec(self,
-                     uid: Union[UUID, str],
-                     scope: str = CITRINE_SCOPE) -> Iterator[MaterialRun]:
+                     uid: Union[UUID, str, LinkByUID, GEMDMaterialSpec],
+                     scope: Optional[str] = None) -> Iterator[MaterialRun]:
```
kroenlein, line 230: Where this and the like are the arguments against having a unified definition.

src/citrine/resources/material_spec.py
```diff
@@ -9,10 +9,11 @@
 from citrine._serialization.properties import List as PropertyList
 from citrine._serialization.properties import Optional as PropertyOptional
 from citrine._serialization.properties import String, LinkOrElse, Mapping, Object
-from citrine.resources.data_concepts import DataConcepts, CITRINE_SCOPE
+from citrine.resources.data_concepts import DataConcepts, CITRINE_SCOPE, _make_link_by_uid
```
kroenlein, line 12: I think we can pretty consistently scrub the need to import `CITRINE_SCOPE`.

kroenlein (commented):
Several points for discussion and possible can-kicking.

src/citrine/resources/data_concepts.py
bfolie: I'm not sure I agree. It would be more compact, but the custom type would be an artifice that only serves to save space. I don't think it's bad to be explicit when saying "we accept any of these 4 types which are _not_ the same but carry equivalent information"

src/citrine/resources/data_concepts.py
bfolie: That issue seems like it's already present, and made neither worse nor better by this PR

src/citrine/resources/material_run.py
bfolie: I agree. Keyword-only args is one of the changes we'll be making this sprint as part of our "breaking v1 changes" work

src/citrine/resources/material_spec.py
bfolie: It's still used in filter_by_template, so it stays for now (but that method is deprecated and will be removed shortly).

If the import wasn't used then flake8 would catch that

docs/source/formulations_example.rst
bfolie: Not sure I follow. The templates are defined with this scope, so nothing is being modified

src/citrine/resources/data_concepts.py
kroenlein: I agree, but this is arguably something that should be addressed before a v1 release.  Customers have hit it when they incorporated `/` into their scopes.  Not blocking.

src/citrine/resources/data_concepts.py
kroenlein: Compactness carries with it resilience to maintenance bugs and ease of reading.  Given how we have a substantial number of methods that accept `UUID, str, LinkByUID` plus a subclass of `BaseEntity`, there's value is having the signatures read `Union[IdThing, BaseEntity]`, so you know what to expect about the variability.

src/citrine/resources/material_spec.py
kroenlein: But `filter_by_template` just calls `list_by_template`.  `scope: str = None` would just function to piggyback on `list_by_template`'s default.

bfolie: good point

kroenlein (approved):
A couple of things worthy of fixing, but fine for today.

src/citrine/resources/data_concepts.py
bfolie: It's coming.

Fortunately this is an internal method, so the damage is contained

PR #615, opened 2021-05-24
Revert "Revert "Data object methods have flexible signatures""

bfolie:
Reverts CitrineInformatics/citrine-python#614


bfolie:
I forgot to thread `scope` through two of the methods, and that caused the e2es to fail. Here's the test on jenkins (which I should have done before merging the first time): https://jenkins.corp.citrine.io/job/nextgen-ITs/16727/

kroenlein (dismissed):

kroenlein (approved):

PR #617, opened 2021-05-25
share resource directly

bfolie:
# Citrine Python PR
PLA-7248

## Description 
Share resources by passing the resource to be shared instead of the resource id and type, which isn't a user-facing concept.

Corresponding nextgen-devkit changes: https://github.com/CitrineInformatics/nextgen-devkit/pull/889

### PR Type:


src/citrine/resources/project.py
```diff
@@ -237,14 +238,44 @@ def ara_definitions(self) -> TableConfigCollection:  # pragma: no cover
...
+            The type of the resource to share. Must be one of DATASET, MODULE, USER,
+            TABLE, OR TABLE_DEFINITION
+        resource_id: Optional[str]
+            [DEPRECATED] Please use ``resource`` instead
+            The id of the resource to share
+
+        """
+        resource_dict = None
+        if resource is not None:
+            resource_dict = resource.as_entity_dict()
```
maxhutch, line 266: I think this `as_entity_dict` name is a bit opaque: I had to look up what that method did.  Could you add a comment here, or change the function name to something more self-documenting?

bfolie: I'll add a comment. Note that it's the same method used by `transfer_resource`, `make_public`, and `make_private`

bfolie: Decided to change the name instead

maxhutch (approved):

PR #620, opened 2021-05-27
Pla 7142/remove non user facing args

bfolie:
# Citrine Python PR
PLA-7142

## Description 
Remove all non-user-facing arguments from __init__ methods. Generally, these parameters relate to the connection to the platform and are set during deserialization. The user has no reason to be setting them directly, so while this is a large breaking change on its face, it is unlikely to require much change to user-written code.

### PR Type:


maxhutch (commented):
I think we probably want to keep the member declarations in the init method, even if they aren't exposed in the constructor.  For example, `self.session: Optional[Session] = None`.  Otherwise these members are really opaque.  

Alternatively, we could set them at the class level to take advantage of inheritance; the behavior when setting a class member on an instance is a bit advanced, but I think the clarity/usability advantage probably justifies it. Also, we've been using that pattern for the schema members.

bfolie:
> I think we probably want to keep the member declarations in the init method, even if they aren't exposed in the constructor. For example, `self.session: Optional[Session] = None`. Otherwise these members are really opaque.
> 
> Alternatively, we could set them at the class level to take advantage of inheritance; the behavior when setting a class member on an instance is a bit advanced, but I think the clarity/usability advantage probably justifies it. Also, we've been using that pattern for the schema members.

To keep using `session` as an example, in some cases it's completely unused. For example, in `Constraint`s. The user would never set it directly, it's not set during deserialization, and even if it were set it wouldn't be used. A `Constraint` doesn't need to know anything about connecting to the platform.

The modules and workflows are a bit trickier, because they do need to know the `Session`, but the user doesn't set it and IMO they shouldn't see it at all. In that case the pattern, already used in some places and here adopted widely, is to declare `_session` [as an attribute](https://github.com/CitrineInformatics/citrine-python/blob/80bc3f6f3134d502500e4ef10f506e02d23ff914/src/citrine/informatics/predictors/predictor.py#L24) and to [set it during building](https://github.com/CitrineInformatics/citrine-python/blob/80bc3f6f3134d502500e4ef10f506e02d23ff914/src/citrine/resources/predictor.py#L35).

I agree that it should be decipherable for a developer, even if it's hidden from the user. In this case, the predictor's session is [when getting the report](https://github.com/CitrineInformatics/citrine-python/blob/80bc3f6f3134d502500e4ef10f506e02d23ff914/src/citrine/informatics/predictors/predictor.py#L37). A developer can easily trace that back up the file to the attribute declaration, but might get lost there, so I think that a docstring on `_session` would be helpful.

src/citrine/informatics/workflows/workflow.py
```diff
@@ -21,7 +23,10 @@ class Workflow(PolymorphicSerializable['Workflow'], AsynchronousObject):
     """
 
     _response_key = None
+    _session: Optional[Session] = None
```
maxhutch, line 26: :+1:

src/citrine/resources/design_execution.py
```diff
@@ -22,15 +22,15 @@ class DesignExecution(Resource['DesignExecution'], Pageable, AsynchronousObject)
...
-    ----------
-    project_id: str
-        Unique identifier of the project that contains the workflow execution
-
     """
 
     _paginator: Paginator = Paginator()
     _collection_key = 'response'
+    # This should really be _session, but _fetch_page assumes there is a 'pageable' parameter
+    session: Optional[Session] = None
```
maxhutch, line 30: Why can't we rename this one?

maxhutch (commented):
I dunno if I just didn't scroll down far enough, but the `_session` pattern looks great.  There's one place where its `session` still (no `_`).  What would it take to rename that one?

src/citrine/resources/design_execution.py
bfolie: > There's one place where its `session` still (no `_`). What would it take to rename that one?

I renamed it. The `_fetch_page` method is more flexible than I realized initially

maxhutch (approved):
:100:

bfolie:
Passing e2e tests using this branch: https://jenkins.corp.citrine.io/job/nextgen-ITs/16985/

PR #622, opened 2021-05-28
Standardize predictor inputs/outputs

bfolie:
# Citrine Python PR
PLA-7244

## Description 
Ideally, the input(s) and output(s) of a predictor would always be `input` or `inputs` and `output` or `outputs`. Unfortunately, `input` is a built-in method in python. As a compromise, a single input is therefore called `input_descriptor`. This makes the `SimpleMixturePredictor` somewhat awkward, since it takes a single input and produces a single output. Having the arguments be `input_descriptor` and `output` seems torturous, so I decided to leave it as `output_descriptor`.

After proceeding through this logic, the only thing that I decided to change was that the `MolecularStructureFeaturizer`'s input should be `input_descriptor` instead of `descriptor`. This required the featurizer to enforce keyword args.

### PR Type:

### DIFF
--- a/src/citrine/builders/predictors.py
+++ b/src/citrine/builders/predictors.py
@@ -59,9 +59,7 @@ def build_mean_feature_property_predictors(
         as inputs to one or more ML models.
 
     """
-    if isinstance(featurizer, MolecularStructureFeaturizer):
-        input_descriptor = featurizer.descriptor
-    elif isinstance(featurizer, ChemicalFormulaFeaturizer):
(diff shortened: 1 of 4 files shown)


bfolie:
Nextgen ITs run with this branch of citrine-python and the [corresponding branch](https://github.com/CitrineInformatics/nextgen-devkit/pull/893) of nextgen-devkit: https://jenkins.corp.citrine.io/job/Devkit-custom/

talltea (approved):
bummer about input(), but this works!

PR #628, opened 2021-06-03
Pla 7141 kwargs

talltea:
# Citrine Python PR

## Description 
add kwargs where necessary

### PR Type:


src/citrine/resources/data_objects.py
```diff
@@ -151,7 +151,7 @@ def _get_attribute_bounds_search_body(attribute_bounds):
         }
 
     def validate_templates(self,
-                           model: DataObjectResourceType,
+                           model: DataObjectResourceType, *,
```
bfolie, line 154: I think we might want to put the asterisk before `model`, since in this case the name `validate_templates` doesn't have a clear, single argument that obviously goes first. And in general, the user will specify a model and some templates.

src/citrine/resources/descriptors.py
```diff
@@ -15,7 +15,7 @@ def __init__(self, project_id: UUID, session: Session):
         self.project_id = project_id
         self.session: Session = session
 
-    def from_predictor_responses(self, predictor: Predictor,
+    def from_predictor_responses(self, predictor: Predictor, *,
```
bfolie, line 18: Both the predictor and the inputs are equally important parts of the function call, so I think they should both require keywords

src/citrine/resources/file_link.py
```diff
@@ -405,7 +405,7 @@ def _complete_upload(self, dest_name: str, uploader: _Uploader):
         url = self._get_path(file_id) + '/versions/{}'.format(version)
         return FileLink(filename=dest_name, url=url)
 
-    def download(self, file_link: FileLink, local_path: str):
+    def download(self, file_link: FileLink, *, local_path: str):
```
bfolie, line 408: Both arguments are used and one is not necessary primary, so I would make them both keyword-args

src/citrine/resources/file_link.py
```diff
@@ -229,7 +229,7 @@ def _as_dict_from_resource(self, file: dict):
         }
         return file_dict
 
-    def upload(self, file_path: str, dest_name: str = None) -> FileLink:
+    def upload(self, file_path: str, *, dest_name: str = None) -> FileLink:
```
bfolie, line 232: It's true that the user can get away without specifying dest_name, but they'll often want to specify it. I think that both arguments are important enough that it doesn't make sense to elevate `file_path`

src/citrine/resources/gemtables.py
```diff
@@ -285,7 +287,7 @@ def delete(self, uid: Union[UUID, str]):
         """Tables cannot be deleted at this time."""
         raise NotImplementedError("Tables cannot be deleted at this time.")
 
-    def read(self, table: Union[GemTable, Tuple[str, int]], local_path: str):
+    def read(self, table: Union[GemTable, Tuple[str, int]],  *, local_path: str):
```
bfolie, line 290: I don't have strong feelings about this one, but since both arguments are equally useful I err towards making them both be keyword args

src/citrine/resources/table_config.py
```diff
@@ -349,8 +349,9 @@ def default_for_material(
         ambiguous = [(Variable.build(v), Column.build(c)) for v, c in data['ambiguous']]
         return config, ambiguous
 
-    def preview(self, *,
+    def preview(self,
                 table_config: TableConfig,
+                *,
```
bfolie, line 354: Once the deprecated `preview_roots` is removed, then this method will have two necessary and equally-important arguments, so they should both be kwargs

src/citrine/resources/descriptors.py
talltea: That makes sense. I was thinking that because predictor was in the method name, it would be redundant, but I think I'd prefer to err on the side of enforcing kwargs.
I'll also add it to https://github.com/CitrineInformatics/citrine-python/pull/628/files#diff-5fb493c283c9b66db198bd69f057b8b7bef132929d0243d8e2a5be268f6da418R46

src/citrine/resources/design_space.py
```diff
@@ -64,7 +64,7 @@ def update(self, model: DesignSpace) -> DesignSpace:
         self.validate_write_request(model)
         return AbstractModuleCollection.update(self, model)
 
-    def create_default(self, predictor_id: UUID) -> DesignSpace:
+    def create_default(self, *, predictor_id: UUID) -> DesignSpace:
```
talltea, line 67: I wasn't sure about this one, but erred on the side of enforcing kwargs

src/citrine/resources/gemtables.py
talltea: I was thinking of having this one and the `initiate_build` match the other methods here like `delete`, `update`, etc

src/citrine/resources/material_run.py
```diff
@@ -108,7 +108,7 @@ def get_type(cls) -> Type[MaterialRun]:
         """Return the resource type in the collection."""
         return MaterialRun
 
-    def get_history(self, scope: Optional[str],
+    def get_history(self, *, scope: Optional[str],
```
talltea, line 111: wasn't sure about this one

src/citrine/resources/predictor_evaluation_execution.py
```diff
@@ -114,7 +114,7 @@ class PredictorEvaluationExecutionCollection(Collection["PredictorEvaluationExec
     _collection_key = 'response'
     _resource = PredictorEvaluationExecution
 
-    def __init__(self, *,
+    def __init__(self,
```
talltea, line 117: I was in general not sure if we want kwargs for Collections, since those aren't created by users generally

src/citrine/resources/material_run.py
bfolie: If the only argument were `id` then I'd support not requiring the keyword. But since `scope`, which is deprecated, was the first argument, if we put `id` first and didn't require a keyword then we'd create unexpected failures. So I think the way you've done it here is what we have to do

src/citrine/resources/predictor_evaluation_execution.py
bfolie: I think not, since they're internal

src/citrine/informatics/scores.py
```diff
@@ -67,8 +67,8 @@ class LIScore(Serializable['LIScore'], Score):
     constraints = properties.List(properties.Object(Constraint), 'constraints')
     typ = properties.String('type', default='MLI')
 
-    def __init__(self, *,
-                 name: Optional[str] = None,
+    def __init__(self,
+                 name: Optional[str] = None, *,
```
bfolie, line 71: `name` is actually deprecated for scores, so I don't think we should privilege it in this way

src/citrine/resources/gemtables.py
bfolie: I agree on `initiate_build`. I suppose they're similar from the user's perspective. The difference I see is that `read` requires both arguments, while `initiate_build` generally requires only one. But it can go either way.

Related, this method is missing a full docstring (with parameters and a return value). Would you mind adding that?

src/citrine/resources/file_link.py
talltea: changing it for the other methods in this file that weren't consistent too

src/citrine/resources/gemtables.py
```diff
@@ -287,16 +287,25 @@ def delete(self, uid: Union[UUID, str]):
...
         """
         Read the Table file from S3.
 
         If a Table object is not provided, retrieve it using the provided table and version ids.
+
+
+        Parameters
+        ----------
+        table:
+            The persisted table config from which to build a table (or its ID).
```
bfolie, line 300:
```suggestion
            The persisted table config from which to build a table (or its ID and version number).
```

src/citrine/jobs/waiting.py
```diff
@@ -27,6 +27,7 @@ def _print_string_status(
 
 def wait_for_asynchronous_object(
     resource: AsynchronousObject,
+    *,
```
bfolie, line 30: to match the other two waiting methods, `resource` should be a kwarg

src/citrine/resources/predictor.py
```diff
@@ -36,7 +36,7 @@ def build(self, data: dict) -> Predictor:
         predictor._project_id = self.project_id
         return predictor
 
-    def check_for_update(self, predictor_id: UUID) -> Optional[Predictor]:
+    def check_for_update(self, uid: UUID) -> Optional[Predictor]:
```
bfolie, line 39: change back to predictor_id for now

bfolie:
I did a scan, and found the following. Otherwise, we're just about ready to go.

* `PlatformVocabulary.from_templates()` should require keyword-args
* `PlatformVocabulary.__init__()` should require keyword-args
* I'm on the fence about the methods in `seeding/find_or_create.py`. They should require kwargs. But I think that they're mostly used internally. But they are publicly facing.

talltea:
> I did a scan, and found the following. Otherwise, we're just about ready to go.
> 
> * `PlatformVocabulary.from_templates()` should require keyword-args
> * `PlatformVocabulary.__init__()` should require keyword-args
> * I'm on the fence about the methods in `seeding/find_or_create.py`. They should require kwargs. But I think that they're mostly used internally. But they are publicly facing.

got those, I had just missed the `seeding` folder entirely :( so i went through and added kwargs for all of those

bfolie:
Running e2es against the [corresponding version](https://github.com/CitrineInformatics/nextgen-devkit/pull/896) of nextgen-devkit: https://jenkins.corp.citrine.io/job/Devkit-custom/547/

bfolie (approved):
Looks good. Just gotta make sure the e2e tests pass

PR #635, opened 2021-06-08
Migrate thing_id to uid where appropriate

bfolie:
Argument "thing_id" should be reserved for cases where the thing's type is _not_ the same as the type of the calling object. There were a few cases where that pattern was not being observed.

In creating a utility method to facilitate the migration, I realized it could also be applied in a few other cases.

# Citrine Python PR

## Description 

### PR Type:


src/citrine/_rest/collection.py
```diff
@@ -30,9 +30,9 @@ class Collection(Generic[ResourceType], Pageable):
     _collection_key: str = 'entries'
     _paginator: Paginator = Paginator()
 
-    def _put_module_ref(self, subpath: str, workflow_id: UUID):
+    def _put_module_ref(self, subpath: str, uid: UUID):
```
bfolie, line 33: since this method is more generic than just workflows

src/citrine/resources/design_execution.py
```diff
@@ -156,24 +156,24 @@ def update(self, model: DesignExecution) -> DesignExecution:
         """Cannot update an execution."""
         raise NotImplementedError("Cannot update a DesignExecution.")
 
-    def archive(self, execution_id: UUID):
+    def archive(self, uid: Union[UUID, str]):
```
bfolie, line 159: For this and other methods that throw an error, might as well just do the breaking migration

src/citrine/_rest/collection.py
talltea: I noticed in other methods you changed the signature to `uid: Union[UUID, str]`, should we change it here too? (this is a nitpick, feel free to ignore)

tests/informatics/test_predictors.py
```diff
@@ -231,21 +231,6 @@ def test_molecule_featurizer(molecule_featurizer):
     }
 
 
-def test_molecule_featurizer_deprecation():
```
talltea, line 234: Why can we 🪓 these tests now? (and the ones in table config)

bfolie: they're testing the different permutations of how arguments can be specified (both/neither/new arg/old arg). But that logic is now consolidated and tested in one place.

bfolie:
e2e tests: https://jenkins.corp.citrine.io/job/nextgen-ITs/17225/ (one failure due to workflow timeout, unrelated)

src/citrine/_rest/collection.py
bfolie: yes

talltea (approved):
lgtm!

PR #636, opened 2021-06-08
Automatically make updates async

bfolie:
# Citrine Python PR
PLA-6824

## Description 
If an update call fails with a BadRequest, this generally means that the validity of the update must be verified, which requires an asynchronous call. For example, async_update() is needed if the bounds of a template are being narrowed. This change catches BadRequest errors in `DataConceptsCollection.register()` (which does double duty as the update method) and calls `async_update()`.

### PR Type:


src/citrine/resources/data_concepts.py
```diff
@@ -485,8 +494,11 @@ def async_update(self, model: ResourceType, *,
             self.poll_async_update_job(job_id, timeout=timeout,
                                        polling_delay=polling_delay)
 
-            # That worked, nothing returned in this case
-            return None
+            # That worked, return nothing or return the object
+            if return_model:
+                return self.get(LinkByUID(scope=scope, id=id))
```
bfolie, line 499: @dleblanc since `async_update` currently returns `None`, I want to make sure that allowing it to get the resource isn't a bad idea. For example, might this `get` try to fetch the resource before the database has finished updating?

talltea (dismissed):
This generally looks fine to me, but I admit I don't have the context for data objects specifically

src/citrine/resources/data_concepts.py
dleblanc: This might be a concern with eventual consistency in Dynamo, but it's no worse than anywhere else that we're doing a read-after-write. I'm ok with it.

src/citrine/resources/data_concepts.py
```diff
@@ -469,7 +478,7 @@ def async_update(self, model: ResourceType, *,
         recursive_foreach(model, lambda x: x.uids.pop(temp_scope, None))  # Strip temp uids
 
         scope = CITRINE_SCOPE
-        id = dumped_data['uids']['id']
+        id = dumped_data['uids'][scope]
```
dleblanc, line 481: Hehe, yeah that's cleaner :)

dleblanc (approved):

talltea (approved):

PR #638, opened 2021-06-10
Real-time status updates

bfolie:
# Citrine Python PR
PLA-7520

## Description 
When the methods `succeeded()`, `in_progress()`, and `failed()` are called, queries the backend for the current status. This s done for modules, workflows, and executions, but not reports, since calling `.report` already fetches the updated report.

There are a few issues that merit discussion.
1. This is not how citrine-python works, fundamentally. Resource objects represent the data model and Collection objects handle the server calls. We're asking a Resource object to do a server call, but it can't re-use the Collection code without running into a circular import. Hence the "get resource" logic has to be duplicated.
2. The user still needs to call `.get` if they want the updated version of the resource, which could be confusing. Consider the following:
```
predictor = AutoMLPredictor(....) # we make a predictor
project.predictors.register(predictor)
predictor.in_progress()
# This throws an exception, because registering the predictor  doesn't update the local copy.
# The in-memory `predictor` doesn't have enough context to make backend calls.
```
Let's say the user knows to update their local copy when registering. We'll still have the following potentially-confusing issue:
```
predictor = AutoMLPredictor(...)  # we make a predictor
predictor = project.predictors.register(predictor)
while predictor.in_progress():
  sleep(5.0)  # Loop until validation finishes
assert predictor.succeeded()
print(predictor.status)  # This will print CREATED or VALIDATING, because we haven't fetched and over-written predictor
predictor = project.predictors.get(predictor.uid)  # Get the updated predictor
print(predictor.status)  # Now this is READY
```

Some of the feedback I've heard boils down to "we want the resources to interact with the platform directly, because going through collections is confusing." The fact that `.status` doesn't automatically update is one manifestation of this issue. But since the client doesn't work that way, I've instead just implemented status checking, since that seems to trip people up a lot. But the additional confusion caused by this behavior might outweigh the convenience. 

### PR Type:


src/citrine/_rest/resource.py
```diff
@@ -39,3 +42,13 @@ def access_control_dict(self) -> dict:
             "type": self._resource_type.value,
             "id": str(self.uid)
         }
+
+
+class ResourceRef(Serializable['ResourceRef']):
```
bfolie, line 47: I originally moved and renamed this class because I was trying to get around circular imports. But I think the move is worth keeping, since this isn't exclusively used for modules anymore.

src/citrine/informatics/modules.py
```diff
@@ -33,23 +38,11 @@ def get_type(cls, data) -> Type['Module']:
...
-
-
-class ModuleRef(Serializable['ModuleRef']):
-    """A reference to a Module by UID."""
-
-    module_uid = properties.UUID('module_uid')
-
-    def __init__(self, module_uid: str):
-        self.module_uid = module_uid
+    def _fetch_status(self) -> 'Module':
```
bfolie, line 41: I ended up duplicating the `get` logic, which isn't great

maxhutch:
I think this merits a broader conversation, because if you pull on the thread of the pattern then I do think you have to address the get call issue and many others stemming from mutable vs immutable objects.

What about a `status(resource)` method that called get and returned the status?  `refresh`, `succeeded`, and `failed` could be done in a similar way.

sbernasek:
I gave it a quick test and this functionality is very convenient - thank you!  

Regarding your concern for users who failed to keep a local copy of the registered resource in memory, in my limited experience I've found that to already be a problem anyway. Without the registered local copy they won't be able to use that resource in subsequent steps of their workflow, like using a registered predictor to trigger a registered predictor evaluation workflow.

I agree that the status attribute being disconnected from the real-time result of _fetch_status is liable to cause some confusion, but again that's confusion that already exists. The only change is that users conditioned on using your new dynamic methods might understandably expect the status attribute to behave similarly. I'm still unfamiliar with the inner workings of citrine python and how we handle object serialization, so I don't know why "status" is a class attribute in the first place. My naive thought would have been to just add a @property decorated "status" method to the Module class that calls _fetch_status. Presumably there's some reason we can't do that.

bfolie:
> My naive thought would have been to just add a @Property decorated "status" method to the Module class that calls _fetch_status. Presumably there's some reason we can't do that.

We could do that. In fact, by duplicating the "get resource from platform" code as we've done in this PR, having the `status` property update in real time would be simple. _But_ now we're making the resource mutable in this one particular way, which is a significant change (and as Hutch said, it tugs on a potentially long thread).

> many others stemming from mutable vs immutable objects

Could you elaborate? I don't think of this change as making the object mutable. Or are you referring to the fact that it pushes us down a path towards mutability (as in the above comment)?

> What about a status(resource) method that called get and returned the status? refresh, succeeded, and failed could be done in a similar way.

We could certainly do that without breaking existing patterns, but I'm not sure it provides much benefit. We'd be exchanging one somewhat-unintuitive method that runs through the collection (`project.predictors.get(pred.uid).succeeded()`) for another somewhat-unintuitive method that runs through the collection (`project.predictors.succeeded(pred)`). It doesn't address the pain point of "I expect resources to be dynamic but they're not".

maxhutch:
> Could you elaborate? I don't think of this change as making the object mutable. Or are you referring to the fact that it pushes us down a path towards mutability (as in the above comment)?

`register` is the other example that comes to mind: do we stick the uid in an object when we register it?  What about fields like `status_description`, `status_info`, `update_time`, etc.  In the current impl,
```
x = pred.status
y = pred.status_description
```
is inconsistent.

> We'd be exchanging one somewhat-unintuitive method that runs through the collection (`project.predictors.get(pred.uid).succeeded()`) for another somewhat-unintuitive method that runs through the collection (`project.predictors.succeeded(pred)`).

I was thinking we'd have a simple method like:
```
x = status(predictor)
y = status_description(predictor)
z = succeeded(predictor)
```
that would use the included session and project to make the update.

bfolie:
> `register` is the other example that comes to mind: do we stick the uid in an object when we register it? What about fields like `status_description`, `status_info`, `update_time`, etc. In the current impl,

For data objects, we do insert a uid during registration _if_ the user has not specified an id. That is the only mutation example. Metadata like `update_time` is not added. AI Engine objects don't self-mutate at all. As a result, the suggested pattern is intentional over-writing:
```
resource = ... # define resource locally
resource = collection.register(resource)  # overwrite with on-platform version
```

> I was thinking we'd have a simple method like:
> 
> ```
> x = status(predictor)
> y = status_description(predictor)
> z = succeeded(predictor)
> ```
> 
> that would use the included session and project to make the update.

Sure, we could do that. I don't think it will quite scratch the user itch, though </speculation>

maxhutch:
I wrote this alternative, in an attempt to be convenient (the user doesn't have to worry about collections) but keep with the immutable/dumb data containers concept that exists today:
https://github.com/CitrineInformatics/citrine-python/pull/665

maxhutch (commented):
I wrote this alternative, in an attempt to be convenient (the user doesn't have to worry about collections) but keep with the immutable/dumb data containers concept that exists today:
https://github.com/CitrineInformatics/citrine-python/pull/665

PR #641, opened 2021-06-11
add try catch to handle any modules that aren't supported by the client

talltea:
# Citrine Python PR
JIRA: https://citrine.atlassian.net/browse/PLA-7606

## Description 
Even though they’ve been deprecated for a while, a user could still have old predictors in their project that include the old GeneralizedMeanPropertyPredictor or DeprecatedExpressionPredictor. And with the recent citrine-python changes, the attempt to deserialize these will result in an error. This will break “list everything and filter by name”, which is a common pattern. This adds a patch to handle this until the backend migrates to the new `/predictor` routes.

### PR Type:


bfolie:
I get a RuntimeError when the deprecated predictor is part of a GraphPredictor, so I added that. I also fixed the flake8 error, and added a warning to give users some idea of why a predictor isn't showing up.

If the user directly tries to get a deprecated predictor, that will fail with an exception. Which isn't ideal, but this patches the most glaring problem. Thanks for fixing it.

bfolie (approved):

bfolie (approved):

talltea:
@bfolie change's look good to me, merging

PR #642, opened 2021-06-11
Adding the tags to MaterialRunByTemplate

pacdaemon:
# Citrine Python PR

## Description 
Replicating the MaterialRunByTemplate modification on the python client. The optional nature of the tag element was taken into account.

Remaining Work:
- [x] Update the docs (this PR)
- Update the GemTableBuilder for filtering terminal material objects.

### PR Type:


src/citrine/gemtables/rows.py
```diff
@@ -53,10 +53,13 @@ class MaterialRunByTemplate(Serializable['MaterialRunByTemplate'], Row):
...
+    tags = properties.Optional(properties.List(properties.String), "tags")
 
     def _attrs(self) -> List[str]:
-        return ["templates", "typ"]
+        return ["templates", "typ", "tags"]
 
     def __init__(self, *,
-                 templates: List[LinkByUID]):
+                 templates: List[LinkByUID],
+                 tags: List[str] = None):
```
fportillo92, line 63: as you already have done in the Scala side, I think this can be a Set as well

src/citrine/gemtables/rows.py
```diff
@@ -53,10 +53,13 @@ class MaterialRunByTemplate(Serializable['MaterialRunByTemplate'], Row):
 
     templates = properties.List(properties.Object(LinkByUID), "templates")
     typ = properties.String('type', default="material_run_by_template", deserializable=False)
+    tags = properties.Optional(properties.List(properties.String), "tags")
```
bfolie, line 56: Could you please add `tags` to the docstring as well. How are the tags used logically?

src/citrine/gemtables/rows.py
pacdaemon: Good catch! thanks

src/citrine/gemtables/rows.py
pacdaemon: I'm about to push one more change that includes the docs and the docstring! I will tag you. Thanks.

docs/source/data_extraction.rst
```diff
@@ -15,12 +15,16 @@ Each Material History corresponds to exactly one row, though the Material Histor
 For example, the material histories of distinct Material Runs will map to exactly two rows even if their histories are identical up to differing Process Runs of their final baking step.
 The only way to define rows right now is through :class:`~citrine.gemtables.rows.MaterialRunByTemplate`, which produces one row per Material Run associated with any of a list of material templates.
 
+The list of tags is optional. The usage scenario is for fine-grained curation. The list of tags should be interpreted as an OR clause, meaning, if a terminal material doesn't contain any of the tags it will be filtered out.
```
pacdaemon, line 18: @bfolie I updated the docs too.

fportillo92 (dismissed):
Looks good!

bfolie (changes requested):
I'm still confused by the tags logic. The docstring says "Terminal materials containing any of those tags are going to be included" which sounds like the match is on material runs that match on any template OR on any tag.
But the long-form docs say "The list of tags should be interpreted as an OR clause, meaning, if a terminal material doesn't contain any of the tags it will be filtered out." That is described as an OR, but it sounds like an AND: the material run must match a template AND at least one tag, if the tags are present.

Correct me if I'm wrong, but my assumption is that the logic is:
```
if there are tags:
  match template AND (match tag1 OR match tag2 OR match tag3 OR....)
else:
  match template
```
Is that correct?

pacdaemon:
> I'm still confused by the tags logic. The docstring says "Terminal materials containing any of those tags are going to be included" which sounds like the match is on material runs that match on any template OR on any tag.
> But the long-form docs say "The list of tags should be interpreted as an OR clause, meaning, if a terminal material doesn't contain any of the tags it will be filtered out." That is described as an OR, but it sounds like an AND: the material run must match a template AND at least one tag, if the tags are present.
> 
> Correct me if I'm wrong, but my assumption is that the logic is:
> 
> ```
> if there are tags:
>   match template AND (match tag1 OR match tag2 OR match tag3 OR....)
> else:
>   match template
> ```
> 
> Is that correct?

Your assumption is correct. I can see that. I word in the opposite way, saying:

> optional list of tags for filtering. If a terminal material doesn't contain any of the tags it will be filtered out.

This way I'm denoting exclusion. Is it ok?

docs/source/data_extraction.rst
```diff
@@ -15,12 +15,16 @@ Each Material History corresponds to exactly one row, though the Material Histor
 For example, the material histories of distinct Material Runs will map to exactly two rows even if their histories are identical up to differing Process Runs of their final baking step.
 The only way to define rows right now is through :class:`~citrine.gemtables.rows.MaterialRunByTemplate`, which produces one row per Material Run associated with any of a list of material templates.
 
+The set of tags is optional. The usage scenario is for fine-grained curation. The list of tags should be interpreted as an OR clause, meaning, if a terminal material doesn't contain any of the tags it will be filtered out.
```
bfolie, line 18:
```suggestion
The set of tags is optional.
The usage scenario is for fine-grained curation.
If a terminal material doesn't contain any of the tags it will be filtered out.
```

bfolie:
> > optional list of tags for filtering. If a terminal material doesn't contain any of the tags it will be filtered out.
> 
> This way I'm denoting exclusion. Is it ok?

Yes, that way of denoting it makes sense. I made a suggestion to remove the part about "interpreted as an OR clause", since it's an AND in one way and an OR in the other way. I also moved each sentence to a separate line, which we do to make future diffs easier to read.

bfolie (approved):

PR #646, opened 2021-06-21
PLA 2841 - API routes for generic storables

sfriedowitz:
# Citrine Python PR

## Description 
Through initial testing I encountered a logical need to query GEMD resources without knowing the specific type of object. It was uncovered that a handful of generic "storables" end points already exist on the product API, and PLA-2841 was open requesting functionality to hit these routes in citrine-python.

There are a handful of issues with the current draft:
- The GemResourceCollection inherits from Collection rather than DataConceptsCollection because the calls can return any type of GEMD objects/templates, rather than a single type. Further, not all methods present on DataConceptsCollection have matching end points for generic storables. It is worth thinking if this is the best inheritance structure.
- Since calls on GemResourceCollection return any type of GEMD object, we do not know the appropriate type to build the JSON response from. This is currently worked around by calling DataConcepts.get_type(data), which accesses the DataConcepts.get_type class method externally. This is not intended usage, and was under a "pragma: no cover" line. Does it make more sense for this lookup to be moved external to DataConcepts? How else to work around?

### PR Type:


src/citrine/resources/gem_resource.py
```diff
@@ -0,0 +1,148 @@
...
+
+from citrine.resources.data_concepts import _make_link_by_uid
+from citrine.resources.data_concepts import DataConcepts
+from citrine._rest.collection import Collection
+from citrine._session import Session
+
+# Proper generic parameter for typing?
+# It is not a single generic type, but rather
+# polymorphic for any GEM object
+GemResourceType = TypeVar('GemResourceType', bound='DataConcepts')
```
bfolie, line 17: I believe that this is correct

src/citrine/resources/gem_resource.py
```diff
@@ -0,0 +1,148 @@
...
+        data: dict
+            A serialized data model object.
+
+        Returns
+        -------
+        GemResourceType
+            A data model object built from the dictionary.
+
+        """
+        return DataConcepts.get_type(data).build(data)
```
bfolie, line 51: Does `DataConcepts.build(data)` not work?

src/citrine/resources/gem_resource.py
```diff
@@ -0,0 +1,148 @@
...
+                 scope: Optional[str] = None) -> GemResourceType:
+        """
+        Get a GEMD resource within the project by its id.
+
+        Parameters
+        ----------
+        uid: Union[UUID, str, LinkByUID, BaseEntity]
+            A representation of the object (Citrine id, LinkByUID, or the object itself)
+        scope: Optional[str]
+            [DEPRECATED] use a LinkByUID to specify a custom scope
```
bfolie, line 63: Noooo! Don't replicate the deprecated pattern! The only argument should be `uid`, and then you call `_make_link_by_uid(uid)`.

bfolie:
I checked the methods and they work (🎉 ).

I left a few small comments. Beyond that, the remaining work is to write tests for full coverage and to fix the flake8 style errors.

kroenlein:
Nitpick: The resources should probably be `GemdResourceType` or `GEMDResourceType` and the file should probably be `gem_resource.py`.  GEMD refers to the data objects and GEM w/o a D (at least in this repo) is only used to refer to GEM Tables, as internal discussions showed people thought "Graphical Expression of Materials Data Tables" was lesser than "Graphical Expression of Materials Tables" (or at least as I recall).  Doesn't impact the function at all.

I also thought the storables endpoint you are hitting covered more than just data objects.

sfriedowitz:
> Nitpick: The resources should probably be `GemdResourceType` or `GEMDResourceType` and the file should probably be `gem_resource.py`. GEMD refers to the data objects and GEM w/o a D (at least in this repo) is only used to refer to GEM Tables, as internal discussions showed people thought "Graphical Expression of Materials Data Tables" was lesser than "Graphical Expression of Materials Tables" (or at least as I recall). Doesn't impact the function at all.
> 
> I also thought the storables endpoint you are hitting covered more than just data objects.

That's a good point, I was unsure the proper naming semantics, so I followed the GEM Table example that was there already. I will go ahead and make that change.

As far as I can tell from testing, the storables end points only works for data objects (including templates). Although I haven't seen any specific documentation clarifying that.

tests/resources/test_gemd_resource.py
```diff
@@ -0,0 +1,160 @@
...
+@pytest.fixture
+def session() -> FakeSession:
+    return FakeSession()
+
+
+@pytest.fixture
+def collection(session) -> GEMDResourceCollection:
+    return GEMDResourceCollection(
+        project_id=UUID('6b608f78-e341-422c-8076-35adc8828545'),
+        dataset_id=UUID('8da51e93-8b55-4dd3-8489-af8f65d4ad9a'),
```
bfolie, line 23: Rather than hard-coding, I suggest using random uuids (`uuid.uuid4()`) and then referring to them by variable if you need to do a comparison (though it doesn't look like any of the tests here check for the specific uuids)

bfolie:
Good catch about GEM vs. GEMD @kroenlein.

@sfriedowitz, when Ken says "only GEMD objects" I believe he means "not templates." We all expect the `storables` endpoint to work for objects and templates, which is why we went with the name `GEMDResourceCollection`, since "resource" seemed general enough to cover both objects and templates. Ken, is there someplace in the PR where it is described as working specifically for data objects?

tests/resources/test_gemd_resource.py
sfriedowitz: Changed in most recent push, along with a few doc updates.

bfolie (dismissed):
Nice job. To close the loop, I suggest tinkering with our end-to-end tests a bit so that the storables route is exercised somewhere

sfriedowitz:
@bfolie Slight change of plan. Just talked with Hutch, and we agree it makes most sense to change the inheritance chain for GEMDResource to derive from DataConceptsCollection. Will edit and post again for review.

tests/resources/test_gemd_resource.py
```diff
@@ -0,0 +1,85 @@
...
+    })
+
+    # When
+    gems = list(collection.list())
+
+    # Then
+    assert 1 == session.num_calls
+    expected_call = FakeCall(
+        method='GET',
+        path='projects/{}/storables'.format(collection.project_id),
```
kroenlein, line 48: Using string formatting methods to assemble URLs isn't great practice (see #650), but it's not worth sweating since this is only a test.  Any reason you didn't use an `f""` string?
```suggestion
        path=format_escaped_url('projects/{}/storables', collection.project_id),
```

src/citrine/resources/gemd_resource.py
```diff
@@ -0,0 +1,92 @@
...
+        DataConcepts
+            A data model object built from the dictionary.
+
+        """
+        return DataConcepts.build(data)
+
+    def update(self, model: DataConcepts) -> DataConcepts:
+        """To update an arbitrary GEMD resource, please use dataset.update instead."""
+        raise NotImplementedError("To update an arbitary GEMD resource,"
+                                  " please use dataset.update instead.")
```
kroenlein, line 53: Are we forbidding the query `dataset.storables.update(...)`?  It seems like these should check if `self.dataset_id` is set, and dispatch if possible.  Regardless, I don't think this is a `NotImplementedError` error, this would probably be a `ValueError` since they tried to invoke it with `dataset is None`.

https://docs.python.org/3/library/exceptions.html#NotImplementedError

src/citrine/resources/gemd_resource.py
```diff
@@ -0,0 +1,92 @@
...
+from uuid import UUID
+
+from gemd.entity.base_entity import BaseEntity
+from gemd.entity.link_by_uid import LinkByUID
+
+from citrine.resources.data_concepts import DataConcepts, DataConceptsCollection
+from citrine._session import Session
+
+
+class GEMDResourceCollection(DataConceptsCollection[DataConcepts]):
```
kroenlein, line 12: The most consistent class name is probably `GEMDEntity` since that's the parent element of both object and template in gemd-python, though that includes several classes that are not directly supported with this class.

kroenlein (commented):
A few more nits.  I started a bunch of comments a few days ago to support consistency w/ #650, but they got obviated with the inheritance from DataConcepts (I think).

tests/resources/test_gemd_resource.py
sfriedowitz: easy fix, thanks for pointing out. and yes, all implementations were passed to DataConceptsCollection after switching inheritance, so should be consistent.

src/citrine/resources/gemd_resource.py
sfriedowitz: @bfolie Can use your input here. The logic was that since the storables route does not have functionality for register/update/delete, we would prefer the user to use the methods on specific collections or on `dataset`, rather than dispatching to that through `GEMDCollection`. 

We can add that dispatch through the parent dataset, but it seems redundant to me. Think it's worth adding in?

src/citrine/resources/gemd_resource.py
bfolie: I agree that "Entity" is the most appropriate term, if we're inheriting terminology from `gemd-python`. Doesn't it include everything? It looks like the things in `gemd-python` that inherit `BaseEntity` are just objects, object templates, and attribute templates.

The term "Entity" is used in some places in `citrine-python`, mostly in types as part of docstrings, so I wouldn't be opposed to calling this `GEMDEntityCollection`. On the other hand, `citrine-python` heavily uses the term "Resource" to refer to objects that are stored on-platform. The directory is `/resources`, and we also talk about "AI Resources" in places.

src/citrine/resources/data_concepts.py
```diff
@@ -223,7 +223,7 @@ def _make_link_by_uid(gemd_object_rep: Union[str, UUID, BaseEntity, LinkByUID],
              "object. To specify a custom scope, use a LinkByUID.", DeprecationWarning)
     if isinstance(gemd_object_rep, BaseEntity):
         if not gemd_object_rep.uids:  # an empty dictionary
-            raise ValueError('GEMD resource must have at least one uid to construct a link.')
+            raise ValueError('GEMD object must have at least one uid to construct a link.')
```
bfolie: Can't this be a template as well?

src/citrine/resources/gemd_resource.py
bfolie: If the dataset is not None, then it's a `NotImplementedError`, because we haven't implemented it :p
If the method is otherwise implemented but the dataset is `None`, then we've been using `RuntimeError`.

register/update/delete is a nice-to-have. I was more ambivalent when we last discussed it, but now I think it's worth doing, both because users will expect it and because it should be easy to implement by calling up to the already-implemented `Dataset` method. It's true that the `storables/` route doesn't explicitly implement, say, a POST method, but that's immaterial to a user. As far as they're concerned, their intent is clear, so we should be able to deal with it.

src/citrine/resources/gemd_resource.py
```diff
@@ -0,0 +1,92 @@
...
+        """Polling is only available on collections of specific GEMD resources."""
+        raise NotImplementedError("Polling is only available on collections"
+                                  " of specific GEMD resources.")
+
+    def _get_relation(self, relation: str, uid: Union[UUID, str, LinkByUID, BaseEntity],
+                      scope: Optional[str] = None, forward: bool = True, per_page: int = 100
+                      ) -> Iterator[DataConcepts]:
+        """Relationship searching is only available on collections of specific GEMD resources."""
+        raise NotImplementedError("Relationship searching is only available"
+                                  " on collections of specific GEMD resources.")
```
bfolie, line 92: This is an internal method, so we don't need to override it

src/citrine/resources/gemd_resource.py
sfriedowitz: I'm falling back to `resource` being the most fitting term within `citrine-python`. It is used significantly throughout the other collections, and seems general enough. I will opt to keep as such for this iteration.

sfriedowitz:
New storables additions to the product API will be soon be added to handle generic register/update/delete calls. This will require an implementation change of the current iteration. Tabling as a draft until then.

sfriedowitz:
The full interface is fleshed out on GEMDResourceCollection. This involves a refactor of dataset methods and associated tests to live on GEMDResourceCollection, as well as a lookup method `DataConcepts.get_collection_type` to determine the associated collection type for a given object.

One main question: Do we want to deprecate calls to `dataset.register`, `dataset.update`, etc.., since these generic methods now live on the generic GEMDResourceCollection, or do we want a subset of these methods to remain on Dataset?

bfolie:
> One main question: Do we want to deprecate calls to `dataset.register`, `dataset.update`, etc.., since these generic methods now live on the generic GEMDResourceCollection, or do we want a subset of these methods to remain on Dataset?

I don't think that we need to deprecate those methods. It doesn't hurt to give users more idiomatic ways to do the same thing, as long as the intent is clear.

bfolie:
Because this is a significant change and involves API routes, I suggest you run the end-to-end tests using this branch and we make sure that they pass before we merge. Do you know how to do that?

sfriedowitz:
> Because this is a significant change and involves API routes, I suggest you run the end-to-end tests using this branch and we make sure that they pass before we merge. Do you know how to do that?

src/citrine/resources/gemd_resource.py
```diff
@@ -0,0 +1,268 @@
...
+
+
+class GEMDResourceCollection(DataConceptsCollection[DataConcepts]):
+    """A collection of any kind of GEMD objects/templates."""
+
+    _path_template = 'projects/{project_id}/storables'
+    _dataset_agnostic_path_template = 'projects/{project_id}/storables'
+    _individual_key = None
+    _collection_key = None
+    _resource = DataConcepts
```
kroenlein, line 24: Why are these defined in this class?  The unit tests pass with these commented out.

src/citrine/resources/gemd_resource.py
```diff
@@ -0,0 +1,268 @@
...
+        return self._collection_for(model).async_update(
+            model,
+            dry_run=dry_run,
+            wait_for_response=wait_for_response,
+            timeout=timeout,
+            polling_delay=polling_delay,
+            return_model=return_model
+        )
+
+    def delete_contents(
```
kroenlein, line 185: This method doesn't grammatically make sense as the gemd-collection; it's supposed to delete the contents of a dataset, so the method should probably continue living there.

kroenlein (commented):
A couple questions.

src/citrine/resources/gemd_resource.py
sfriedowitz: > This method doesn't grammatically make sense as the gemd-collection; it's supposed to delete the contents of a dataset, so the method should probably continue living there.

Makes sense, changed it back!

src/citrine/resources/gemd_resource.py
sfriedowitz: Yes they appear unused, so am removing in the changes.

sfriedowitz:
> Because this is a significant change and involves API routes, I suggest you run the end-to-end tests using this branch and we make sure that they pass before we merge. Do you know how to do that?

ITs passing with current build (https://jenkins.corp.citrine.io/job/nextgen-ITs/18296/). Also planning to add some test coverage to nextgen specific to these routes.

sfriedowitz:
> > Because this is a significant change and involves API routes, I suggest you run the end-to-end tests using this branch and we make sure that they pass before we merge. Do you know how to do that?
> 
> ITs passing with current build (https://jenkins.corp.citrine.io/job/nextgen-ITs/18296/). Also planning to add some test coverage to nextgen specific to these routes.

And a test on nextgen for when it goes live: https://github.com/CitrineInformatics/nextgen-devkit/tree/PLA-2841/generic-storables

src/citrine/resources/gemd_resource.py
bfolie: They're not used in this case, but they are part of the interface, so I would vote for retaining them.

But I won't block on it, and am approving.

bfolie (approved):

PR #650, opened 2021-06-29
Bugfix/url escape

kroenlein:
# Citrine Python PR

## Description 
This solves a potential injection issue by systematically changing urls that had been manually constructed to utilize urllib to percent encode reserved and high-bit characters.

https://citrine.atlassian.net/browse/PLA-7198

### PR Type:


src/citrine/_session.py
```diff
@@ -35,7 +37,7 @@ def __init__(self,
                  port: Optional[str] = None):
         super().__init__()
         self.scheme: str = scheme
-        self.authority = ':'.join([host or '', port or ''])
+        self.authority = ':'.join([host] if host else [] + [port] if port else [])
```
kroenlein, line 40: This might actually be better written as 
```suggestion
        self.authority = ':'.join([host] + [port] if port else [])
```
since the host cannot be `None` if the signature is being respected.

bfolie: But there's nothing stopping a user from passing `host=None`. Python just doesn't demand that kind of respect :p

Is the point of this change that, when `port` isn't specified, the url should be `https://host/` instead of `https://host:/`?

kroenlein: That is an issue that both versions of the line address.  If we want to be strictly compliant with the spec, we should have `https:FAQ.html` as a relative path when no host or path is specified -- the unsplit should handle that, though we'll get a bad result if port is specified and host is None.

I'm going with this proposed change should be axed.

kroenlein:
Passes end-to-end tests:
https://jenkins.corp.citrine.io/job/nextgen-ITs/17912/

bfolie (approved):
nice

PR #651, opened 2021-07-01
Reverted failing port change

dleblanc:
Trying to fix the dev0 regression.

### PR Type:


kroenlein (approved):
Seems odd to cause the problem, but reasonable considering the issue.  Line 78 change could be responsible as well.

src/citrine/_session.py
```diff
@@ -37,7 +37,8 @@ def __init__(self,
                  port: Optional[str] = None):
         super().__init__()
         self.scheme: str = scheme
-        self.authority = ':'.join([host] if host else [] + [port] if port else [])
+        # self.authority = ':'.join([host] if host else [] + [port] if port else [])
+        self.authority = ':'.join([host or '', port or ''])
```
kroenlein, line 41:
```suggestion
        self.authority = ':'.join(([host] if host else []) + [port] if port else [])
```
The actual fix; poor testing on my part, sorry.

kroenlein (approved):
Order of operations, a pair of parentheses got accidentally deleted.

PR #653, opened 2021-07-08
Auto configure workflow

sfriedowitz:
## Description 
Goal to implement a chain of auto configure calls from material to design execution.

Current additions include:
1) Utility method to extract attribute names for a provided material history, for use in providing a score function to the configure workflow
2) Auto configure chain that follows: Material History -> GemTable -> AutoPredictor -> Auto DesignSpace -> Design Workflow -> Execution w/ provided score

In principle, this structure works. Though in practice, the auto-configured assets fail to validate for most test datasets. Opening as a draft work in progress for feedback.

### PR Type:


src/citrine/builders/auto_configure.py
```diff
@@ -0,0 +1,183 @@
...
+        Dictionary of keys of the form
+        {'properties': [...], 'conditions': [...], 'parameters': []}
+
+    """
+    if mode not in {"FORMULATION", "SIMPLE"}:
+        msg = "Called with mode: {}.  Expected 'SIMPLE' or 'FORMULATION'.".format(mode)
+        raise ValueError(msg)
+
+    link = _make_link_by_uid(material)
+    history = project.material_runs.get_history(scope=link.scope, id=link.id)
```
maxhutch, line 66: If this is long running, we might want to avoid it in the `!full_history` case.  Actually doing that isn't necessary, but I'd add a comment.

src/citrine/builders/auto_configure.py
```diff
@@ -0,0 +1,183 @@
...
+def auto_configure_material(
+        *,
+        project: Project,
+        material: Union[MaterialRun, LinkByUID, str, UUID],
+        score: Score,
+        mode: str = 'SIMPLE',
+        label: str = '',
+        print_status_info: bool = False,
+) -> DesignExecution:
+    """[ALPHA] Description.
```
maxhutch, line 100: "Description" needs to be filled in.

src/citrine/builders/auto_configure.py
```diff
@@ -0,0 +1,183 @@
...
+    if full_history:
+        search_history = flatten(history)
+    else:
+        # Limit the search to contain the terminal material/process/measurement
+        search_history = [history.spec.template, history.process.template]
+        search_history.extend([msr.template for msr in history.measurements])
+
+    keys = {'properties': [], 'conditions': [], 'parameters': []}
+    for obj in search_history:
+        parent_name = f'{obj.name}~' if mode == 'SIMPLE' else ''
```
maxhutch, line 77: Have we tested both modes?  I think Strehlow and Cook is a "Simple" example.

src/citrine/builders/auto_configure.py
```diff
@@ -0,0 +1,183 @@
...
+        Project to use when accessing the Citrine Platform
+    material: Union[MaterialRun, LinkByUID, str, UUID]
+        Sample material to extract descriptor keys
+    score: Score
+        Scoring function used to rank candidates during design execution.
+        Must contain objectives/constraints with matching descriptor keys
+        to those found within the provided material history.
+    mode: str
+        Option in {'SIMPLE', 'FORMULATION'} that controls formatting descriptor keys
+        Default: 'SIMPLE'
```
maxhutch, line 114: Is "Simple" the right default?  cc @kroenlein

src/citrine/builders/auto_configure.py
```diff
@@ -0,0 +1,183 @@
...
+            for parameter in obj.parameters:
+                keys['parameters'].append(parent_name + parameter[0].name)
+    return keys
+
+
+def auto_configure_material(
+        *,
+        project: Project,
+        material: Union[MaterialRun, LinkByUID, str, UUID],
+        score: Score,
```
maxhutch, line 95: I think just doing an objective could be even simpler.  Or you could have both as options and just make sure either the score or the objective is given.

maxhutch (commented):
Didn't get all the way through yet.

src/citrine/builders/auto_configure.py
kroenlein: I would guess "SIMPLE" is the expected default -- the most surprising state would be to just pass a Strehlow-like table and get feedback that it doesn't make sense.

Secondary comment: if you are using an enumeration, you should use a BaseEnumeration:
https://github.com/CitrineInformatics/citrine-python/blob/6d8cafc69c511498175fdef77d2195e0ccb49a71/src/citrine/resources/table_config.py#L24

sfriedowitz: @kroenlein Is there ever a case where a user could wish for a `TableBuildAlgorithm.SINGLE_ROW` table, but then a default predictor using the `FORMULATION` option? Or visa-versa?

If not, then I can introduce a new Enum for handling the single option pattern on this method.

src/citrine/builders/auto_configure.py
bfolie, line 66: `get_history` takes a `Union[MaterialRun, LinkByUID, str, UUID]` as the argument to `id`, and the argument `scope` is deprecated. So just call `project.material_runs.get_history(id=material)`

src/citrine/builders/auto_configure.py
kroenlein: Linking was explicitly for reference in building your own enumeration, thought reusing isn't a terrible idea.  If a user wanted to reference an existing property table (e.g., someone transcribed something out of Perry's), they might build the main table as a single row (they need the maceration time in the same row as the terminal material) but then want to use a mean property predictor informed by the property table.  But, given that I'm talking about multiple tables as inputs, that might already be deviating outside the scope of this work.

src/citrine/builders/auto_configure.py
sfriedowitz: In the case of a str/UUID being passed for material, we still need to get the associated templates for the terminal material, which could possibly be done with multiple get requests. But that seems cruder. Since it's generally not long running, having a single `get_history` call seems to be the most convenient.

I will add comment for clarity.

sfriedowitz: This actually is a small bug on `material_runs.get_history`, which does not have an appropriate default option of `scope=None` to allow for that signature.

I will change, and add the default argument in my changes.

src/citrine/builders/auto_configure.py
```diff
@@ -0,0 +1,183 @@
...
+    -------
+    DesignExecution
+        Triggered design execution given the auto-configured workflow and provided `score`
+
+    """
+    if mode not in {"FORMULATION", "SIMPLE"}:
+        msg = "Called with mode: {}.  Expected 'SIMPLE' or 'FORMULATION'.".format(mode)
+        raise ValueError(msg)
+
+    suffix = ' - {}'.format(label) if label else ''
```
maxhutch, line 132: We might want to use the label as a prefix, otherwise multiple invocations willr esult in a bunch of "Default {Asset Name}" in the lists.

src/citrine/builders/auto_configure.py
```diff
@@ -0,0 +1,183 @@
...
+    )
+    table_config = project.table_configs.register(table_config)
+    table = project.tables.build_from_config(table_config)
+
+    print('Building default predictor from GEM table...')
+    formulation = None
+    if mode == 'FORMULATION':
+        formulation = FormulationDescriptor('Formulation descriptor')
+    data_source = GemTableDataSource(
+        table_id=table.uid, table_version=table.version, formulation_descriptor=formulation
```
maxhutch, line 153: I believe it will work without `formulation_descriptor`, since we fill in a default one if we see formulation data but none is provided.  The descriptor key is just "Formulation", which you could get from the [`descriptors_from_data_source` method](https://github.com/CitrineInformatics/citrine-python/blob/main/src/citrine/resources/descriptors.py#L48)

maxhutch (commented):
A couple more inline comments.  Overall, this looks great!

src/citrine/builders/auto_configure.py
bfolie: Yep that's a bug, thank you

src/citrine/builders/auto_configure.py
sfriedowitz: Tested and that is the case, will remove the conditional check

src/citrine/builders/auto_configure.py
sfriedowitz: If specifying an objective (or list of objectives), how should we include baseline information? My only thought would be to read the table, determine appropriate min/max of training data columns, and set the appropriate baseline given which type of objective.

Although at that point, maybe it is safer to have the user pass a pre-constructed Score where they can control those parameters?

src/citrine/builders/auto_configure.py
```diff
@@ -0,0 +1,183 @@
...
+    full_history: bool
+        Whether to extract descriptor keys from the full material history,
+        or only the provided (terminal) material
+        Default: True
+
+    Returns
+    -------
+    dict
+        Dictionary of keys of the form
+        {'properties': [...], 'conditions': [...], 'parameters': []}
```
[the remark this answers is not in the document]
sfriedowitz: Reflected in recent edits.

sfriedowitz:
Have made a significant design overhaul given the desire for shortcutting with table/predictor.

Users can now create a `AutoConfigureWorkflow`, and use the methods `from_material`, `from_table`, or `from_predictor` to go the rest of the way. Also now includes optional PEW/PEE during the auto configure steps.

Still need to finish test coverage, but would appreciate input in the meanwhile.

maxhutch:
Could you include some usage examples (even just sketch -- they don't need to actually run)?  You could stick them in the comments here and migrate them to docs after feedback.  In particular, I'm wondering how a user resumes after fixing an asset.  Turning this into a class gives us some options there.

Also, we do have a method for getting a default predictor evaluation workflow [here](https://github.com/CitrineInformatics/citrine-python/blob/main/src/citrine/resources/predictor_evaluation_workflow.py#L64), so that could be invoked if no evaluator or evaluation workflow is given.

sfriedowitz:
> Could you include some usage examples (even just sketch -- they don't need to actually run)? You could stick them in the comments here and migrate them to docs after feedback. In particular, I'm wondering how a user resumes after fixing an asset. Turning this into a class gives us some options there.

Sure! Here is the usage I have been testing

```
material = MaterialRun(...) # A real one

pv = PlatformVocabulary.from_material(project=project, material=mr, mode=AutoConfigureMode.PLAIN)
keys = list(pv.keys()) # The descriptors used throughout this MR

# A sample score/evaluator to pass to auto config
score = LIScore(
    objectives=[ScalarMaxObjective(descriptor_key=keys[-1])],
    baselines=[100]
)
evaluator = CrossValidationEvaluator(
    name='CV Evaluator',
    n_folds=8,
    n_trials=3,
    responses={keys[-1]},
    metrics={RMSE(), PVA()}
)

# Create auto config class
auto_config = AutoConfigureWorkflow(project=project, name='Test v1')
auto_config.from_material(
    material=mr, 
    score=score,
    evaluator=evaluator,
    mode=AutoConfigureMode.PLAIN, 
    print_status_info=False
)
```

My logic was to store assets only after they are registered. Let's say, for example, a default predictor is created and registered, but it fails validation. The methods will raise and halt there, but the predictor is still accessible via `auto_config.predictor`. The user can then take that predictor, correct it somehow, and pass to `auto_config.from_predictor(...)`.

If the workflow fails during design space/design workflow validation, there is currently no method to resume from at that stage (unless we implemented a `auto_config.from_design_space(...)` shortcut. But, the assets are still stored in the class and can be re-used in some way.

maxhutch:
This looks like a great start (and we can go ahead and get it in to avoid the forever-PR).  It leads to a few ideas that I think would all be easy extensions:
 * Creating an `AutoConfigureWorkflow` could look for existing assets in the project that match the name and use them to setup the object.  This would help when resuming across working sessions.
 * `AutoConfigureWorkflow` could have a `.status -> String` method that returns the most recent issue
 * `AutoConfigureWorkflow` could have a `.what_is_nex() -> String` method that gives the user instructions about where they are and what they should do next
 *  Use the auto-PEW to avoid requiring an evaluator
 * Accept a score **or** an objective.  If its an objective, then pull the baseline from the training data table
 * Parse the score from the most recent execution, in the case of successfully auto-configured workflows, so that the user can tweak it.  We need this client-side because Score aren't identifiable assets today.
 * Have a `.execute()` method that will execute the design workflow.  Useful after tweaks.

sfriedowitz:
> * `AutoConfigureWorkflow` could have a `.status -> String` method that returns the most recent issue
> * `AutoConfigureWorkflow` could have a `.what_is_nex() -> String` method that gives the user instructions about where they are and what they should do next
> * Use the auto-PEW to avoid requiring an evaluator
> * Have a `.execute()` method that will execute the design workflow.  Useful after tweaks.

These additions are now implemented. Flagging this ready for review now that test coverage is complete.

maxhutch:
> These additions are now implemented.

> and we can go ahead and get it in to avoid the forever-PR

Overachiever...

src/citrine/builders/auto_configure.py
```diff
@@ -0,0 +1,558 @@
...
+        execution = self.design_workflow.design_executions.trigger(score)
+        self._design_execution = execution
+        return self.design_execution
+
+    def from_material(
+            self,
+            *,
+            material: Union[str, UUID, LinkByUID, MaterialRun],
+            evaluator: Optional[PredictorEvaluator] = None,
+            design_space: Optional[DataSourceDesignSpace] = None,
```
maxhutch, line 220:
```suggestion
            design_space: Optional[DesignSpace] = None,
```
Why limit to a `DataSourceDesignSpace`?  A user might want to re-use an existing product design space.

This applies to all the `DataSourceDesignSpace` signatures.

maxhutch (changes requested):
Looks really good.  There's one small mechanical issue with the [default PEW](https://citrineinformatics.github.io/citrine-python/reference/citrine.resources.predictor_evaluation_workflow.html?highlight=create_default#citrine.resources.predictor_evaluation_workflow.PredictorEvaluationWorkflowCollection.create_default), mentioned inline.

This is very **non-blocking**: You said you implemented `.what_is_nex() -> String` but I don't quite see that.  Happy to do it in a follow-up PR.

src/citrine/builders/auto_configure.py
sfriedowitz: This has been updated for any design space type.

src/citrine/builders/auto_configure.py
```diff
@@ -322,47 +370,164 @@ def from_predictor(
...
+        print("Configuring predictor evaluation workflow/execution...")
+        if evaluator is not None:
+            # We got an evaluator, so make a new PEW for it
             pew = PredictorEvaluationWorkflow(
                 name=f'{self._prefix}Predictor Evaluation Workflow',
                 evaluators=[evaluator]
             )
+        else:
+            # No evaluator, so create a default PEW
+            pew = self.project.predictor_evaluation_workflows.create_default(
```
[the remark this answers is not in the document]
sfriedowitz: Tweaked the structure of this block to not trigger a new execution from the default PEW. Rather, we obtain it by `pew.executions.list()` after waiting for the default PEW to validate

sfriedowitz:
> This is very **non-blocking**: You said you implemented `.what_is_next() -> String` but I don't quite see that. Happy to do it in a follow-up PR.

This is somewhat encompassed by the status tracking, which would allow a user to see where things are/failed in a structured manner. Although I can see utility in an explicit status telling you which step to use next.

I agree, I can follow-up with this change and any other additions we think of by then.

src/citrine/builders/auto_configure.py
```diff
@@ -0,0 +1,562 @@
...
+        self._status_info = pew.status_info
+
+        if pew.status == 'FAILED':
+            # Can proceed without raising error, but can't get PEE
+            self._status = AutoConfigureStatus.PEW_FAILED
+            warnings.warn("Predictor evaluation workflow failed -- unable to configure execution.")
+        else:
+            if evaluator is None:
+                # Get execution from default
+                self._predictor_evaluation_execution = next(pew.executions.list(), None)
```
maxhutch, line 489: The module validation completion and execution creation aren't guaranteed to occur in any particular order, so there's a chance for a small race condition here where the PEW is valid but the execution hasn't been created.  They occur pretty tightly together, so a short wait here might be appropriate.  Alternatively, you could keep listing until an execution is present, up to a timeout.

sfriedowitz: added a 5 sec timeout on PEE List

maxhutch (approved):

PR #654, opened 2021-07-14
Pla 7421/enforce deprecations

bfolie:
# Citrine Python PR

## Description 
Since it can be difficult to keep track of planned deprecations, I wrote some code to scan the repo for outdated annotations and throw a flag if it finds any. Since it seemed independent, I then put this code in a separate repo and here it is called as part of the Travis build script.

This works, as can be seen by the fact that it caught two outdated deprecation annotations, but the vast majority of our deprecations are on _arguments_ and are noted with a `DeprecationWarnings`. This presents a problem, because none of the extant Python deprecation libraries allow you to deprecate an argument and provide a removal date/version. The options are:
* [Deprecation](https://github.com/briancurtin/deprecation), which we use, doesn't let you deprecate an argument
* [Deprecated](https://github.com/tantale/deprecated) doesn't let you deprecate an argument nor does it specify a removal version
* [Twisted](https://twistedmatrix.com/documents/current/api/twisted.python.deprecate.html) lets you deprecate an argument, but you can't specify the removal version

Three options for going forward are:
1. Abandon the idea
2. Hijack the `DeprecatedWarning` class in `Deprecation` for this purpose (which I started doing [here](https://github.com/bfolie/derp/compare/catch-deprecated-warning?expand=1) but it requires some more work because the current code assumes deprecations only come in the form of decorators)
3. Make a PR into the `Deprecation` library to allow for deprecated arguments, but since somebody [tried this](https://github.com/briancurtin/deprecation/pull/35) years ago and it foundered, I'm not optimistic about it working

### PR Type:


maxhutch:
I'd be inclined to take the coverage where it exists (on classes and methods) and leave the other cases unchecked.  In my mind, this is more of a reminder / backstop than a tool for deciding when to remove deprecated code, so we shouldn't reject it just because it doesn't cover everything.

Thoughts @talltea ?

sfriedowitz:
> I'd be inclined to take the coverage where it exists (on classes and methods) and leave the other cases unchecked. 

I tend to agree with this. I feel like since this is just serving as a reminder for us, we should try to avoid as much heavy lifting as possible and take the coverage that works. Also, if option (2) from @bfolie is pursued, it could potentially be added into a future version of derp

sfriedowitz:
My only concern is that we just added another dependency in the form of derp into the mix, which may need to be maintained. Do you feel like it's better for these checks to live within citrine-python somewhere?

sfriedowitz (approved):

PR #655, opened 2021-07-15
Bugfix/table config uid attribute

sbernasek:
# Citrine Python PR

## Description 
Defines `TableConfig.uid` to alias  the `TableConfig.config_uid` attribute. This fix allows table configs to be shared using the `Project.share()` method, but does not deprecate the existing `TableConfig.config_uid` attribute as it has not yet been decided how to reconcile the fact that multiple different TableConfig versions may share the same identifier value.

### PR Type:


src/citrine/__version__.py
```diff
@@ -1 +1 @@
-__version__ = '1.2.0'
+__version__ = '1.2.1'
```
maxhutch, line 1: We're adding a new field `uid`, so this should be a minor rather than patch bump.  I know it feels weird given the magnitude of the change...

src/citrine/resources/table_config.py
```diff
@@ -106,6 +107,13 @@ def __init__(self, name: str, *, description: str, datasets: List[UUID],
             raise ValueError("The data_source of the columns must match one of the variable names,"
                              " but {} were missing".format(missing_variables))
 
+    @property
+    def config_uid(self) -> UUID:
+        """[DEPRECATED] Unique ID of the table config, independent of its version."""
+        warnings.warn("The config_uid attribute is deprecated in favor of the uid attribute.",
```
maxhutch, line 113: Does the `deprecated` annotation work, as [here](https://github.com/CitrineInformatics/citrine-python/blob/main/src/citrine/resources/table_config.py#L269)?  I think that will play more nicely with the checking that @bfolie is trying to introduce in https://github.com/CitrineInformatics/citrine-python/pull/654

maxhutch (changes requested):
Excellent first PR!  The change is good and the approach is very comprehensive.  I left a few inline comments, but this is very close.

The test failures are flake8 (see [here](https://app.travis-ci.com/github/CitrineInformatics/citrine-python/jobs/524581208#L629)).

src/citrine/__version__.py
sbernasek: Updated to 1.3.0

maxhutch:
Now that I'm looking at this more carefully, I'm wondering if `config_id` is the id that we want as the `uid`.  I actually think `version_uid` is the more appropriate id to use as the uid, since multiple `TableConfig` objects can share the same `uid` (with different `version_number`) but no two `TableConfigs` will have the same `version_uid`.

That would suggest a few things:
 * Fix the `Project.share()` method by looking for `TableConfig` classes specifically as using their `config_uid`.
 * Don't deprecate anything, since `config_uid` and `version_uid` are more descriptive names than `uid`
 * [TBD] Define a `uid` alias to `version_uid`.  Q: What would be the implications of this?

There's a counter-argument based on symmetry with GemTables, which have a `uid` that corresponds to a `config_uid` rather than a `version_uid` (which seems to be absent).

kroenlein:
It seems like we missed bringing consistency here during the v1 release push.  Ugh.

@maxhutch, the way the IDs around Table Configs were supposed to work was:

* Each Table Config (in all its versions) shares 1 UID
* Each version is just an increasing integer 
* Any query that lacks a version number should default to the most current version

We should put some engineering time into determining if that's how things are actually working.

Note that since we are past v1, we do not permit breaking changes to interface code.

maxhutch:
I chatted with @djack a bit and I think this is what we want to do:
 * Keep the alias of `uid` for `config_uid`.  I'd probably have `uid` alias to `config_uid` instead of the other way around, but either is fine.
 * Don't deprecated `config_uid` yet.  Its a bit more precise and clearer given that there's `version_uid` in there too.
 * Leave the `Project.share` method as is.  Having `uid` alias to `config_uid` will make it work as expected, which was the original point of this PR.

sbernasek:
I reverted all internal references back to TableConfig.config_uid and added a TableConfig.uid property that aliases the config_uid. I also removed the deprecation warning for the time being.

maxhutch (approved):
LGTM!

maxhutch (dismissed):

maxhutch (approved):

PR #656, opened 2021-07-15
show deprecation warnings by default

bfolie:
# Citrine Python PR

## Description 
Python doesn't show DeprecationWarnings to the user, but since we use them to tell users "stop using this feature", we should make it the default.

### PR Type:


kroenlein (approved):
Looks good.

Minor comment: the warn statement itself could have been written without the backslashes.

PR #657, opened 2021-07-15
Bugfix/numpy_dtype_deprecation_warning

sbernasek:
# Citrine Python PR

## Description 
Changes test requirements to require NumPy<1.20 because v>=1.20 deprecates the numpy dtype aliases used by pandas 1.0.3, which we also require, and that discrepancy triggers a deprecation warning that causes our `test_joined_oversize_warnings()` test in `tests/builders/test_design_spaces.py` to fail.

Users may still use NumPy>=1.20, as we do not currently dictate which version of pandas they must use.

### PR Type:


kroenlein:
This could also be addressed by:
1. Updating the required pandas version to >= 1.1.0
2. Correcting the test so that it doesn't get triggered by a deprecation error thrown by another library (i.e., actually fixing the bug)

Why did you pick the approach taken here?

sbernasek:
> This could also be addressed by:
> 
> 1. Updating the required pandas version to >= 1.1.0
> 2. Correcting the test so that it doesn't get triggered by a deprecation error thrown by another library (i.e., actually fixing the bug)
> 
> Why did you pick the approach taken here?

@maxhutch and I briefly discussed both of the above in slack. This was meant to be a temporary fix while we investigate which numpy/pandas version combinations throw this warning.  Generally speaking, I would've thought we want to be made aware of deprecation warnings thrown by other libraries, just not those thrown by calls from those libraries to a secondary dependency of their own (e.g. pandas to numpy).

maxhutch:
If upgrading to pandas 1.1.0 in our requirements files (not setup.py) will keep us from stubbing our toe on test failures, I think that's a great solution.  We don't have a systematic way to keep those up to date, but we typically want those requirements files to be using newer versions of our dependencies.

sbernasek:
Pandas 1.1.0 does indeed resolve the numpy deprecation warning issue. All tests also pass without issue running the latest version of Pandas (1.3.0). I'm inclined to set the test_requirements version to 1.1.0 then allow >=1.1.0 in requirements.txt. 

Does that sound reasonable?

PR #658, opened 2021-07-19
Add uid alias to data concepts that pulls the citrine id.

maxhutch:
# Citrine Python PR

## Description 
This is a convenience method that makes the data concepts resources
look more like the other resources.

### PR Type:


bfolie (approved):
nice

tests/resources/test_material_run.py
```diff
@@ -80,6 +80,8 @@ def test_nomutate_gemd(collection, session):
     # Then
     assert before == after
```
kroenlein, line 81: I'm actually a little concerned about what `==` means in this context.  Is this all the behavior we actually wanted?

kroenlein (approved):
I approve of the immediate change, but I'm not sure I agree with some of the existing mutation logic.  I thought `==` objects were `id() == id()`, which should not work here.

tests/resources/test_material_run.py
maxhutch: I believe that it inherits from [DictSerializable](https://github.com/CitrineInformatics/gemd-python/blob/main/gemd/entity/dict_serializable.py#L138).  Value based equality seems appropriate here.

PR #659, opened 2021-07-21
Bugfix/numpy dtype deprecation

sbernasek:
# Citrine Python PR

## Description 
Changes pandas dependency requirement to Pandas>=1.1.0 in order to avoid numpy dtype alias deprecation warnings thrown by pandas<=1.1 when NumPy>=1.20.

### PR Type:


requirements.txt
```diff
@@ -9,3 +9,4 @@ boto3==1.17.93
 botocore==1.20.95
 deprecation==2.0.7
 urllib3==1.26.5
+pandas>=1.1.0
```
maxhutch, line 12: Why has this been added?

maxhutch (changes requested):
Shouldn't we bump the lower bound in [setup.py](https://github.com/CitrineInformatics/citrine-python/blob/main/setup.py#L35) as well?

requirements.txt
sbernasek: I hadn't seen that we were handling pandas versioning via `extras_require` in setup.py and was intending to shield users from the same numpy vs pandas version discrepancy that was causing my local tests to fail, but I suppose that's outside the scope of our requirements. I'll drop this requirement and upgrade the optional dependency to 1.1

sbernasek: Alternatively, if we don't care about the warning and just want to ensure that our tests pass, we can leave the user requirement for pandas as is (1.0.03 as a builders dependency in setup.py) and just change the test requirement.

maxhutch: I think we should bump the lower bound in the builders dependency and leave it out of here.

sbernasek: Ok, I dropped the pandas dependency in requirements.txt, bumped the lower bound on the builders dependency, and left the test_requirements version as 1.1.0.

maxhutch (approved):

PR #662, opened 2021-08-03
From a list of GEMDs create a table of Attribute Templates

lkubie:
# Citrine Python PR

## Description 
This PR adds in the first of several template-creation helper-functions. Given a list of GEMD objects, this method returns a Pandas DataFrame containing all of the GEMD-values for each of Attributes of the same name and type. This allows a user to see all of the different values and value types associated with the "same" Attribute.

### PR Type:
- [ X] New feature (non-breaking change which adds functionality)

- [X ] I have added tests for 100% coverage
- [ X] I have written Numpy-style docstrings for every method and class.
- [X ] I have communicated the downstream consequences of the PR to others.
- [ X] I have bumped the version in setup.py

### DIFF
--- a/src/citrine/_utils/template_util.py
+++ b/src/citrine/_utils/template_util.py
@@ -1,0 +1,41 @@
+import pandas as pd
+from gemd.entity.object import *
+from gemd.entity.attribute import PropertyAndConditions
+
+def make_attribute_table(gems: list):
+    """[ALPHA] the current status of make_attribute_table
+
+    Given a list of GEMD Objects, this method returns a Pandas DataFrame where each row represents an attribute-containing
+    Object and each column represent a unique Attribute Type + Name pair. The values within the cells are the various GEMD Value Types.
+
+    """
+    types_with_attributes = (
+        ProcessSpec,
+        ProcessRun,
+        MaterialSpec,
+        MeasurementSpec,
+        MeasurementRun
+    )
+    all_rows = []
+    gems = [x for x in gems if isinstance(x, types_with_attributes)]
+    for gem in gems:
+        row_dict = {
+            "object" : gem,
+            "object_type": type(gem).__name__
+        }
+        if hasattr(gem, "conditions"):
+            for cond in gem.conditions:
+                row_dict[f"CONDITION: {cond.name}"] = cond.value
+        if hasattr(gem, "parameters"):
+            for param in gem.parameters:
+                row_dict[f"PARAMETER: {param.name}"] = param.value
+        if hasattr(gem, "properties"):
+            for prop in gem.properties:
+                if isinstance(prop, PropertyAndConditions):
+                    row_dict[f"PROPERTY: {prop.property.name}"] = prop.property.value
+                    for cond in prop.conditions:
+                        row_dict[f"CONDITION: {cond.name}"] = cond.value
+                else:
+                    row_dict[f"PROPERTY: {prop.name}"] = prop.value
+        all_rows.append(row_dict)
+    return pd.DataFrame(all_rows)
--- a/tests/_util/test_template_util.py
+++ b/tests/_util/test_template_util.py
@@ -1,0 +1,129 @@
+from citrine._utils.template_util import make_attribute_table
+from gemd.entity.object import *
+from gemd.entity.attribute import *
+from gemd.entity.value import *
+from gemd.entity.link_by_uid import LinkByUID
+import pandas as pd
+
+def _make_list_of_gems():
+    faux_gems = [
+        ProcessSpec(
+            name = "hello world",
+            parameters = [
+                Parameter(
+                    name = "param 1",
+                    value = NominalReal(nominal=4.2, units="g")
+                ),
+                Parameter(
+                    name = "param 2",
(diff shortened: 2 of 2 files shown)


maxhutch (commented):
Overall this looks great!  I have a couple questions.

* How important is it that this method returns a dataframe instead of `all_rows` (which care trivially built into a data frame)?  Right now, pandas isn't a required dependency but this is in the primary code, so we'd have to add pandas as a required dependency if we want to do this.

* The objects in the `gems` could contain links to other objects.  Does it make sense for this to walk those links (with a call to flatten)?  Or to have an option to load the history from the platform?

* Could you include a little example use snippet in the PR description?  It doesn't need to be actually run-able, but it'd be helpful to see how we expect this to be used.  PR description instead of sphinx docs since its alpha (although you could include it in the docstring).

lkubie:
> Overall this looks great! I have a couple questions.
> 
> * How important is it that this method returns a dataframe instead of `all_rows` (which care trivially built into a data frame)?  Right now, pandas isn't a required dependency but this is in the primary code, so we'd have to add pandas as a required dependency if we want to do this.
> * The objects in the `gems` could contain links to other objects.  Does it make sense for this to walk those links (with a call to flatten)?  Or to have an option to load the history from the platform?
> * Could you include a little example use snippet in the PR description?  It doesn't need to be actually run-able, but it'd be helpful to see how we expect this to be used.  PR description instead of sphinx docs since its alpha (although you could include it in the docstring).

I think removing the DataFrame portion would still be very workable. Even thinking through things I want to do with this table after, it should still work. I will make this change.

I don't think a call to the Platform makes much sense, but I think adding in a flatten would be beneficial. I will add the flatten!

I will also add more info in the docstring so that it's very accessible to all parties.

src/citrine/_utils/template_util.py
```diff
@@ -0,0 +1,61 @@
...
+    ProcessSpec,
+    ProcessRun,
+    MaterialSpec,
+    MeasurementSpec,
+    MeasurementRun,
+)
+from gemd.entity.attribute import PropertyAndConditions
+
+
+def make_attribute_table(gems: list):
```
maxhutch, line 12:
```suggestion
def make_attribute_table(gems: List[DataConcepts]) -> List[Mapping[str, BaseValue]]:
```
Let's make the signature a bit more explicit.  These types will need to be imported too.

src/citrine/_utils/template_util.py
```diff
@@ -0,0 +1,61 @@
...
+    This dictionary can easily be converted into a Pandas DataFrame where there is a
+    row for each Object, and the values within the cells are that Object's type or
+    the various GEMD Value Types.
+
+    This function will help users identify attribute usage across their Objects,
+    and get a sense of the consistency of the data found in their GEMD objects.
+    With the returned dictionary, one can easily do further analysis on both
+    attribute values and value type consistency to assist in data cleaning and
+    Attribute Template creation.
+
```
maxhutch, line 28: Can we add parameters and return value for a full numpy-style docstring (like [here](https://github.com/CitrineInformatics/citrine-python/blob/main/src/citrine/builders/design_spaces.py#L297))?

src/citrine/_utils/template_util.py
```diff
@@ -0,0 +1,61 @@
...
+    """
+    flattened_gems = recursive_flatmap(
```
maxhutch, line 30: Is there a reason you didn't call [`flatten`](https://github.com/CitrineInformatics/gemd-python/blob/main/gemd/util/impl.py#L196)?

tests/_util/test_template_util.py
```diff
@@ -0,0 +1,141 @@
...
+                Property(
+                    name = "prop 1",
+                    value=NominalReal(nominal=4.1, units="")
+                )
+            ]
+        )
+    ]
+    return faux_gems
+
+def test_attribute_alignment():
```
maxhutch, line 131: Could you add some docs here?

maxhutch, line 131: I think it would be good to have a test that lights up the `flatten`.  The simplest unit would be to just have two linked objects (e.g. process run and process spec) and only pass the process run in.

maxhutch (commented):
A couple comments inline.

I still think a usage example would be helpful, even if just in this PR description.

tests/_util/test_template_util.py
lkubie: I do have this in the tests. Line 41 is where I have my nestled Object. The shape assertion confirms it has been flattened. Do you feel this is inadequate or should be more standalone?

src/citrine/_utils/template_util.py
maxhutch: It does (and it does some other things, which you may or may want it to do).  Given that `flatten` is a first class concept, I just thought it was strange to not use it.

tests/_util/test_template_util.py
maxhutch: No, that covers it.  I just didn't see it.

maxhutch (approved):
LGTM!

PR #663, opened 2021-08-04
Modify private method names

kroenlein:
# Citrine Python PR

## Description 
There were a few private methods whose names did not have a leading underscore.  This PR either adds the underscore or inlines the method.

### PR Type:


maxhutch (commented):
All of these functions are inner functions.  I don't think those need `_` since they are not independently reference-able anyways.  In fact, they are more private as inner methods than they would be with `_`

maxhutch:
Semi-related: https://github.com/google/styleguide/blob/gh-pages/pyguide.md#26-nestedlocalinner-classes-and-functions

kroenlein:
Prior art within repo that prepends underscores to inner functions:

https://github.com/CitrineInformatics/citrine-python/blob/ef373cde3b2f9be9bd5657b3e59966f75dc111ff/src/citrine/builders/descriptors.py#L132
https://github.com/CitrineInformatics/citrine-python/blob/ef373cde3b2f9be9bd5657b3e59966f75dc111ff/src/citrine/informatics/reports.py#L179

I believe the underscore conveys scoping intent, independent of whether that method is visible.  I do not think that conflicts PEP/linter behaviors, though I could be mistaken.

maxhutch (approved):

PR #665, opened 2021-08-05
Sketch status fetching methods as an alternative to getters

maxhutch:
# Citrine Python PR

## Description 
This is an alternative to https://github.com/CitrineInformatics/citrine-python/pull/638

### PR Type:


maxhutch:
> Reading over the changes to informatics.**init**, the approach generally makes sense to me if we want to keep the resource objects static, though if that's the case maybe the methods should have a "fetch_" or "get_" prefix.

We could do that.

> Would the intent be to simultaneously deprecate the .status , .status_info, and .status_description resource attributes?

I don't think so.  We want those objects to bind the API, and the API has those members in its responses.  My long term vision is that most users seldom interact with those objects directly and use higher-level abstractions (like the nascent AutoConfigureWorkflow) instead.

PR #669, opened 2021-08-13
Added user prompt in dataset.delete_contents

sdedkins:
# Citrine Python PR

## Description 
Add user prompt in delete_contents so that when users are using this method in an interactive setting ( such as prototyping in a notebook) there is safeguard against accidental invocation.

### PR Type:


src/citrine/resources/dataset.py
```diff
@@ -248,7 +248,8 @@ def delete_contents(
             self,
             *,
             timeout: float = 2 * 60,
-            polling_delay: float = 1.0
+            polling_delay: float = 1.0,
+            prompt_to_confirm: bool = True,
```
kroenlein, line 252: This parameter should be documented below.

src/citrine/resources/dataset.py
```diff
@@ -273,6 +274,18 @@ def delete_contents(
...
+        if prompt_to_confirm:
+            print(f"Do you really want to delete the contents of dataset {self.uid}? [Y/N]")
+            user_response = None
+            while user_response not in {'Y', 'N'}:
+                user_response = input()
+            if user_response == 'Y':
+                pass
+            elif user_response == 'N':
+                print(f'Aborting deletion of dataset {self.uid} contents')
+                return None
```
kroenlein, line 288: This is inconsistent with the return type of _poll_for_async_batch_delete_result.
```suggestion
                return []
```

maxhutch (commented):
Conceptually, this makes sense to me.  Is this code you've actually tested in a jupyter notebook @sdedkins ?  If so, is that the interface you'd like/expect?

sdedkins:
> Conceptually, this makes sense to me. Is this code you've actually tested in a jupyter notebook @sdedkins ? If so, is that the interface you'd like/expect?

The interface seems quite reasonable to me.

Screen Shot 2021-08-13 at 1 04 17 PM

src/citrine/resources/dataset.py
```diff
@@ -273,6 +276,18 @@ def delete_contents(
...
+            user_response = None
+            while user_response not in {'Y', 'N'}:
+                user_response = input()
+            if user_response == 'Y':
+                pass
+            elif user_response == 'N':
+                print(f'Aborting deletion of dataset {self.uid} contents')
+                return []
+            else:
+                raise ValueError(f'Expected user response of Y or N. Got {user_response}')
```
kroenlein, line 290:
```suggestion
        while(prompt_to_confirm):
            print(f"Confirm you want to delete the contents of {self.uid} [Y/N]")
            user_response = input()
            if user_response.lower() in {'y', 'yes'}:
                break  # return to main flow
            elif user_response.lower() in {'n', 'no'}:
                raise RuntimeError("delete_contents was invoked unintentionally")
            else:
                print(f'"{user_response}" is not a valid response')
```

I could see `return []` as possibly more appropriate that the exception, but the rational in PLA-8079 was accidental execution of a Jupyter cell -- in that scenario it seems more correct to stop execution and allow for manual intervention rather than trying to execute the next thing in the cell (or the next cell).  This user verification step should not be relied upon for flow control at the user level.

kroenlein (changes requested):
Generally makes sense.

You also need to bump the version number in `src/citrine/__version__.py` and add tests to `tests/resources/test_dataset.py`.

src/citrine/resources/dataset.py
kroenlein: I'd used `break` instead of `prompt_to_confirm = False` (which would have the same functional effect) because I thought there might be value in knowing what the passed parameter was, or at least there was no benefit in clobbering that value.

tests/resources/test_dataset.py
```diff
@@ -367,7 +367,7 @@ def test_gemd_batch_delete(dataset):
     with pytest.raises(TypeError):
         dataset.gemd_batch_delete([7, False])
 
-def test_delete_contents_ok(dataset):
+def test_delete_contents_ok(dataset,monkeypatch):
```
kroenlein, line 370:
```suggestion

def test_delete_contents_ok(dataset, monkeypatch):
```

tests/resources/test_dataset.py
```diff
@@ -398,4 +401,10 @@ def test_delete_contents_ok(dataset):
         path='projects/{}/datasets/{}/contents'.format(dataset.project_id, dataset.uid)
     )
     assert len(session.calls) == 2
-    assert session.calls[0] == expected_call
\ No newline at end of file
+    assert session.calls[0] == expected_call
+
+def test_delete_contents_abort(dataset,monkeypatch):
```
kroenlein, line 406:
```suggestion

def test_delete_contents_abort(dataset,monkeypatch):
```

tests/resources/test_dataset.py
```diff
@@ -386,8 +386,11 @@ def test_delete_contents_ok(dataset):
     session = dataset.session
     session.set_responses(job_resp, failed_job_resp)
 
+    user_responses = iter(['bad user response', 'Y'])
+    monkeypatch.setattr('builtins.input', lambda : next(user_responses))
```
kroenlein, line 390:
```suggestion
    monkeypatch.setattr('builtins.input', lambda: next(user_responses))
```

kroenlein (approved):
Minor formatting changes.  Implementation looks good.

kroenlein (approved):
LGTM

PR #671, opened 2021-08-13
Check when uid=None for table_configs.get

sfriedowitz:
## Description 

This is the default behavior on other collections, but `TableConfigCollection` overrides this method and the protection was left off.

I found that the product API request still goes through uid=None, but returns faulty data, so this is good to have in place.


kroenlein (dismissed):
LGTM

kroenlein (approved):

PR #672, opened 2021-08-14
Added missing numpy import for resp_to_pandas in Data Extraction docs

sdedkins:
# Citrine Python PR

## Description 
Add missing import statement to code snippet in docs. 

### PR Type:


docs/source/data_extraction.rst
```diff
@@ -150,6 +150,7 @@ or even wrap it in a method that displays multi-row headers:
         import warnings
         from io import StringIO
         import pandas as pd
+        import numpy as np
```
maxhutch, line 153:
```suggestion
        import numpy as np
        
```
Suggestion: I think it looks better if we keep the whitespace.

src/citrine/__version__.py
```diff
@@ -1 +1 @@
-__version__ = '1.6.2'
+__version__ = '1.7.1'
```
maxhutch, line 1: Why the jump to 1.7.1?

maxhutch (changes requested):
Nice catch!  The only blocking issue is that the version jump is too big.

src/citrine/__version__.py
sdedkins: I was anticipating merging the my PR with a minor version bump before this one. I anticipated incorrectly :-)

maxhutch (approved):

PR #675, opened 2021-08-17
Clarify how regression and classification behave

maxhutch:
# Citrine Python PR

## Description 

They are auto-selected based on the descriptor type.  Also remove
the restriction that F1 is only computed for multi-class since
we support it for binary now too.

### PR Type:


sfriedowitz (approved):
LGTM. Nice clarifications

PR #676, opened 2021-08-17
Checkpoint: pull existing fakes out into fakes module

maxhutch:
# Citrine Python PR

## Description 
There are lots of fakes already in the test code.  This just
extracts some of them.

### PR Type:

### DIFF
--- a/tests/seeding/test_find_or_create.py
+++ b/tests/seeding/test_find_or_create.py
@@ -12,6 +12,9 @@ from citrine.resources.project import ProjectCollection, Project
 from citrine.seeding.find_or_create import (find_collection, get_by_name_or_create, get_by_name_or_raise_error,
                                             find_or_create_project, find_or_create_dataset,
                                             create_or_update)
+from tests.utils.fakes.fake_dataset import FakeDatasetCollection
+from tests.utils.fakes.fake_predictor import FakePredictorCollection
+from tests.utils.fakes.fake_project import FakeProjectCollection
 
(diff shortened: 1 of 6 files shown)


src/citrine/builders/design_spaces.py
```diff
@@ -329,9 +329,8 @@ def migrate_enumerated_design_space(*,
 
     if cleanup:
         # archive the old enumerated design space
-        enumerated_ds.archived = True
         try:
-            project.design_spaces.update(enumerated_ds)
+            project.design_spaces.archive(enumerated_ds.uid)
```
maxhutch, line 333: This change makes the method more robust and more idiomatic.  Its also easier to fake in the tests.

tests/seeding/test_find_or_create.py
```diff
@@ -349,7 +279,6 @@ def test_create_or_update_unique_found(predictor_collection):
                             latent_variables = [])
     #verify that the returned object is updated
     returned_pred = create_or_update(collection=predictor_collection, resource=pred)
-    assert returned_pred.uid == pred.uid
```
maxhutch, line 352: The behavior of the non-fake client is that it doesn't modify the object that was passed in, so this was testing fake behavior that's inconsistent with the real behavior.

sfriedowitz (approved):
This looks great! And will be easy to continue adding to for specific module methods, and `FakeDataConceptsCollection`.

PR #678, opened 2021-08-19
Bugfix: build GemTable with 'id' not a 'uid'

maxhutch:
# Citrine Python PR

## Description 

In JSON, GemTable has an `id` field, not a `uid` field.
The uid -> id mismatch was causing uid to be empty (since its
optional) and therefore the GemTable wasn't resolvable.

### PR Type:


sfriedowitz (approved):

kroenlein:
Why wasn't this caught by existing tests?

sfriedowitz:
> Why wasn't this caught by existing tests?

I suspect it was due to the fact that GemTable.build({"uid": ...}) does not throw an error in itself. It just produces a GemTable without a UID field.

The error would be thrown only when trying to use that UID for access. The test `test_get_and_read_table_from_collection` is what covered that line, but the mocking probably covered that error up.

maxhutch:
> I suspect it was due to the fact that `GemTable.build({"uid": ...})` does not throw an error in itself.

Those fields are all declared as optional, but I'm not sure why.  My best guess is that it made testing easier.  I think we can recover some coverage by making them required.

PR #684, opened 2021-09-13
GEMD Builders CSV prototype

kroenlein:
# Citrine Python PR
Draft PR for API-driven CSV ingest.  Feeback appreciated, tests not quite passing.

## Description 
This is initial work to incorporate annotated-CSV driven workflows into citrine python.  As implemented, this uses the header keywords that DE has been using in internal tooling as the basis for a file-driven ingest.  This layer is expected to be abstracted so that the underpinning workflow mimics our proposed UI-driven workflows.

### PR Type:

### DIFF
--- a/src/citrine/seeding/parse_csv.py
+++ b/src/citrine/seeding/parse_csv.py
@@ -1,0 +1,450 @@
+from gemd.entity.object import MaterialRun, MaterialSpec, ProcessRun, ProcessSpec, \
+    MeasurementRun, MeasurementSpec, IngredientRun, IngredientSpec
+from gemd.entity.object.base_object import BaseObject
+from gemd.entity.value import NominalReal, NominalCategorical, NominalInteger, \
+    InChI, Smiles, EmpiricalFormula
+from gemd.entity.template import MaterialTemplate, ProcessTemplate, MeasurementTemplate, \
+    PropertyTemplate, ConditionTemplate, ParameterTemplate
+from gemd.entity.template.base_template import BaseTemplate
+from gemd.entity.template.attribute_template import AttributeTemplate
+from gemd.entity.bounds import RealBounds, CategoricalBounds, IntegerBounds, \
+    MolecularStructureBounds, CompositionBounds
+from gemd.entity.attribute import Property, Condition, Parameter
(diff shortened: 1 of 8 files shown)


src/citrine/seeding/parse_csv.py
```diff
@@ -0,0 +1,450 @@
...
+                labels=list(label_list),
+            )
+
+            # TODO: Option for mass/number/volume/absolute
+            ing_fraction = NominalReal(nominal=amount / total, units="")
+            ing_run = IngredientRun(  # noqa: F841  Easier to follow w/ unused variable
+                material=mat_index[upstream],
+                spec=ing_spec,
+                process=mat_index[downstream].process,
+                mass_fraction=ing_fraction,
```
sfriedowitz, line 447: It would be nice to have an option to choose which quantity the columns are indicating. Is this something that should be defined in a header annotation? Or a method argument on the builder?

sfriedowitz (commented):
What are your thoughts on wrapping these tools in a stateful builder tool?

I like the utility of leaving them available for a more specific use case. But for the general user, I was imagining something along the lines of:

```
builder = GEMDBuilder("list of CSV files/data frames")
templates = builder.guess_templates(header_annoations="Supply some of your own here if needed")
objects = builder.create_gemd(ingredient_quantity=MASS FRACTION, etc. etc.)
```

If this isn't the immediate goal, it would be straightforward to add on in a follow-up.

kroenlein:
> What are your thoughts on wrapping these tools in a stateful builder tool?

Not opposed, but first up I wanted feature parity with existing tool sets.  I can see benefits to both sides on the question of a config object vs. notebook-style call chain.

ventura-rivera:
Just a thought looking forward: The keywords in here are specific to PIF and the Citrination platform. This is the format used for the pif-to-gemd ingester, but in the future, it would be great to incorporate synonyms to use that will be more relevant to GEMD and the Citrine Platform.

kroenlein:
Closing this draft because it's largely been subsumed by [GEMDIngest](https://github.com/CitrineInformatics/gemd-ingest)

PR #685, opened 2021-09-24
Feature/pla 8321 citrine python version header

crleblanc:
# Citrine Python PR

## Description 
This PR uses a custom user-agent header when making requests, as per PLA-8321.  This will let us identify the version of requests, Python and Citrine Python being used by our clients.  This header will be in the form of 'python-requests/2.26.0 CPython/3.7.10 citrine-python/1.9.0'.

This attempts to conform to the MDN guidelines, but there is a lot of variability between browsers/crawlers/bots: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent.  Using a space delimited format makes this easy to parse.

We only filter on the user-agent header in one place (https://github.com/CitrineInformatics/shire/blob/dc0966842e7e53a10658df643e3219ae21207069/services/pippin/src/middlewares/requireProjectPermissions.ts). This user-agent change won't impact the way this middleware works.

### PR Type:


src/citrine/_session.py
```diff
@@ -42,8 +44,14 @@ def __init__(self,
         self.access_token: Optional[str] = None
         self.access_token_expiration: datetime = datetime.utcnow()
 
+        agent = "python-requests/{} {}/{} citrine-python/{}".format(
+            requests.__version__, platform.python_implementation(), platform.python_version(),
+            cp_version)
```
kroenlein, line 49: From a hierarchy perspective, it seems like
```suggestion
        agent = "{}/{} python-requests/{} citrine-python/{}".format(
            platform.python_implementation(), 
            platform.python_version(),
            requests.__version__,
            cp_version)
```
would be more consistent.  Optional.

kroenlein (approved):
Minor optional tweak.

dleblanc (approved):

PR #686, opened 2021-09-25
Removed empty file test_session.py and added a test for the user-agent

crleblanc:
# Citrine Python PR

## Description 
Bugfix PR.  There is a strange bug related to the difference between `import citrine; citrine.__version__` and `from citrine import __version__ as cp_version`.  The latter prints out fine on the console but not in our logs.  The former works as expected in our logs.

The PR adds a unit test for the user agent string and removes an empty file called test_sesison.py that was incorrectly added in the past by yours truly.

### PR Type:


maxhutch (approved):
Change looks good.

Do we know what the difference between those two import modes is?

crleblanc:
> Do we know what the difference between those two import modes is?

We're currently seeing this string output as the user-agent: `CPython/3.6.15 python-requests/2.26.0 citrine-python/<module 'citrine.__version__' from '/usr/local/lib/python3.6/site-packages/citrine/__version__.py'>`.

Instead of using Citrine's version number string defined in __version__.py it appears to be including the string produced from `repr(citrine.__version__)`.  This is a strange bug, printing this on the console with Python 3.6.9 gives "'1.9.0'" but not when we send it as a header.

The module is named `__version__.py` and the version string is also named `__variable__`.  It appears we could get the expected string using this import: `from citrine.__version__ import __version__ as cp_version`.  I think that's quite a bit uglier than using `citrine.__version__`.  I wonder if using this approach invokes a bit of Python import magic that the other import approach bypassed?  This rejected PEP describes some issues with __version__: https://www.python.org/dev/peps/pep-0396/.

crleblanc:
Rebuild!

PR #690, opened 2021-10-06
Fix the flakiness of test_functions.py::test_shadow_classes_in_module

shenganzhang:
# Citrine Python PR

## Description 
This PR aims to fix flaky `tests/_util/test_functions.py::test_shadow_classes_in_module`. In previous versions, the test will run into failure when running for multiple times. And the reason is that the status of `target_mod` module is not cleared. The test failure can be reproduced by
`pip install pytest-flakefinder`
`pytest --flake-finder --flake-runs=2 tests/_util/test_functions.py::test_shadow_classes_in_module`
Notice that the PR is modifying the test to make it more robust without changing the source code.

### PR Type:

- [ ] I have added tests for 100% coverage (no change in source code).
- [ ] I have written Numpy-style docstrings for every method and class (no new methods nor classes). 
- [ ] I have bumped the version in setup.py (no library updated).


shenganzhang:
> You update the version file, [here](https://github.com/CitrineInformatics/citrine-python/blob/main/src/citrine/__version__.py). This type of change corresponds to a [patch version bump](https://semver.org/).

I've updated the version to 1.10.1

bfolie (approved):

PR #692, opened 2021-10-11
PLA-8251: Make autoconfigure function support INFER pattern

dkao-citrine:
# Citrine Python PR

## Description 
In https://citrine.atlassian.net/browse/PLA-8050, we made the INFER pattern available in the product API. This third option allows the `auto_configure` function to scan the attached `DataSource` for a `FormulationsDescriptor`. If it finds at least one, then it infers that the data is meant for training a formulation predictor and creates a formulation predictor. If there's no such descriptor to be found, then it infers that the data is training a plain predictor and creates one.

This pull request makes this option available to citrine-python users.

Here is a passing Devkit-custom job that ran all of the existing `autoconfigure()` tests and successfully exercised this new `INFER` pattern:
https://jenkins.corp.citrine.io/job/Devkit-custom/572/

### PR Type:

### Remaining Tasks (see discussion below):
- [ ] Although changing the default pattern from `PLAIN` to `INFER` will not break any existing scripts or workflows, we still need to communicate this change to the Customer Team so they are aware of it.
- [ ] Hold the valid patterns in a Python `Enum` instead of a set for greater flexibility. Examples:
https://docs.python.org/3.11/howto/enum.html
https://github.com/CitrineInformatics/citrine-python/blob/main/src/citrine/_rest/resource.py#L9
https://github.com/CitrineInformatics/citrine-python/blob/main/src/citrine/gemtables/columns.py#L25

### DIFF
--- a/src/citrine/resources/predictor.py
+++ b/src/citrine/resources/predictor.py
@@ -18,6 +18,7 @@ class AutoConfigureMode(BaseEnumeration):
 
     * PLAIN corresponds to a single-row GEM table and plain predictor
     * FORMULATION corresponds to a multi-row GEM table and formulations predictor
+    * INFER auto-detects the GEM table and predictor type
     """
 
     PLAIN = 'PLAIN'


src/citrine/__version__.py
```diff
@@ -1 +1 @@
-__version__ = '1.10.0'
+__version__ = '1.10.1'
```
kroenlein, line 1:
```suggestion
__version__ = '1.11.0'
```
You are changing a default behavior which means this is a potentially breaking change.  Even if you don't change the default, you are still arguably making an API change, which would still require a minor bump.

src/citrine/resources/predictor.py
```diff
@@ -75,7 +75,7 @@ def check_for_update(self, uid: Union[UUID, str] = None,
         else:
             return None
 
-    def auto_configure(self, *, training_data: DataSource, pattern="PLAIN") -> Predictor:
+    def auto_configure(self, *, training_data: DataSource, pattern="INFER") -> Predictor:
```
kroenlein, line 78: Why is changing the default behavior motivated?  In theory, this could be breaking and we aren't supposed to make any breaking changes until 2.0.  Is this feature still in Alpha?

src/citrine/resources/predictor.py
```diff
@@ -87,16 +87,20 @@ def auto_configure(self, *, training_data: DataSource, pattern="PLAIN") -> Predi
...
 
         Returns
         -------
         Predictor
             Automatically configured predictor for the training data
 
         """
-        if pattern not in {"FORMULATION", "PLAIN"}:
-            msg = "Called with pattern: {}.  Expected 'FORMULATION' or 'PLAIN'.".format(pattern)
+        if pattern not in {"FORMULATION", "PLAIN", "INFER"}:
```
kroenlein, line 101: Why are you not creating an enumeration here?  That would solve the user problem of guessing capitalization, etc.

kroenlein (commented):
Sorry if I'm missing context.

src/citrine/resources/predictor.py
```diff
@@ -87,16 +87,20 @@ def auto_configure(self, *, training_data: DataSource, pattern="PLAIN") -> Predi
         training_data: DataSource
             The data to configure the predictor to model.
         pattern: str
-            The predictor pattern to use, either "PLAIN" or "FORMULATION"
+            The predictor pattern to use, either "PLAIN", "FORMULATION", or "INFER".
+            The "INFER" pattern auto-detects whether the `DataSource` is formulations data or not.
+            If it is, then a formulation predictor is created.
+            If not, then a plain predictor is created.
```
bfolie, line 93:
```suggestion
            The "INFER" pattern auto-detects whether the `DataSource` contains formulations data.
            If it does, then a formulation predictor is created.
            If not, then a plain predictor is created.
```

tests/resources/test_predictor.py
```diff
@@ -300,3 +301,39 @@ def test_returned_predictor(valid_graph_predictor_data):
     assert len(result.predictors) == 2
     assert isinstance(result.predictors[0], uuid.UUID)
     assert isinstance(result.predictors[1], ExpressionPredictor)
+
+def test_returned_plain_predictor(valid_simple_ml_predictor_data):
```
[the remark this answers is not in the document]
dkao-citrine: Ah, yes. I wasn't sure how to communicate this part. Because we mock out the response no matter what, whether you set the pattern to `PLAIN`, `INFER`, or `FORMULATION`, this test and the tests below will always pass. It doesn't matter what kind of predictor is in the mocked response. I tried it out during development. You are absolutely right about all of that.

Now, it feels small, but I think what the team was after here was making sure that `autoconfigure` could recognize all three of these patterns, `PLAIN`, `INFER`, and `FORMULATION`, and _nothing else_. The nothing else is tested above where we tried feeding `autoconfigure` the non-existent `yogurt` pattern:

https://github.com/CitrineInformatics/citrine-python/blob/main/tests/resources/test_predictor.py#L277

`PLAIN` was exercised and accepted in the happy path. We wanted to make sure the other two patterns were accepted with no hassle, as well. Since we got this far, I thought I might as well make the predictor type in the mocked response match the pattern being exercised. But I agree that the tests as written are coming across as confusing. We could:

1. Change the test names to `test_returned_predictor` (reversion), `test_simple_pattern`, `test_formulation_pattern`. Change the docstrings to `"""Exercise auto_configure PLAIN pattern"""` and `"""Exercise auto_configure FORMULATION pattern"""`,  or
2. Revert my updates entirely. 

I prize readability and am fine with either, so let me know what feels best to you as a programmer.

src/citrine/resources/predictor.py
kroenlein: I'm okay with the above rational.  Sounds like this is not breaking a workflow.

tests/resources/test_predictor.py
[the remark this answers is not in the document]
bfolie: If you just want to test that an argument is accepted without error, then the minimum test to do that would be something like:
```
session = FakeSession()
pc = PredictorCollection(uuid.uuid4(), session)
pc.auto_configure(training_data=GemTableDataSource(table_id=uuid.uuid4(), table_version=0), pattern="INFER")
```

But this is another reason to use an enumeration. With an enumeration, that guarantee is baked in.

src/citrine/resources/predictor.py
talltea: Turns out, we even already have an AutoConfigure Enumeration hanging around elsewhere (thanks to Sean F!) https://github.com/CitrineInformatics/citrine-python/blob/main/src/citrine/builders/auto_configure.py#L28

sfriedowitz: Yes! Please do use that Enum, and move it to a more appropriate location to be used as a standard in the code-base.

I was very annoyed that we had three different ways of matching `FORMULATION` vs `PLAIN` vs etc etc..

bfolie (approved):

src/citrine/resources/predictor.py
```diff
@@ -11,6 +13,18 @@
 CreationType = TypeVar('CreationType', bound=Predictor)
 
 
+class AutoConfigureMode(BaseEnumeration):
+    """[ALPHA] The format to use in building auto-configured assets.
+
+    * PLAIN corresponds to a single-row GEM table and plain predictor
+    * FORMULATION corresponds to a multi-row GEM table and formulations predictor
```
bfolie, line 20: Please add a line for INFER

talltea: Added. (words are hard though, lmk if there's a more helpful tagline)

PR #699, opened 2021-11-14
Pla 8400 accounts v3 team

talltea:
# Citrine Python PR

## Description 
This adds the Team collection. Teams are new as of the CP2 release of the Citrine Platform Web Interface.

### PR Type:


src/citrine/citrine.py
```diff
@@ -38,3 +39,8 @@ def projects(self) -> ProjectCollection:
     def users(self) -> UserCollection:
         """Return the collection of all users."""
         return UserCollection(self.session)
+
+    @property
+    def teams(self) -> TeamCollection:
+        """Returns a resource representing all visible teams."""
```
talltea, line 45: don't return this if not in accounts v3

src/citrine/resources/team.py
```diff
@@ -0,0 +1,200 @@
...
+        Time the team was created, in seconds since epoch.
+
+    """
+
+    _response_key = 'team'
+    _resource_type = ResourceTypeEnum.TEAM
+    _api_version = "v3"
+
+    name = properties.String('name')
+    description = properties.Optional(properties.String(), 'description')
```
talltea, line 68: look into nullable descriptions on update

src/citrine/resources/team.py
```diff
@@ -0,0 +1,200 @@
...
+        return '<Team {!r}>'.format(self.name)
+
+    def _path(self):
+        return format_escaped_url('/teams/{team_id}', team_id=self.uid)
+
+    def list_members(self) -> List[TeamMember]:
+        """List the Team Members."""
+        members = self.session.get_resource(self._path() + "/users",
+                                            version=self._api_version)["users"]
+        return [TeamMember(id=m["id"], screen_name=m["screen_name"], email=m["email"],
```
talltea, line 91: instead of a new TeamMembers object, see if we can make Users backwards compatible?

talltea:
Successful run of the nextgen devkit e2es against this branch in https://jenkins.corp.citrine.io/job/Devkit-custom/578/

src/citrine/exceptions.py
```diff
@@ -115,6 +115,12 @@ class PollingTimeoutError(NonRetryableException):
     pass
 
 
+class AccountsV3Exception(CitrineException):
```
talltea, line 118: thinking about this again, I think we don't need this specific error, and all cases where I would use it would be better served with other, existing errors (Not implemented and similar)

dkao-citrine: Sounds good! The simpler, the better.

src/citrine/citrine.py
```diff
@@ -38,3 +40,11 @@ def projects(self) -> ProjectCollection:
...
+        if self.session._accounts_service_v3:
+            return TeamCollection(self.session)
+        else:
+            raise AccountsV3Exception("Teams are not available, please continue using projects")
```
dkao-citrine, line 50: Just curious about how this flag will work in practice -- so when a Python SDK user secretly knows about and wants to use these v3 features, would they manually initialize the session with `_account_service_v3 = True`?

src/citrine/resources/team.py
```diff
@@ -0,0 +1,291 @@
...
+class TeamMember:
+    """A Member of a Team."""
+
+    def __init__(self,
+                 *,
+                 user: User,
+                 team: 'Team',  # noqa: F821
+                 actions: TEAM_ACTIONS):
+        self.user = user
+        self.team: 'Team' = team  # noqa: F821
```
dkao-citrine, line 27: Was flake8 just being annoying here (and on line 24)?

src/citrine/citrine.py
talltea: This flag gets updated every time the user's token is refreshed by a call to the product api's config endpoint. A particularly motivated user can definitely update that private property, but it wouldn't do them any good since the backend services wouldn't have been updated to understand v3 calls.

src/citrine/resources/team.py
talltea: There's a recursive import loop here, because a TeamMember has a Team they are part of, and a Team method `list_members` returns a TeamMember. This pattern is similar to the ProjectUser pattern we have elsewhere

tests/resources/test_team.py
```diff
@@ -0,0 +1,242 @@
...
+    return FakePaginatedSession()
+
+
+@pytest.fixture
+def team(session) -> Team:
+    team = Team(
+        name='Test Team',
+        session=session
+    )
+    team.uid = uuid.UUID('16fd2706-8baf-433b-82eb-8c7fada847da')
```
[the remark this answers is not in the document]
talltea: naw, just copy/pasta from already existing tests 🤷

src/citrine/resources/team.py
```diff
@@ -0,0 +1,291 @@
...
+                 name: str,
+                 *,
+                 description: str = "",
+                 session: Optional[Session] = None):
+        self.name: str = name
+        self.description: str = description
+        self.session: Session = session
+
+    def __str__(self):
+        return '<Team {!r}>'.format(self.name)
```
dkao-citrine, line 82: Nice

src/citrine/resources/team.py
```diff
@@ -0,0 +1,291 @@
...
+        -------
+        List[TeamMember]
+            The members of the current team
+
+        """
+        members = self.session.get_resource(self._path() + "/users",
+                                            version=self._api_version)["users"]
+        return [TeamMember(user=User.build(m), team=self, actions=m["actions"]) for m in members]
+
+    def remove_user(self, user_id: Union[str, UUID]) -> bool:
```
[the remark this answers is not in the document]
talltea: Mostly because single user updates was the pattern in the previously used `projects` methods, and I stuck with it. We can add a `remove_users` method that can do many users at once (same for the other multi action methods)

src/citrine/citrine.py
dkao-citrine: Aaah okay, that makes more sense. Thank you!

src/citrine/resources/team.py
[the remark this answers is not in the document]
dkao-citrine: If single user updates is the existing pattern, then we should stick to it. Product can request multi-action methods later if there's a demand.

talltea:
https://jenkins.corp.citrine.io/job/Devkit-custom/579/console < the latest using this branch
I will bump the version later tonight, and the PR builds will pass too 🤞

src/citrine/resources/team.py
```diff
@@ -0,0 +1,293 @@
...
+        self.session.checked_post(self._path() + "/users/batch-remove",
+                                  json={"ids": [str(user_id)]}, version=self._api_version)
+        return True  # note: only get here if checked_post doesn't raise error
+
+    def add_user(self, user_id: Union[str, UUID], *,
+                 actions: Optional[List[TEAM_ACTIONS]] = None) -> bool:
+        """
+        Add a User to a Team.
+
+        Adds User with ``READ`` action to the Team.
```
anoto-moniz, line 125: Nitpick: Not if the caller provides other actions, as noted below. The caller doesn't need to call update_user_action if they use that parameter.

anoto-moniz (approved):

dkao-citrine (approved):

src/citrine/resources/team.py
talltea: I'll fix this docstring in the followup pr

src/citrine/citrine.py
```diff
@@ -38,3 +39,11 @@ def projects(self) -> ProjectCollection:
...
         """Return the collection of all users."""
         return UserCollection(self.session)
+
+    @property
+    def teams(self) -> TeamCollection:
+        """Returns a resource representing all visible teams."""
+        if self.session._accounts_service_v3:
+            return TeamCollection(self.session)
+        else:
+            raise NotImplementedError("Teams are not available, please continue using projects")
```
crleblanc, line 49: I like this approach

crleblanc (approved):
I see some other excellent PR comments/discussion.  This looks really good to me.

PR #700, opened 2021-11-16
Reconcile gemd & citrine serialiations

kroenlein:
# Citrine Python PR

## Description 
This PR reconciles some friction between the gemd-python and citrine-python serialization strategies.  It also repairs an issue where `register_all` was corrupting object uids in dry-run mode.

This PR depends on [gemd-python PR-157](https://github.com/CitrineInformatics/gemd-python/pull/157) and thus must be merged after that and after gemd-python is published to pypi.

### PR Type:


src/citrine/resources/data_concepts.py
```diff
@@ -279,21 +281,53 @@ def get_json_support(cls):
...
+            if serialized is not None:
+                result[field] = serialized
+        return result
+
     def as_dict(self) -> dict:
         """
         Dump to a dictionary (useful for interoperability with gemd).
 
+        Note that something in the serialization stack changes the result df __dict__ dramatically
+        between gemd.entity.dict_serializable and this class.
```
dleblanc, line 308: Accidental reordering of comment?

kroenlein: Yes.  Fixed.

tests/resources/test_gemd_resource.py
```diff
@@ -469,3 +472,96 @@ def test_batch_delete(gemd_collection, session):
 def test_batch_delete_bad_input(gemd_collection):
     with pytest.raises(TypeError):
         gemd_collection.batch_delete([False])
+
+
+def test_type_passthrough(gemd_collection, session):
```
sfriedowitz, line 477: I think these two tests need some comments explaining what they're doing. It is far from clear to the naive reader.

src/citrine/resources/data_concepts.py
```diff
@@ -279,21 +281,50 @@ def get_json_support(cls):
             )
         return cls.json_support
 
+    def dump(self) -> dict:
```
sfriedowitz, line 284: Is this calling `dump` from the inherited GEMD objects, and then just filling in the gaps with client specific fields in citrine python?

That seems logical for reconciling the different serialization strategies, but maybe I am interpreting this wrong.

src/citrine/_serialization/properties.py
```diff
@@ -156,7 +156,11 @@ def __set__(self, obj, value: typing.Union[SerializedType, DeserializedType]):
             value_to_set = self.deserialize(value, base_class=base_class)
 
         if base_class is not None and self.override:
-            getattr(base_class, self.serialization_path).fset(obj, value_to_set)
```
sfriedowitz, line 159: This seems to be the hot spot that caused today's dev0. Maybe add some comment here clarifying why the change is being made?

sfriedowitz (commented):
A few minor comments requesting some added comments.

fportillo92 (approved):
Hopefully this solves the dev0

src/citrine/resources/data_concepts.py
kroenlein: Yes, which is necessary so that when you copy a citrine-python reflection of a gemd-python object, you maintain the `audit_info` and `dataset` attributes as populated.

tests/resources/test_gemd_resource.py
kroenlein:
```suggestion
def test_type_passthrough(gemd_collection, session):
    """Verify objects that are not directly referenced by objects (e.g., a tuple of Templates) don't get type information stripped."""
```

src/citrine/_serialization/properties.py
kroenlein: Done

src/citrine/_serialization/properties.py
```diff
@@ -156,7 +156,11 @@ def __set__(self, obj, value: typing.Union[SerializedType, DeserializedType]):
...
+            prop = getattr(base_class, self.serialization_path)
+            if prop.fset is not None:
```
kroenlein, line 160:
```suggestion
            if prop.fset is not None:  # It's a property with a setter
```

fportillo92 (approved):

src/citrine/resources/data_concepts.py
```diff
@@ -439,12 +469,33 @@ def register(self, model: ResourceType, *, dry_run=False):
...
+            # this might leave objects with just the temp ids, which we want to strip later
+            if CITRINE_SCOPE not in model.uids:
+                registered.uids.pop(CITRINE_SCOPE, None)
+            if registered.tags is not None:
+                todo = [tag for tag in registered.tags
+                        if re.match(f"^{CITRINE_TAG_PREFIX}::", tag)]
+                for tag in todo:  # Covering this block would require dark art
+                    if tag not in model.tags:
+                        registered.tags.remove(tag)
+        return registered
```
andyczerwonka, line 498: @kroenlein any chance you remember the context of this tag manipulation use case? It's fine if "no"... I'm just trying to trace back the `citr_auto` use cases.

kroenlein: The platform returns registered versions of all objects during the register_all operation.  However, because of order of operations, if enough materials are passed so that submissions must be batched, all material_runs returned by the platform will be annotated as terminal materials, since they are all registered before the ingredients that would reference them.

kroenlein: The SDK still batches into groups of 50 because of service limits.  Is that no longer the case?

We used the full `citr_auto::` rather than specifically terminal checks because we figured it was not a user-supplied value, so it probably should be that user facing.

andyczerwonka: So the SDK adds the appropriate terminal material designation to all objects and then it batches after that?

PR #701, opened 2021-11-17
PLA 8395 accounts v3 projects

talltea:
# Citrine Python PR

## Description 
Note: I haven't done the last proofreading pass on this yet, sorry if there are typos/missing docs. Feel free to point them out if you see them, I'll fix! 

Updates the Project collection to work with Accounts V3. This includes:
1. updating the `_check_accounts_version` call to happen on creation of a Session. This required a bunch of changes to the tests. (That's what all the new mock calls are about)
2. A bunch of sharing methods in projects are no longer available in accounts v3. I think there are ways to make the new ones work under the hood, but let's pull that out into a different PR if we decide we want it.
3. I've expanded the TeamCollection to have a projects property so that users can call things like `team.projects.get('some_project_id')`
4. Registering a new team automatically checks your user id and grants you read, write, share permissions to that team
5. There's a `find_or_create_team` method.

### PR Type:


talltea:
The custom devkit job is running here: https://jenkins.corp.citrine.io/job/Devkit-custom/581/

src/citrine/resources/project.py
```diff
@@ -82,15 +84,36 @@ class Project(Resource['Project']):
...
+
+    def _post_dump(self, data: dict) -> dict:
+        non_empty_data = {}
+        for key, value in data.items():
+            if value is not None:
+                non_empty_data[key] = value
+
+        return non_empty_data
+
+    # def _post_dump(self, data: dict) -> dict:
```
talltea, line 108: woops. will remove this obviously

tests/seeding/test_find_or_create.py
```diff
@@ -67,6 +70,20 @@ def _make_project(search_implemented: bool = True):
     return _make_project
 
 
+@pytest.fixture
+def team_collection() -> Callable[[bool], TeamCollection]:
+
+    def _make_team(search_implemented: bool = True):
```
anoto-moniz, line 76: In each case you use this fixture, you don't pass any arguments on invocation. Any reason to not flatten it out?

tests/seeding/test_find_or_create.py
```diff
@@ -152,6 +169,32 @@ def test_get_by_name_or_raise_error_exist(fake_collection):
...
+    collection = team_collection()
+    old_team_count = len(list(collection.list()))
+    result = find_or_create_team(team_collection=collection, team_name="team 2")
+    new_team_count = len(list(collection.list()))
+    assert result.name == "team 2"
+    assert new_team_count == old_team_count
+
+
+def test_find_or_create_raise_error_team_no_exist(team_collection):
+    # test when tea, doesn't exist and raise_error flag is on
```
anoto-moniz, line 193: s/tea,/team/

tests/resources/test_project.py
```diff
@@ -45,15 +57,40 @@ def project(session) -> Project:
...
+    return project
+
+
 @pytest.fixture
 def collection(session) -> ProjectCollection:
     return ProjectCollection(session)
 
 
+@pytest.fixture
+def collection_v3(session_v3,) -> ProjectCollection:
```
anoto-moniz, line 77: Stray comma.

tests/resources/test_project.py
```diff
@@ -45,15 +57,40 @@ def project(session) -> Project:
...
+
+
 def test_string_representation(project):
     assert "<Project 'Test Project'>" == str(project)
 
 
+def test_share_post_content_v3(project_v3):
+    # Given
+    dataset_id = str(uuid.uuid4())
+    other_team_id = uuid.uuid4()
```
anoto-moniz, line 88: Why is one of these a str and the other a UUID?

tests/resources/test_project.py
```diff
@@ -103,6 +140,108 @@ def test_share_post_content(project, session):
...
+    dataset = project_v3.datasets.build(dict(
+        id=dataset_id,
+        name="public dataset", summary="test", description="test"
+    ))
+    assert project_v3.pull_in_resource(resource=dataset)
+
+    assert 1 == session_v3.num_calls
+    expected_call = FakeCall(
+        method='POST',
+        path='/teams/{}/projects/{}/outside-resources/{}/batch-pull-in'.format(project_v3.team_id, project_v3.uid, 'DATASET'),
```
anoto-moniz, line 195: As a string, DATASET should be part of the path.

anoto-moniz (approved):
Looks good to me! I have a number of questions or comments or suggestions, but none of them are blockers.

src/citrine/resources/project.py
```diff
@@ -228,6 +251,86 @@ def table_configs(self) -> TableConfigCollection:
...
+            Returns ``True`` if resource successfully published
+
+        """
+        if not self.session._accounts_service_v3:
+            raise NotImplementedError("Not available, you may be looking for project.make_public")
+
+        resource_access = resource.access_control_dict()
+        resource_type = resource_access["type"]
+        self.session.checked_post(
+            self._path() + f"/published-resources/{resource_type}/batch-publish",
```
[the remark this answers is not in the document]
crleblanc: I remember there was an argument against using f-strings for backwards compatibility.  These were added in Python 3.6 and we're already using them in this repo, so I think we're good to use them everywhere.

tests/resources/test_project.py
crleblanc: This also looks like a good candidate for an f-string.

src/citrine/_session.py
```diff
@@ -147,6 +147,7 @@ def checked_request(self, method: str, path: str,
 
         if self._is_access_token_expired():
             self._refresh_access_token()
+            self._check_accounts_version()
```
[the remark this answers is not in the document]
crleblanc: I wouldn't expect this to change very often, but this is controlled by a feature flag which could get toggled at any point.  It's probably good to check this on refresh to make sure we have the right setting.

src/citrine/resources/project.py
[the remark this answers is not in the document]
kroenlein: Correct; we dropped 3.5 support once long term support for 3.5 expired several months ago.  We should be planning to drop 3.6 support once it becomes obsolete at the end of this month.

kroenlein (commented):
Should we be adding deprecation statements to methods that are v1 only?

src/citrine/_session.py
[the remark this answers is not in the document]
talltea: I also have it here in case someone has an open script that they are in the middle of running (which happens sometimes because of all the long running tasks) when the switch over occurs.

talltea:
> Should we be adding deprecation statements to methods that are v1 only?

Good point, will do

src/citrine/resources/project.py
```diff
@@ -458,6 +583,8 @@ def remove_user(self, user_uid: Union[str, UUID]) -> bool:
             Returns ``True`` if user successfully removed
 
         """
+        if self.session._accounts_service_v3:
```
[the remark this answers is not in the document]
talltea: this got a little longer than expected, since in some cases I want to check that we're not in v3 instead of checking for v3, but I think it was still worth it

tests/resources/test_project.py
talltea: the share method takes `project_id: Optional[Union[str, UUID]]` but only `resource_id: Optional[str]` 🤷‍♀️ (it's an existing method, so I didn't change the signature) 
In particular, I think UUIDs aren't json serializable. 

But that should be other _project_id, instead of other_team_id

talltea:
> > Should we be adding deprecation statements to methods that are v1 only?
> 
> Good point, will do

Thinking more about this, I'm not sure deprecation warnings are appropriate here. Since there's nothing that a user can do when they are on v1 and using these methods, I think it would be confusing to get a deprecation warning.

There's a `PendingDeprecationWarning` that I think would work here though. https://docs.python.org/3/library/exceptions.html#PendingDeprecationWarning
The docs say it's rarely used, but this might be a situation where it's the correct warning. Thoughts, @kroenlein ?

kroenlein:
> > > Should we be adding deprecation statements to methods that are v1 only?
> > 
> > 
> > Good point, will do
> 
> Thinking more about this, I'm not sure deprecation warnings are appropriate here. Since there's nothing that a user can do when they are on v1 and using these methods, I think it would be confusing to get a deprecation warning.
> 
> There's a `PendingDeprecationWarning` that I think would work here though. https://docs.python.org/3/library/exceptions.html#PendingDeprecationWarning The docs say it's rarely used, but this might be a situation where it's the correct warning. Thoughts, @kroenlein ?

I think your thought process makes sense here, and maybe we should just hold off on adding the warnings until the methods are actually deprecated.  If there are v1 methods that cannot be made to be v3 compliant until the platform is v3, then that would be very frustrating for the user.  Maybe just add a deprecation warning to methods that can be migrated?

tests/resources/test_project.py
```diff
@@ -439,12 +708,17 @@ def test_delete_project(collection, session):
     assert expected_call == session.last_call
 
 
-def test_update_project(collection, project):
+def test_update_project(collection: ProjectCollection, project):
     project.name = "updated name"
     with pytest.raises(NotImplementedError):
         collection.update(project)
```
dkao-citrine, line 714: One of the requirements listed in the ticket is that the Project Collection `update` method should now be able to update both the name and description. Did I miss something here?

tests/resources/test_project.py
```diff
@@ -439,12 +708,17 @@ def test_delete_project(collection, session):
     assert expected_call == session.last_call
```
dkao-citrine, line 708: Also, according to the ticket description, in v3, users are not supposed to be able to delete projects, anymore. Am I reading that correctly?
https://citrine.atlassian.net/browse/PLA-8395

tests/resources/test_project.py
```diff
@@ -339,6 +578,27 @@ def test_list_projects(collection, session):
     assert 5 == len(projects)
 
 
+def test_list_projects_v3(collection_v3, session_v3):
```
dkao-citrine, line 581: Is the `team_id` argument supposed to be required in the v3 API?

tests/resources/test_project.py
```diff
@@ -339,6 +578,27 @@ def test_list_projects(collection, session):
...
+    # Given
+    projects_data = ProjectDataFactory.create_batch(5)
+    session_v3.set_response({'projects': projects_data})
+
+    # When
+    projects = list(collection_v3.list())
```
dkao-citrine, line 587: Is the `team_id` argument supposed to be required in the v3 API?

tests/resources/test_project.py
```diff
@@ -306,7 +468,84 @@ def test_project_registration(collection, session):
...
 
-def test_get_project(collection, session):
+def test_failed_register_v3():
+    team_id = uuid.uuid4()
+    session = mock.Mock()
+    session.post_resource.side_effect = NotFound(f'/teams/{team_id}/projects',
+                                                 FakeRequestResponse(400))
+    project_collection = ProjectCollection(session=session, team_id=team_id)
+    with pytest.raises(ModuleRegistrationFailedException) as e:
+        project_collection.register("Project")
```
dkao-citrine, line 478: Is the `team_id` argument supposed to be required in the v3 API?

tests/resources/test_project.py
talltea: Let's do this in a followup and check in with product that we actually need updates of team names/descriptions in the sdk

tests/resources/test_project.py
talltea: We actually did get that api endpoint, so the ticket is out of date and project delete should work

tests/resources/test_project.py
talltea: the team_id is set in the creation of the collection, so there's no need for the user to specify team_id in the call (the docs might be out of date, this was a change in the design mid pr)

tests/resources/test_project.py
talltea: it's part of the collection creation

crleblanc (approved):
This seems like a thoroughly reviewed PR.  I think we should merge it and if anything comes up we can address it before the next Citrine Python release.

anoto-moniz (approved):

tests/resources/test_project.py
dkao-citrine: Ah, okay. Can you update or amend the ticket, then, please?

talltea: done

tests/resources/test_project.py
dkao-citrine: Sounds great. Can you link the follow-up ticket to PLA-8395?

talltea: let's hold off until we get a request from product, I'm not convinced there's a use case for updating team names in the sdk

tests/resources/test_project.py
dkao-citrine: Gotcha, gotcha. Can you just update this Collection methods requirement in the ticket, then?

`citrine.projects.list           -  add an optional team_id kwarg that is required for v3`

tests/resources/test_project.py
dkao-citrine: Same request here.

tests/resources/test_project.py
talltea: updated

tests/resources/test_project.py
talltea: updated

tests/resources/test_project.py
dkao-citrine: I see it; thanks!

tests/resources/test_project.py
dkao-citrine: I see the update; thank you.

tests/resources/test_project.py
dkao-citrine: I see the update. 👍🏼

tests/resources/test_project.py
dkao-citrine: Okay, sounds good.

PR #702, opened 2021-12-02
[PLA-8726] Adding support for branches.

anoto-moniz:
As branches act as containers for design workflows, this also moves design workflows to use the v2 endpoints in almost all cases. The one exception is creation when the DesignWorkflowCollection doesn't have a branch ID (namely, `Project.design_workflows.register`), in which case we create the design workflow with the v1 endpoint and then retrieve the full object with the v2 endpoint.

### PR Type:


src/citrine/resources/project.py
```diff
@@ -136,7 +142,7 @@ def predictor_evaluation_executions(self) -> PredictorEvaluationExecutionCollect
     @property
     def design_workflows(self) -> DesignWorkflowCollection:
         """Return a collection representing all visible design workflows."""
-        return DesignWorkflowCollection(self.uid, self.session)
+        return DesignWorkflowCollection(project_id=self.uid, branch_id=None, session=self.session)
```
anoto-moniz, line 145: I ended up not deprecating this because `/api/v2/design-workflows` accepts an optional `branch` query parameter for filtering. Omitting it gets them all the (unarchived) design workflows in the project, just like this property currently does. So it seems reasonable to allow a user to invoke that by calling `Project.design_workflows`.

However, that also means that until the v1 design workflow endpoints are deprecated, the user cannot actually invoke that form. They either call `Project.design_workflows`, which gets all the design workflows but doesn't include their branch IDs, or they call `Branch.design_workflows`, which has their branch IDs but only gets design workflows from that branch.

That being said, I don't think it causes any practical issues, as I have a hard time imagining a real-world flow where a user is hunting for a specific design workflow just to get the branch ID.

src/citrine/resources/branch.py
```diff
@@ -0,0 +1,100 @@
...
+
+    @property
+    def design_workflows(self) -> DesignWorkflowCollection:
+        """Return a resource representing all workflows contained within this branch."""
+        if getattr(self, 'project_id', None) is None:
+            raise AttributeError('Cannot initialize workflow without project reference!')
+        return DesignWorkflowCollection(project_id=self.project_id, branch_id=self.uid, session=self.session)
+
+
+class BranchCollection(Collection[Branch]):
```
anoto-moniz, line 45: I left `delete` alone since it's currently present in the API docs, although I should note that when trying it against development I get a 500. And as archiving branches is still being discussed, I figure that should probably be a follow-on ticket.

src/citrine/_rest/resource.py
```diff
@@ -21,6 +22,7 @@ class ResourceTypeEnum(BaseEnumeration):
 
     TEAM = "TEAM"
     PROJECT = "PROJECT"
+    BRANCH = "BRANCH"
```
talltea, line 25: I think BRANCH is actually just a MODULE under the hood (like the rest of ai engine assets)

src/citrine/resources/branch.py
talltea: Go ahead and do a NotImplementedError for delete for now, we can follow up with an archiving method or a delete implementation depending on what we go with in the backend. Unless we expect delete branch to work (checking with andy)

src/citrine/_rest/resource.py
andyczerwonka: Depending what you mean by module, yes. We don't differentiate them in the accounts service. That may be a mistake though, not sure what it's used for.

src/citrine/resources/branch.py
```diff
@@ -0,0 +1,100 @@
...
+        self.project_id: UUID = project_id
+        self.session: Session = session
+
+    def build(self, data: dict):
+        branch = Branch.build(data)
+        branch.session = self.session
+        branch.project_id = self.project_id
+        return branch
+
+    def register(self, name: str) -> Branch:
```
talltea, line 63: I wonder if it would be better to not over-ride this method. our options for the interface here are:
1. `projects.branches.register(Branch("my branch"))`
2. `projects.branches.register("my branch")`. (what you have here)

The second way makes sense to me since all we need is the name, but it's not consistent with other Orion collections like predictors, workflows, etc. I can see this going either way tbh.

src/citrine/resources/branch.py
```diff
@@ -0,0 +1,100 @@
...
+        """
+        Create and upload new branch.
+
+        Parameters
+        ----------
+        name: str
+            Name of the project to be created.
```
talltea, line 70: `project` -> `branch`

src/citrine/resources/branch.py
```diff
@@ -0,0 +1,100 @@
...
+
+        """
+        return super().register(Branch(name))
+
+    def list(self, *,
```
talltea, line 75: I think somewhere in the paginator, we've deprecated the option to pass in page as an argument. If that's right, let's not add it here. (if we find we need it, we can add it back in, but no need to pass it in now)

src/citrine/resources/design_workflow.py
```diff
@@ -17,15 +17,23 @@ class DesignWorkflowCollection(Collection[DesignWorkflow]):
...
+    def register(self, model: DesignWorkflow) -> DesignWorkflow:
+        model.branch_id = self.branch_id
+        return super().register(model)
+
     def build(self, data: dict) -> DesignWorkflow:
         """Build an individual DesignExecution."""
         workflow = DesignWorkflow.build(data)
         workflow._session = self.session
         workflow.project_id = self.project_id
+        workflow.branch_id = self.branch_id
```
talltea, line 36: I think we should be able to get this for free from the response body since you've added it here: https://github.com/CitrineInformatics/citrine-python/pull/702/files#diff-85cbb32077676199db6ca0d43596047d4b3956d22f1d1a055fd28ea3872d1eb5R35

src/citrine/resources/branch.py
anoto-moniz: Confirmed with Andy that we expect to be able to delete an empty branch. I filed a ticket for us to more gracefully decline when the user requests we delete a non-empty branch: https://citrine.atlassian.net/browse/PLA-8755.

Also note, there are two issues which arise when deleting an empty branch. One is the Response object returned does not contain the status code due to how Session.delete_resource works: https://citrine.atlassian.net/browse/PLA-8764. The other is that it also doesn't contain the ID of the deleted branch due to the endpoing returning the wrong Content-Type: https://citrine.atlassian.net/browse/PLA-8765

src/citrine/_rest/resource.py
anoto-moniz: Sure.

For my own info, what is ResourceTypeEnum used for?

src/citrine/resources/branch.py
anoto-moniz: I was modelling after ProjectCollection, which does do this. However, since the majority of the others accept an object, and it means less code, I'll change it to method 1.

src/citrine/resources/branch.py
anoto-moniz: Gotcha. I did see somewhere we state it's deprecated, but I figured since it's still there we would need to keep using it until it's gone. I'll remove it.

src/citrine/resources/design_workflow.py
anoto-moniz: Ah, thank you! I'd meant to try this out, and clearly forgot. Just tried now (register, get, and list), and all populated it properly.

src/citrine/resources/project.py
anoto-moniz: Conversing with James and Andy, the solution we settled on is to just always hit v2. What this means is that to create a design workflow without a branch ID, we'll hit the v1 API, then retrieve the branch ID with a get.

andyczerwonka (commented):
Do we really want to enable this in the SDK? The workflow is different....

anoto-moniz:
> Do we really want to enable option predictors and design spaces in the SDK? The workflow is different....

Personally, my default philosophy is that if you can do it in the UI, you should be able to do it in the SDK. If that's not our approach to citrine-python, I can of course revert those optional fields from DesignWorkflow.

andyczerwonka:
> > Do we really want to enable option predictors and design spaces in the SDK? The workflow is different....
> 
> Personally, my default philosophy is that if you can do it in the UI, you should be able to do it in the SDK. If that's not our approach to citrine-python, I can of course revert those optional fields from DesignWorkflow.

I would revert. I can't see a use case for scripting a partial design workflow. I can only see it creating SPT tickets, e.g. _why isn't my workflow validating? @jspeerless Can you think of any use cases here?

anoto-moniz:
> > > Do we really want to enable option predictors and design spaces in the SDK? The workflow is different....
> > 
> > 
> > Personally, my default philosophy is that if you can do it in the UI, you should be able to do it in the SDK. If that's not our approach to citrine-python, I can of course revert those optional fields from DesignWorkflow.
> 
> I would revert. I can't see a use case for scripting a partial design workflow. I can only see it creating SPT tickets, e.g. _why isn't my workflow validating? @jspeerless Can you think of any use cases here?

I also looked back at the ticket, and that isn't needed to support design workflows. I do need to make sure I can deserialize responses in which they're null, so I'm going to add that back in along with some tests shortly.

andyczerwonka:
> I do need to make sure I can deserialize responses in which they're null

Yep, that makes sense. You're right, you have to support it on `read`, so I think your change makes sense. No point in having different semantics and `read` vs `write`. 👍🏻

anoto-moniz:
> > I do need to make sure I can deserialize responses in which they're null
> 
> Yep, that makes sense. You're right, you have to support it on `read`, so I think your change makes sense. No point in having different semantics and `read` vs `write`. 👍🏻

Sorry, when you say "your change makes sense", are you referring to my original change (optional in the DesignWorkflow constructor), or the current version (only optional for serialization/deserialization)?

andyczerwonka:
> Sorry, when you say "your change makes sense", are you referring to my original change (optional in the DesignWorkflow constructor), or the current version (only optional for serialization/deserialization)?

I didn't go through your changes in detail, but I assume your original change. If a use can get a handle to a partial design workflow, then they need the tools to edit them. It feels like you were on the right path when I first asked about the use case.

anoto-moniz:
> > Sorry, when you say "your change makes sense", are you referring to my original change (optional in the DesignWorkflow constructor), or the current version (only optional for serialization/deserialization)?
> 
> I didn't go through your changes in detail, but I assume your original change. If a use can get a handle to a partial design workflow, then they need the tools to edit them. It feels like you were on the right path when I first asked about the use case.

Got it. Just updated the PR with that original change, which allows users to create a partial design workflow via the SDK (along with retrieving them, of course).

src/citrine/resources/design_workflow.py
```diff
@@ -16,13 +17,65 @@ class DesignWorkflowCollection(Collection[DesignWorkflow]):
...
+        model: DesignWorkflow
+            The design workflow to be uploaded.
+
+        Return
+        -------
+        DesignWorkflow
+            The newly created design workflow.
+
+        """
+        if self.branch_id is None:
```
talltea, line 46: in this case, might it be worthwhile to check that predictor and design space aren't None?

src/citrine/resources/design_workflow.py
```diff
@@ -16,13 +17,65 @@ class DesignWorkflowCollection(Collection[DesignWorkflow]):
...
+            created_dw = super(DesignWorkflowCollection, v1).register(model)
+            return super().get(created_dw.uid)
+        else:
+            # branch_id is in the body of design workflow endpoints, so it must be serialized.
+            # This means the collection branch_id might not match the workflow branch_id. The
+            # collection should win out, since the user is explicitly referencing the branch
+            # represented by this collection.
+            # To avoid modifying the parameter, and to ensure the only change is the branch_id, we
+            # deepcopy, modify, then register it.
+            model_copy = deepcopy(model)
```
talltea, line 60: This is an interesting case. Would this ever be intentional? I think this should raise a warning, and possibly even an error.

src/citrine/resources/project.py
```diff
@@ -136,7 +142,7 @@ def predictor_evaluation_executions(self) -> PredictorEvaluationExecutionCollect
     @property
```
talltea, line 142: Did we settle on this not needed a deprecation? (sorry if I missed the discussion)

src/citrine/informatics/workflows/design_workflow.py
```diff
@@ -17,21 +17,23 @@ class DesignWorkflow(Resource['DesignWorkflow'], Workflow, AIResourceMetadata):
...
         a description of the workflow
 
     """
 
-    design_space_id = properties.UUID('design_space_id')
+    design_space_id = properties.Optional(properties.UUID, 'design_space_id')
     processor_id = properties.Optional(properties.UUID, 'processor_id')
-    predictor_id = properties.UUID('predictor_id')
+    predictor_id = properties.Optional(properties.UUID, 'predictor_id')
+    branch_id: Optional[UUID] = properties.Optional(properties.UUID, 'branch_id')
```
talltea, line 35: partial design workflows also contain `data_source_id`.  We'll probably want to expose this here too

andyczerwonka: That I would not expose, it’s going to confuse everyone and it’s not used downstream at all. @jspeerless do you agree?

talltea:
Let's add this to the docs too

src/citrine/resources/design_workflow.py
anoto-moniz: I chatted with James about this, and the conclusion was that this allows a user to create the same design workflow on different branches more conveniently. I don't know how common this will be of course, but it seems reasonable to support.

src/citrine/resources/design_workflow.py
anoto-moniz: I was on the fence, but I think I agree with you.

src/citrine/resources/project.py
anoto-moniz: Correct. In the design workflows v2, branch is optional when listing. Without branch, it will list all design workflows in the project, just like today.

https://docs.corp.citrine.io/orion2/#operation/getApiV2Design-workflows

anoto-moniz:
@talltea I'd like your opinion on something that just occurred to me.

Design workflows support updating. In v2, the body of the PUT includes its branch ID. Workflows aren't identified by their branch, so this just allows moving a workflow to another branch. Which means the following code will work.

```
dw = branch1.design_workflows.register(DesignWorkflow(name="example dw"))
dw.branch_id = branch3.uid
branch2.design_workflows.update(dw)
```

The result is `dw` will be on `branch3`, and `branch2` was not involved at all. Which feels awkward to me, and like it could cause confusion. Of course, they can use `project.design_workflows.update(dw)` instead, which I find much clearer. So the question is if it's worth disallowing this for clarity by raising an error and directing them to ``project.design_workflows.update`. My instinct is yes, we should raise an error, but I want a second opinion.

anoto-moniz:
Just rebased on top of recent changes, and bumped the version to 1.19.0.

talltea:
This is looking great. I kicked off a devkit-custom job here just to confirm that everything will work smoothly when we merge (I'm sure it'll be fine, but it's better safe than in a dev0) https://jenkins.corp.citrine.io/job/Devkit-custom/583/

I'm wondering if we foresee python sdk users creating partial design workflows, and if so, if we could add that to the docs?

anoto-moniz:
> I'm wondering if we foresee python sdk users creating partial design workflows, and if so, if we could add that to the docs?

I can't say if they will, of course. But I decided that since the current docs don't say all three design workflow elements are required, I could get away without explicitly mentioning they're not. Which also keeps things a bit simpler, since I don't have to mention that in order to create a DW without them, you have to be on a branch. Happy to change that if we think it's worthwhile, though!

talltea:
> I can't say if they will, of course. But I decided that since the current docs don't say all three design workflow elements are required, I could get away without explicitly mentioning they're not. Which also keeps things a bit simpler, since I don't have to mention that in order to create a DW without them, you have to be on a branch. Happy to change that if we think it's worthwhile, though!

Naw, I think we're good for now 👍

talltea (approved):

dkao-citrine (approved):
I agree with @talltea. What an entrance! 👏🏼 I especially appreciate all of the code comments and detailed docstrings, as they made the review more accessible and will make future modifications easier.

PR #706, opened 2021-12-15
Deprecate positional arguments for Citrine and Session

kroenlein:
# Citrine Python PR

## Description 
We've gotten signal that customers find the fact that the second positional argument to Citrine is `scheme` is confusing.  Having four valid positional arguments is also against [our coding-style](https://github.com/CitrineInformatics/citrine-python/blob/main/CONTRIBUTING.md#coding-style).  This deprecates having more than one argument (still allowing people to pass the API key).

Somehow, in the existing testing, the full signature was never actually exercised.  I've also added tests to address this.

### PR Type:


tests/test_citrine.py
```diff
@@ -22,14 +21,37 @@ def refresh_token(expiration: datetime = None) -> dict:
...
 def test_citrine_creation():
-
     with requests_mock.Mocker() as m:
         m.post('https://citrine-testing.fake/api/v1/tokens/refresh', json=token_refresh_response)
         m.get('https://citrine-testing.fake/api/v1/utils/runtime-config', json=dict())
 
         assert '1234' == Citrine(api_key='1234', host='citrine-testing.fake').session.refresh_token
 
 
+def test_citrine_signature():
```
anoto-moniz: This should probably include a test that we can provide `api_key` as a positional argument without raising a `DeprecationWarning`.

tests/test_session.py
```diff
@@ -49,6 +49,32 @@ def session():
     return session
 
 
+def test_session_signature():
```
anoto-moniz: Similar to above, but `refresh_token`.

src/citrine/_session.py
```diff
@@ -32,11 +33,31 @@ class Session(requests.Session):
     """Wrapper around requests.Session that is both refresh-token and schema aware."""
 
     def __init__(self,
-                 refresh_token: str = environ.get('CITRINE_API_KEY'),
```
kroenlein, line 35: When `environ.get` is in the method signature, the value is set at build time, which means we can't mock the value.

tests/test_citrine.py
```diff
@@ -22,14 +21,57 @@ def refresh_token(expiration: datetime = None) -> dict:
...
 def test_citrine_creation():
-
     with requests_mock.Mocker() as m:
         m.post('https://citrine-testing.fake/api/v1/tokens/refresh', json=token_refresh_response)
         m.get('https://citrine-testing.fake/api/v1/utils/runtime-config', json=dict())
 
         assert '1234' == Citrine(api_key='1234', host='citrine-testing.fake').session.refresh_token
 
 
+def test_citrine_signature(monkeypatch):
```
anoto-moniz, line 31: We should test that passing `api_key` as the only positional argument works.

tests/test_session.py
```diff
@@ -49,6 +48,52 @@ def session():
     return session
 
 
+def test_session_signature(monkeypatch):
```
anoto-moniz, line 51: We should test that passing `refresh_token` as the only positional argument works.

tests/test_session.py
```diff
@@ -290,30 +335,3 @@ def test_patch(session: Session):
         m.patch('http://citrine-testing.fake/api/v1/bar/something', status_code=200, json=json_to_validate)
         response_json = session.patch_resource('bar/something', {"ignored": "true"})
         assert response_json == json_to_validate
-
-
-@mock.patch.object(Session, '_check_accounts_version')
-@mock.patch.object(Session, '_refresh_access_token')
-def test_base_url_assembly(*_):
```
anoto-moniz, line 297: I'm not clear why we're removing this test.

src/citrine/citrine.py
```diff
@@ -26,10 +26,24 @@ class Citrine:
...
-                 scheme: str = 'https',
+                 legacy_scheme: Optional[str] = None,
                  host: str = environ.get('CITRINE_API_HOST'),
-                 port: Optional[str] = None):
-        self.session: Session = Session(api_key, scheme, host, port)
+                 port: Optional[str] = None,
+                 *,
+                 scheme: str = 'https'):
+        if legacy_scheme is not None:
+            warn("Creating a session with positional arguments other than refresh_token "
```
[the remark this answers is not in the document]
kroenlein: Yes.  Copypasta following from Session changing the name of the argument.

tests/test_session.py
kroenlein: It's redundant with the tests above, and this test is written to do clever things depending on implementation.  For example, if you actually were to call `Citrine(host=None)`, you get an `InvalidURL` exception fired off from deeper in the stack, so under no circumstances would we care how it builds a URL with a null host.  But this test mocks the deeper calls, so this error is bypassed and a fatal scenario is treated as reasonable.

kroenlein:
Raising an exception for a missing API key was breaking some tests.  Upon deeper digging, it made sense to remove that requirement because running with local devkit uses no API key and adding an extra conditional checking for `host=="localhost"` seemed potentially complicated.

tests/test_citrine.py
```diff
@@ -22,14 +21,53 @@ def refresh_token(expiration: datetime = None) -> dict:
...
+        patched_key = "5678"
+        patched_host = "monkeypatch.citrine-testing.fake"
+        monkeypatch.setenv("CITRINE_API_KEY", patched_key)
+        monkeypatch.setenv("CITRINE_API_HOST", patched_host)
+        m.post(f'https://{patched_host}/api/v1/tokens/refresh', json=token_refresh_response)
+        m.get(f'https://{patched_host}/api/v1/utils/runtime-config', json=dict())
+
+        assert patched_key == Citrine().session.refresh_token
+        assert patched_key == Citrine(patched_key).session.refresh_token
+
```
anoto-moniz, line 52: I think we should add a confirmation that ti works when the API key is omitted and CITRINE_API_KEY is not defined.

tests/test_session.py
```diff
@@ -49,6 +48,48 @@ def session():
...
+        patched_key = "5678"
+        patched_host = "monkeypatch.citrine-testing.fake"
+        monkeypatch.setenv("CITRINE_API_KEY", patched_key)
+        monkeypatch.setenv("CITRINE_API_HOST", patched_host)
+        m.post(f'https://{patched_host}/api/v1/tokens/refresh', json=token_refresh_response)
+        m.get(f'https://{patched_host}/api/v1/utils/runtime-config', json=dict())
+
+        assert patched_key == Session().refresh_token
+        assert patched_key == Session(patched_key).refresh_token
+
```
anoto-moniz, line 73: Same as above.

anoto-moniz (approved):
There's one small check I think we should add in the tests, but other than that looks great!

anoto-moniz (approved):
LGTM!

PR #723, opened 2022-01-31
PLA-9062- Update MPP to work with categoricals

sfriedowitz:
# Citrine Python PR

## Description 
Mithril and the backend now supports categorical property values in the `MeanPropertyPredictor`. Let's update the SDK to utilize this feature, by updating the docs and serialization signature.

I'll update the main docs in a separate PR.

### PR Type:


src/citrine/informatics/predictors/mean_property_predictor.py
```diff
@@ -36,36 +44,42 @@ class MeanPropertyPredictor(
...
     """
 
     _resource_type = ResourceTypeEnum.MODULE
 
     input_descriptor = _properties.Object(FormulationDescriptor, 'config.input')
-    properties = _properties.List(_properties.Object(RealDescriptor), 'config.properties')
-    p = _properties.Integer('config.p')
-    training_data = _properties.List(_properties.Object(DataSource),
-                                     'config.training_data', default=[])
+    properties = _properties.List(_properties.Object(Descriptor), 'config.properties')
```
sfriedowitz, line 69: I tried to make this 
```
properties = _properties.List(
_properties.Union([_properties.Object(RealDescriptor), _properties.Object(CategoricalDescriptor)]),
'config.properties'
)
```
but it was failing deserialization of platform objects. 

Do you know how to make that work instead of just the polymorphic object? I think we'd rather be explicit.

src/citrine/informatics/predictors/mean_property_predictor.py
```diff
@@ -1,18 +1,26 @@
...
 
 
-class MeanPropertyPredictor(
-        Resource['MeanPropertyPredictor'], Predictor, AIResourceMetadata):
-    """A predictor interface that computes mean component properties.
+class MeanPropertyPredictor(Resource['MeanPropertyPredictor'], Predictor, AIResourceMetadata):
+    """A predictor that computes a component-weighted mean of real or categorical properties.
+
+    Each component in a formulation contributes to the mean property
+    a weight equal to its quantity raised to the `p` power.
```
kroenlein, line 19:
```suggestion
    a weight equal to its quantity raised to the power `p`.
```
I think this is the more common turn of phrase.

src/citrine/informatics/predictors/mean_property_predictor.py
```diff
@@ -22,11 +30,11 @@ class MeanPropertyPredictor(
         description of the predictor
     input_descriptor: FormulationDescriptor
         descriptor that represents the input formulation
-    properties: List[RealDescriptor]
-        List of descriptors to featurize
-    p: int
-        Power of the `generalized mean <https://en.wikipedia.org/wiki/Generalized_mean>`_.
-        Only integer powers are supported.
+    properties: List[Union[RealDescriptor, CategoricalDescriptor]]
+        list of real or categorical descriptors to featurize
```
kroenlein, line 34:
```suggestion
        List of real or categorical descriptors to featurize
```
Match capitalization pattern

src/citrine/informatics/predictors/mean_property_predictor.py
kroenlein: https://github.com/CitrineInformatics/citrine-python/blob/177795d6408d9667d053a25e2f26a3b4a079bbcd/src/citrine/informatics/predictors/graph_predictor.py#L37-L38
https://github.com/CitrineInformatics/citrine-python/blob/177795d6408d9667d053a25e2f26a3b4a079bbcd/src/citrine/informatics/design_spaces/product_design_space.py#L35-L37

Looks consistent with other usages.  Might the default be necessary for something about the endpoint?

sfriedowitz: I think I solved it in this next commit. There was an issue in the serialization pckage where a `RuntimeError` was being thrown instead of a `ValueError`, which caused the `Union` type to fail during deserialization.

kroenlein (approved):
Minor, optional changes to documentation.

src/citrine/informatics/predictors/mean_property_predictor.py
bfolie: I would expect what you tried to work. Were you able to track down the root cause of the deserialization failure?

sfriedowitz: The error is occurring because `properties.Union` attempts to deserialize w/ the element types in order in a try/catch block. When those element types are `properties.Object`, we end up calling `Property.deserialize_from_dict` which raises a `RuntimeError` on line 93.

This gets a bit iffy. That _should_ probably be a `RuntimeError`, so catching both `ValueError` and `RuntimeError` inside `properties.Union` may be the best approach.

sfriedowitz: ehh maybe line 93 could logically also be a `ValueError`. Switching that to `ValueError` breaks a few other tests, though, which I'll have to trouble shoot.

sfriedowitz:
I'll run a devkit test against this branch b4 merging.

src/citrine/_serialization/properties.py
```diff
@@ -498,10 +498,13 @@ def _deserialize(self, value: typing.Any) -> typing.Any:
...
             except ValueError:
                 pass
+            except RuntimeError:
+                # passes on RuntimeError from Property.deserialize_from_dict
+                pass
         # this is a failsafe that shouldn't actually ever get hit, hence no cover
         msg = "An unexpected error occurred while trying to deserialize {} to one of the " \
               "following types: {}.".format(value, self.underlying_types)  # pragma: no cover
-        raise ValueError(msg)  # pragma: no cover
+        raise RuntimeError(msg)  # pragma: no cover
```
bfolie, line 507: Why change the error type?

bfolie:
You'll also want to update the docs

src/citrine/_serialization/properties.py
sfriedowitz: I think it was incorrect. RuntimeError is consistent with the serialize method, and other similar errors thrown in this file.

sfriedowitz:
> You'll also want to update the docs

I plan to do so in another PR, tomorrow.

docs/source/workflows/predictors.rst
```diff
@@ -494,14 +494,18 @@ where :math:`d` is density and :math:`f` is relative ingredient fraction.
 If the densities of water and salt are known, we can compute the expected density of a candidate mixture using this predictor.
 
 The :class:`~citrine.informatics.predictors.mean_property_predictor.MeanPropertyPredictor` computes mean properties of formulation ingredients.
+For real-valued properties, a quantity-weighted mean value is computed.
+For categorical-valued properties, a quantity-weighted distribution of property values is accumulated.
```
kroenlein, line 498: Strange that one is computed and another is `accumulated`.

docs/source/workflows/predictors.rst
```diff
@@ -512,30 +516,40 @@ Missing ingredient properties can be handled one of three ways:
...
-+---------------+----------------+
++---------------+----------------+--------------------+
+| Ingredient id | Density (g/cc) | Acetone Solubility |
++===============+================+====================+
+| water         | 1.0            | Yes                |
++---------------+----------------+--------------------+
+| salt          | 2.16           | No                 |
++---------------+----------------+--------------------+
+| boric acid    | N/A            | N/A                |
++---------------+----------------+--------------------+
```
kroenlein, line 542: Might use `boric acid` as `slightly soluble` versus water `soluble` and salt `insoluble`.

src/citrine/_serialization/properties.py
```diff
@@ -498,10 +498,13 @@ def _deserialize(self, value: typing.Any) -> typing.Any:
                 return prop.deserialize(value)
             except ValueError:
                 pass
+            except RuntimeError:
+                # passes on RuntimeError from Property.deserialize_from_dict
+                pass
```
kroenlein, line 503: I realize you are responding to existing code, but the error in `Property.deserialize_from_dict` should really be a ValueError.
[ValueError](https://docs.python.org/3.10/library/exceptions.html?highlight=runtimeerror#ValueError)
https://github.com/CitrineInformatics/citrine-python/blob/177795d6408d9667d053a25e2f26a3b4a079bbcd/src/citrine/_serialization/properties.py#L83-L103

src/citrine/_serialization/properties.py
kroenlein, line 507: Except you should be able to easily test this by handing it a malformed value.

kroenlein (commented):
The change from properties to compositions being raised to a power, while I believe to be a much more motivated scenario physically, is technically a breaking change.  We should either support both or go through a complex deprecation cycle.

src/citrine/_serialization/properties.py
sfriedowitz: I'll have to fix a handful of tests that rely on that being a `RuntimeError`, but I can address it.

src/citrine/_serialization/properties.py
sfriedowitz: I cannot explain the thoughts of the witty lad who built our custom serde library.

src/citrine/_rest/collection.py
```diff
@@ -134,13 +134,7 @@ def _build_collection_elements(self,
...
         """
         for element in collection:
-            try:
-                yield self.build(element)
-            except(KeyError, ValueError) as e:
-                # TODO:  This is a patch to handle deprecated predictors client side.
-                # Remove when predictors are migrated
-                warnings.warn(f"Building element skipped due to error: {e}")  # pragma: no cover
-                pass  # pragma: no cover
+            yield self.build(element)
```
sfriedowitz, line 137: This allows that test that expects a `ValueError` to property function. We were previously catching the `ValueError` based on this `TODO` message fallback. 

I am not sure if it is safe to remove this, but just posting a version with it removed and `Property.deserailize_from_dict` throwing a `ValueError` instead of a `RuntimeError`.

src/citrine/_serialization/properties.py
```diff
@@ -498,10 +498,8 @@ def _deserialize(self, value: typing.Any) -> typing.Any:
                 return prop.deserialize(value)
             except ValueError:
                 pass
-        # this is a failsafe that shouldn't actually ever get hit, hence no cover
-        msg = "An unexpected error occurred while trying to deserialize {} to one of the " \
-              "following types: {}.".format(value, self.underlying_types)  # pragma: no cover
-        raise ValueError(msg)  # pragma: no cover
+        raise RuntimeError("An unexpected error occurred while trying to deserialize {} to one of the "
```
sfriedowitz, line 501: This was already covered.

tests/resources/test_project.py
```diff
@@ -614,7 +614,7 @@ def test_list_projects_filters_non_projects(collection, session):
     session.set_response({'projects': projects_data})
 
     # Then
-    with pytest.raises(RuntimeError):
+    with pytest.raises(ValueError):
```
sfriedowitz, line 617: And this change is based on the type of error thrown by `Property.deserialize_from_dict`.

bfolie:
> The change from properties to compositions being raised to a power, while I believe to be a much more motivated scenario physically, is technically a breaking change. We should either support both or go through a complex deprecation cycle.

I do not think that it counts as a breaking change. All existing scripts continue to work and predictor configurations can remain unchanged. Predictors will automatically be re-trained without users having to do anything.

We are changing internal modeling behavior. Admittedly it's a substantial change, but just as we don't go through a deprecation cycle for a change to candidate scoring logic or optimizer logic or for a bugfix (even a bugfix that will change prediction behavior) we don't need a deprecation cycle for this.

src/citrine/_rest/collection.py
sfriedowitz: As it turns out, we cannot remove this yet. It is related to #641 , and the new `/predictor` routes are not fully intregrated to the backend yet.

Because of that, this needs to stay with the try/catch. The `properties.Union` needs to throw a value error that is caught here, and `Property.deserialize_from_dict` needs to throw  a `RuntimeError` to bypass this point.

Ugly, but that's what we do for now. I'll ticket it so we remember to clean this up at some point.

sfriedowitz:
@bfolie Okay so here is this ugly chain that was going on here. #641 introduced that try/catch patch for good reason, to avoid errors when listing collections with deprecated predictors. We cannot remove it yet, as the backend has not migrated to the new `/predictors` route.

This caused a chain of misery in the deserialization methods. The `try/catch` was catching `ValueError` and `KeyError`, which is why `properties.Union` was updated to throw a `ValueError` and `Property.deserialize_from_dict` was throwing a `RuntimeError`. They should be flipped, logically.

We need to be able to catch errors inside `properties.Union` from `Property.deserialize_from_dict` to allow a `Union(properties.Object, properties.Object)` deserialization to work, which is what we need for the new MPP that handles real/categorical descriptors. To do this, I just added a new catch on `RuntimeError` inside `Union.deserialize`.

I ticketed this and updated all the TODOs to document the chain that should eventually be fixed when we can remove that `try/catch`.

tests/_serialization/test_simple_properties.py
```diff
@@ -209,7 +209,7 @@ def test_union_runtime_errors():
         Union([Enumeration(BaseEnumeration)]).serialize(EnumerationExample.FOO)
     # The serialized type is correct (dict) but it is missing fields
     incomplete_dataset_dict = {'name': 'name'}
-    with pytest.raises(RuntimeError):
+    with pytest.raises(ValueError):
```
sfriedowitz, line 212: This was no-covered in the code before, which I think is why it passed. I removed the no-cover and just updated the test.

docs/source/workflows/predictors.rst
```diff
@@ -512,30 +516,40 @@ Missing ingredient properties can be handled one of three ways:
 1. If ``impute_properties == False``, all ingredients must define a value for all featurized properties.
    Otherwise, the row will not be featurized.
    Use this option if you expect ingredient properties to be dense (always present) and would like to exclude rows when properties are missing.
-2. If ``impute_properties == True`` and no ``default_properties`` are specified, missing properties will be filled in using the average value across the entire dataset.
-   The average is computed from any row with data corresponding to the missing property, regardless of label or material type (i.e., the average is computed from all leaf ingredients and mixtures).
-3. If ``impute_properties == True`` and ``default_properties`` are specified, the specified property value will be used when an ingredient property is missing (instead of the average over the dataset).
+2. If ``impute_properties == True`` and no ``default_properties`` are specified,
+   missing properties will be filled in using the average value across the entire dataset.
+   For real-valued properties, this average is the mean over the training data,
```
bfolie, line 521:
```suggestion
   For real-valued properties this average is the mean over the training data,
```

docs/source/workflows/predictors.rst
```diff
@@ -512,30 +516,40 @@ Missing ingredient properties can be handled one of three ways:
...
+   while for categorical-valued properties, it is the distribution of property values in the dataset.
```
bfolie, line 522:
```suggestion
   while for categorical-valued properties it is the distribution of property values in the dataset.
```

docs/source/workflows/predictors.rst
```diff
@@ -512,30 +516,40 @@ Missing ingredient properties can be handled one of three ways:
...
+| boric acid    | N/A            | N/A                |
++---------------+----------------+--------------------+
 
 If ``impute_properties == False``, any mixture that includes boric acid will not be featurized.
-If ``impute_properties == True`` and no ``default_properties`` are specified, an density of :math:`\left( 1.0 + 2.16 \right) / 2 = 1.58` will be used.
-If a value other than 1.58 should be used (e.g., 2.0), this can be specified by setting ``default_properties = {'density': 2.0}``.
+If ``impute_properties == True`` and no ``default_properties`` are specified,
+a mean density of :math:`\left( 1.0 + 2.16 \right) / 2 = 1.58`
+and a distribution of acetone solubility with weights ``{'Soluble': 0.5, 'Insoluble': 0.5}`` will be used.
+If something other than the imputed values should be used (e.g. 2.0 or 'Slightly Soluble'),
```
bfolie, line 548:
```suggestion
If something other than the imputed values should be used (e.g., 2.0 or 'Slightly Soluble'),
```

src/citrine/_rest/collection.py
```diff
@@ -137,8 +137,8 @@ def _build_collection_elements(self,
             try:
                 yield self.build(element)
             except(KeyError, ValueError) as e:
-                # TODO:  This is a patch to handle deprecated predictors client side.
-                # Remove when predictors are migrated
+                # TODO(PLA-9109): This is a patch to handle deprecated predictors client side.
```
bfolie, line 140: Good call on matching the TODO with a ticket

sfriedowitz:
This new job is running against this branch: https://jenkins.corp.citrine.io/job/nextgen-ITs/24319/

kroenlein:
I'll reiterate that I think we should change those `RuntimeError`s to `ValueError`s, but the PR as assembled is good work.

sfriedowitz:
> I'll reiterate that I think we should change those `RuntimeError`s to `ValueError`s, but the PR as assembled is good work.

We could swap those, and then catch RuntimeError in the collection.py method maybe?

bfolie (approved):
LGTM, but you should wait until we discuss in huddle today before merging

PR #728, opened 2022-02-18
[PLA-9127] AutoMLPredictor should support all outputs.

anoto-moniz:
# Citrine Python PR

## Description 
The AutoML predictor payload contains a list of outputs, all of which
should be available to the user. This involves deprecating the output
field, which only contains the first element of the list, in favor of a
new field containing the whole list.

Also removed dead code dealing with the responses field of the payload.

### PR Type:


sfriedowitz:
@bfolie may want to weigh in here, since this will be officially opening up the new Mithril capabilities to users.

src/citrine/informatics/predictors/auto_ml_predictor.py
```diff
@@ -51,28 +54,52 @@ def __init__(self,
...
+
+    @output.setter
+    def output(self, value):
+        """[DEPRECATED] Set the output descriptor."""
+        msg = ('The "output" field is deprecated as of 1.24.0 and will be removed in 2.0.0. '
+               'Please use the "outputs" field instead.')
+        warnings.warn(msg, category=DeprecationWarning)
+        if not isinstance(value, Descriptor):
+            # Mirroring the error emited by serialization.
+            raise ValueError(f"{value} is not one of valid types: <class 'dict'> for output")
```
bfolie, line 100: Shouldn't this be `Descriptor` instead of `dict`? Both because the serialization error prints the underlying type, and also because it makes more sense given what the user is trying to do here.

bfolie:
I didn't realize that this was being exposed already, but am happy to see it.

The one thing I want to highlight is that by allowing citrine-python users to create Auto ML predictors with multiple outputs, we will soon see these payloads being delivered to the UI. And so the UI, which I believe is currently set up to only display a single output, needs to be updated as well.

bfolie:
Weighing in some more, now that I've read ticket PLA-9127 and understand the bug that led to this PR.

Up until very recently a user _could not_ create an AutoMLPredictor with more than 1 output, whether through the python client or through the UI. The backend representation allowed for multiple outputs, but that was done to anticipate future changes.

We have recently made the backend changes that allow for multiple outputs. Updating the python client to allow for multiple outputs effectively releases that feature to users, which might be an unanticipated side-effect of this bug fix ticket. I'm all for releasing it, but it should be done with the knowledge of the Product team, especially since there are some corresponding changes that will need to happen on the front end.

anoto-moniz:
> Line 129: should this be "expression_predictor.uid"?

No clue. That's not from my change, so I'll default to leaving it as is.

anoto-moniz:
> Line 89 has a typo. It should be "bulk_modulus".

Sure, if I need to make other changes I'll dip back in to fix this.

anoto-moniz:
> Weighing in some more, now that I've read ticket PLA-9127 and understand the bug that led to this PR.
> 
> Up until very recently a user _could not_ create an AutoMLPredictor with more than 1 output, whether through the python client or through the UI. The backend representation allowed for multiple outputs, but that was done to anticipate future changes.
> 
> We have recently made the backend changes that allow for multiple outputs. Updating the python client to allow for multiple outputs effectively releases that feature to users, which might be an unanticipated side-effect of this bug fix PR. I'm all for releasing it, but it should be done with the knowledge of the Product team, especially since there are some corresponding changes that will need to happen on the front end.

Got it. It wasn't super clear that this was a new capability, as it looked like a bug, so thanks for the info. I think product is aware, as I've been discussing this in standup with James, but I'll talk to him directly today to make sure there's clear communication about it.

In any case, as the ticket indicates, this is currently impacting the customer team's ability to interact with certain predictors which were somehow created with zero outputs. So if product doesn't want to release this just yet, I'll update this to cap the outputs at one.

src/citrine/informatics/predictors/auto_ml_predictor.py
anoto-moniz: That was also my assumption. However, this is the error message I get from the existing serialization mechanism, so I wanted to keep it constant. Maybe it indicates an erroneous error message in our serialization code?

bfolie: I'd say it's... not a bug, but a misfire in the setter code.

The [setter code](https://github.com/CitrineInformatics/citrine-python/blob/main/src/citrine/_serialization/properties.py#L145-L156) first checks if the value is of the correct underlying type (in this case, a `Descriptor`). If not, then it assumes the user has passed in the serialized version of the type (which is a `dict`) and it tries to deserialize it. That's why we get an error message saying "I expected a `dict`".

I think that a better flow would be
```
If value is an underlying type:
  set value
else if value is a serialized type:
  deserialize and set value
else:
  throw a ValueError that lists both the underlying and serialized types as possibilities
```

Regardless of whether or not you feel like changing that, we don't need to be held hostage by previous oversights. Might as well make _this_ message more clear.

anoto-moniz:
> > Weighing in some more, now that I've read ticket PLA-9127 and understand the bug that led to this PR.
> > Up until very recently a user _could not_ create an AutoMLPredictor with more than 1 output, whether through the python client or through the UI. The backend representation allowed for multiple outputs, but that was done to anticipate future changes.
> > We have recently made the backend changes that allow for multiple outputs. Updating the python client to allow for multiple outputs effectively releases that feature to users, which might be an unanticipated side-effect of this bug fix PR. I'm all for releasing it, but it should be done with the knowledge of the Product team, especially since there are some corresponding changes that will need to happen on the front end.
> 
> Got it. It wasn't super clear that this was a new capability, as it looked like a bug, so thanks for the info. I think product is aware, as I've been discussing this in standup with James, but I'll talk to him directly today to make sure there's clear communication about it.
> 
> In any case, as the ticket indicates, this is currently impacting the customer team's ability to interact with certain predictors which were somehow created with zero outputs. So if product doesn't want to release this just yet, I'll update this to cap the outputs at one.

After following up with James, we're going to have the SDK cap at one output for now. I'll make that tweak shortly.

bfolie:
> After following up with James, we're going to have the SDK cap at one output for now. I'll make that tweak shortly.

I also just remembered that there's a [backend validation check](https://github.com/CitrineInformatics/platform-backend/blob/master/code/orion/ml-config-core/src/main/scala/io/citrine/orion/core/domain/ModuleValidator.scala#L126-L128) that enforces at most 1 output, which is just another reason to keep the SDK capped at one output for now.

PR #734, opened 2022-03-07
Feature/pla 9213 prefer valid override option

crleblanc:
# Citrine Python PR

## Description 

As per PLA-9213 this change adds the prefer_valid optional keyword argument to PredictorCollection auto_configure method.  This option has also been added to code that calls this method.  This arg will be passed in the post payload to the `default-predictor` endpoint.

When calling this endpoint without supplying prefer_valid, we use a feature flag to get the fallback value of this option.  We plan to remove this feature flag and default to True.  Therefore it makes sense to use prefer_valid=True as a default for this param.

### PR Type:


src/citrine/builders/auto_configure.py
```diff
@@ -399,6 +400,11 @@ def from_material(
         print_status_info: bool
             Whether to print the status info during validation of assets.
             Default: False
+        prefer_valid: Boolean
+            If True, this sets the filterSparse option to True and sets targetMaxModels and
```
sfriedowitz, line 404: I think this comment needs to be adjusted. The `filterSparse` and `targetMaxModels` parameters are never user exposed, and a user has no knowledge of "predictor build settings".

Perhaps something along the lines of:
`"If True, enables filtering of sparse descriptors and trimming of excess graph components in attempt to return a default configuration that will pass validation."`

src/citrine/builders/auto_configure.py
```diff
@@ -458,6 +466,11 @@ def from_table(
         print_status_info: bool
             Whether to print the status info during validation of assets.
             Default: False
+        prefer_valid: Boolean
```
sfriedowitz, line 469: Same comment as above.

src/citrine/resources/predictor.py
```diff
@@ -119,6 +119,11 @@ def auto_configure(self, *, training_data: DataSource,
             data or not.
             If it does, then a formulation predictor is created.
             If not, then a plain predictor is created.
+        prefer_valid: Boolean
```
sfriedowitz, line 122: Same comment as above.

sfriedowitz (commented):
This looks good, just the one docs comment to address.

One thing to note. Setting `prefer_valid=True` by default may break one (and maybe others?) devkit IT, [found here](https://github.com/CitrineInformatics/nextgen-devkit/blob/master/end-to-end-tests/tests/orion/test_auto_configure_default.py). It may be wise to test against this branch before merging. 

It may also be wise for us to encode the expected behavior of `prefer_valid=True/False` in those tests, but that may be a job for me another day.

src/citrine/builders/auto_configure.py
crleblanc: That's a good point, I copied the prefer_valid feature flag description.  I'll use your suggestion.

crleblanc:
> This looks good, just the one docs comment to address.
> 
> One thing to note. Setting `prefer_valid=True` by default may break one (and maybe others?) devkit IT, [found here](https://github.com/CitrineInformatics/nextgen-devkit/blob/master/end-to-end-tests/tests/orion/test_auto_configure_default.py). It may be wise to test against this branch before merging.
> 
> It may also be wise for us to encode the expected behavior of `prefer_valid=True/False` in those tests, but that may be a job for me another day.

I'll add James to the review since he has some more context on this.

I was thinking it would be good to add an IT for this before merging.  I'll make this a draft PR so we don't accidentally merge it prematurely.

@jspeerless I thought we were going to enable this by default once the feature flag was removed.  Is that right?  We could make the default here `prefer_valid=False` which would leave the behavior unchanged and shouldn't cause any breakages.  This would still let our users manually set it to `True` if they wish.

anoto-moniz:
LGTM! (Just waiting on docstring update and whether the default value needs to change)

crleblanc:
I updated the docs with your suggested wording, that's much more customer friendly.

I also successfully ran the ITs using the branch in this PR, here's the Jenkins job: https://jenkins.corp.citrine.io/job/Devkit-custom/584/console.  James mentioned that setting this feature flag to `true` has worked well for the past couple weeks or so, so it seems safe to have the prefer_valid kwarg default to `true`.

anoto-moniz (dismissed):

crleblanc:
I just merged changes from the main branch and bumped the patch version in __version__.py

andyczerwonka (approved):

PR #738, opened 2022-03-16
[PLA-9375] create_or_update propagates branch ID.

anoto-moniz:
# Citrine Python PR

## Description 
seeding.find_or_create.create_or_update attempts to retrieve the passed
resource by name. If successful, the original object is copied, and the
uid is copied from the retrieved resource before updating. We also need
to copy the branch_id if the resource has a branch_id field, which only
applies to design workflows right now, since the branch ID is needed to
perform the update.

### PR Type:


src/citrine/seeding/find_or_create.py
```diff
@@ -147,6 +147,10 @@ def create_or_update(*, collection: Collection[CreationType], resource: Creation
         # Copy so that passed-in resource is unaffected
         new_resource = deepcopy(resource)
         new_resource.uid = old_resource.uid
+        # This is only relevant for locally created design workflows, since
+        # they likely won't have a branch ID but need one to be updated.
+        if hasattr(old_resource, "branch_id") and new_resource.branch_id is None:
```
kroenlein, line 152: Can we do this with class methods instead of python introspection?  Mixins?

kroenlein (commented):
I've been trying to reduce reliance on direct Python introspection.  Can this not be accomplished with `isinstance`?

src/citrine/seeding/find_or_create.py
anoto-moniz: I'm not super clear on what you're thinking. Care to elaborate?

As I see it, this is the only place we need to copy the `branch_id`, and checking the existence of the field instead of the type future-proofs it. And since it's the only place, propagating some method through the resources feels like overkill. But maybe I'm missing an obvious better way.

(PS I do need to tweak this line to check that `new_resource` has `branch_id` before confirming it's unset).

anoto-moniz: Ah, just noticed your overall comment now (I always forget to check that). Yes, I can use `isinstance`, I was just defending against future resources with this pattern. But sure, I'll make that tweak.

src/citrine/seeding/find_or_create.py
```diff
@@ -147,6 +148,10 @@ def create_or_update(*, collection: Collection[CreationType], resource: Creation
         # Copy so that passed-in resource is unaffected
         new_resource = deepcopy(resource)
         new_resource.uid = old_resource.uid
+        # Locally created design workflows likely won't have a branch ID but
+        # need one to be updated.
+        if isinstance(old_resource, DesignWorkflow):
+            new_resource.branch_id = old_resource.branch_id
```
kroenlein, line 154: Thanks.  If you are worried that other types will have `branch_id`s in the future, we should create a `has_branch` mixin for other classes

kroenlein (dismissed):
LGTM

PR #740, opened 2022-03-22
Updating the rewrite links functionality

dleblanc:
Removed the hardcoded rewrite rule from citrine-python. This wouldn't
work because:
 - it used the old port, but localstack now uses the unified port (4566)
- we also need to include S3_USE_SSL and S3_ADDRESSING_STYLE for
localstack access to work
- we only ever called it with an explicit S3 endpoint anyway, so that
hardcoded logic wasn't even called

Hitting S3 on localstack always needs these env vars: S3_USE_SSL,
S3_ADDRESSING_STYLE and S3_ENDPOINT_URL. Those get wired in via the e2e's at it is. I've also included this in the download portion of file_link, allowing some local file tests to work again (with the above variables set).

# Citrine Python PR

## Description 

### PR Type:


fportillo92 (dismissed):
LGTM!

src/citrine/resources/file_link.py
```diff
@@ -427,7 +427,8 @@ def download(self, *, file_link: FileLink, local_path: str):
         content_link_path = file_link.url + '/content-link'
         content_link_response = self.session.get_resource(content_link_path)
         pre_signed_url = content_link_response['pre_signed_read_link']
-        download_response = requests.get(pre_signed_url)
+        rewritten_url = rewrite_s3_links_locally(pre_signed_url, self.session.s3_endpoint_url)
```
dleblanc, line 430: We weren't rewriting links here, which was causing some tests to fail.

fportillo92 (approved):

crleblanc (approved):
This looks like some good work on a bit of tech debt.

PR #741, opened 2022-03-24
Add ability to download off-platform FileLinks

kroenlein:
# Citrine Python PR

## Description 
This PR augments the `files.download` route so that if a user stored a reference to an off-platform resource (e.g., a spreadsheet in their data lake), the file is retrieved.  This parallels the behavior of the UI.

If this is undesirable behavior, we should modify behavior so that a exception is thrown if `files.download` sees an absolute path.  Right now, a very confusing error is thrown if a user attempts this.

There is an argument for inserting complementary changes in `process`, `poll_file_processing_job`, `file_processing_result` and `delete` so that all absolute URLs through exceptions pointing out that an off-platform resource was referenced.

### PR Type:


dleblanc (commented):
I'm not sure where those relative links are coming from, does citrine-python do some magic to make those relative when they come fro the platform?

kroenlein:
> Where do these files live? I assume these are coming from a different source than the backend's (S3 based) files? I'm happy with being more general in that regard.

Literally catch-all for files that should be referenced but do not reside on our system.  So, if someone's Excel workbooks are all on box, users can reference those documents directly.

> I'm not sure where those relative links are coming from, does citrine-python do some magic to make those relative when they come fro the platform?

For example, in [this material history](https://development.citrine-platform.com/api/v1/projects/e2de14ad-a6b5-4c67-8283-e3c579f4f553/material-history/id/d6612b8d-a91b-491a-b2d2-848c59e2fe44), the terminal material references the file link
```JSON
{
	"file_links": [
		{
			"type": "file_link",
			"filename": "toothpick.jpg",
			"url": "projects/21417020-9ade-4fac-9cad-8aaadb7b5ba4/datasets/2abf6aa8-4873-4ce9-96d6-26b11d24e65b/files/008a1e08-87f7-4815-8d68-a0b64ec76f52/versions/80b83efa-ecc6-4ead-950a-ef8e0b7b2631"
		}
	]
}
```
where this is JSON returned from the backend.  Right now (I believe) this is the relative URL that citrine-python uses to resolve to a s3 bucket.

dleblanc:
Hmmn, interesting. File_links is a property on the GEMD objects, but literally only contains the name and the URL. We really should have something richer for determining if it's an internally stored reference or not, relying on the presence or absence of relative paths for determining if it's a platform file link or not seems too implicit to me.

dleblanc (approved):

PR #742, opened 2022-03-30
Feature/pla 8803 add orion integer support

crleblanc:
# Citrine Python PR

This PR adds support for the IntegerDescriptor type.

## Description 

This PR adds a new IntegerDescriptor class.  It's unit is always "dimensionless", but we could add support for a 'units' kwarg in the future.  I have some additional e2e tests that exercise this new type and they pass (will create a PR when this is merged).

### PR Type:

  This is a new feature and should not create any downstream consequences.


src/citrine/informatics/descriptors.py
```diff
@@ -82,6 +99,46 @@ def __repr__(self):
...
+
+    def __init__(self,
+                 key: str,
+                 *,
+                 lower_bound: int,
+                 upper_bound: int):
+        self.key: str = key
+        self.lower_bound: int = lower_bound
+        self.upper_bound: int = upper_bound
+        self.units = "dimensionless"
```
kroenlein, line 132: Should this be a default or just enforced as constant?

src/citrine/informatics/descriptors.py
```diff
@@ -30,8 +31,30 @@ def get_type(cls, data) -> Type[Serializable]:
...
+        attrs: List[str]
+            A list of attribute names to lookup and compare
+
+        """
+        try:
+            return all([
+                self.__getattribute__(key) == other.__getattribute__(key) for key in attrs
+            ])
+        except AttributeError:
+            return False
```
kroenlein, line 56: I realize that this is just abstracted code (good call), but in the case where both objects are missing the same key, it returns False even though the objects are identical.  Not sure how much it matters.

kroenlein (commented):
Minor questions.  Generally seems reasonable.

src/citrine/informatics/descriptors.py
crleblanc: I was on the fence about this one.  We could remove this line and the code will continue to work, due to our 'units' type hint which defaults to 'dimensionless'.

I'm not clear on how what you mean by enforce it as a constant.  Do you mean `self.unit = DIMENSIONLESS` which is defined elsewhere, e.g. the top of the file?  This could also be a class attribute, say `__DIMENSIONLESS` to discourage modification.

src/citrine/informatics/descriptors.py
crleblanc: I think there's a good solution here.  If any of the `attrs` are missing from `self`, then we should raise an exception, because this is a clear programmer error.  We probably don't want to do this for `other` instance, because this could be any type of class which may not have this attrs.  The unit tests expect `descriptor == None` to be `False` which seems reasonable.

src/citrine/informatics/descriptors.py
kroenlein: Your approach would have done sufficed as well.  I meant mostly removing any route for a user to set the value (short of Python object hackery).

tests/builders/test_descriptors.py
```diff
@@ -33,23 +34,28 @@
 from citrine.builders.auto_configure import AutoConfigureMode
 
 
-density_desc = RealDescriptor("density", lower_bound=0, upper_bound=100, units="gram / centimeter ** 3")
+density_desc = RealDescriptor("density", lower_bound=0,
+                              upper_bound=100, units="gram / centimeter ** 3")
```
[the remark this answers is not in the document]
crleblanc: I see VSCode is using autopep8 as it's formatter.  I'll see if I use flake8 instead.

src/citrine/informatics/descriptors.py
crleblanc: Removing the instance attribute works fine, so I'll go with that approach.

kroenlein (approved):
LGTM

PR #745, opened 2022-04-05
Download by files by name

kroenlein:
# Citrine Python PR

## Description 
This PR adds functionality to the files route so users can retrieve files based upon a FileLink object, a filename or the file UUID.

This unifies logic and removes sensitivity to elements that the backend ignores when resolving URLS (project & dataset in the FileLink URL).  A longer term correction would likely be to store either only `fileID/versionID`, `http://datasvc.instance.citrine-platform.net:8002/api/v1/files/{fileID}/versions/{versionID}` (the `versioned_url` returned by the platform) or a citrine-specific constructed URI for file identification.  All 3 should be supported by the current logic.  Deferring that work for the necessary cross-team coordination.

This uncovered a bug where the `id` field of the returned file object is actually the versionID, not the fileID as documented in the API.  A bug has been filed to address that issue.  PLA-9482.

### PR Type:


kroenlein:
Passing integration tests:
https://jenkins.corp.citrine.io/job/nextgen-ITs/26070/

src/citrine/resources/file_link.py
```diff
@@ -149,6 +149,31 @@ def __init__(self, project_id: UUID, dataset_id: UUID, session: Session):
...
+            new_path.append(quote(action))
+
+        return base._replace(path='/'.join(new_path)).geturl()
+
+    def _get_path_from_file_link(self, file_link: FileLink,
+                                 *,
+                                 action: str = None) -> str:
+        """Build the path for taking an action with a particular file link."""
+        # Use this sessions project/dataset credentials and the URL's file / version
+        parsed = urlparse(file_link.url)
```
dleblanc, line 171: Shouldn't the file link object itself just store the file and version IDs are properties? Actually seems like this whole thing should live in FileLink, no?

src/citrine/resources/file_link.py
```diff
@@ -547,10 +575,77 @@ def delete(self, file_link: FileLink):
...
-        assert split_url[-2] == 'versions' and split_url[-4] == 'files', \
-            "File URL is expected to end with '/files/{{file_id}}/version/{{version id}}', " \
-            "but FileLink instead has url {}".format(file_link.url)
         file_id = split_url[-3]
         data = self.session.delete_resource(self._get_path(file_id))
         return Response(body=data)
+
+    def _resolve_file_link(self, identifier: Union[str, UUID]) -> FileLink:
+        """Generate the FileLink object referenced by the passed argument."""
+        if isinstance(identifier, FileLink):
```
dleblanc, line 589: Isn't a FileLink outside the expected types of {str, UUID}? Is this just a "if they passed us a file link" exception case?

src/citrine/resources/file_link.py
```diff
@@ -547,10 +575,77 @@ def delete(self, file_link: FileLink):
...
+            filename = urlparse(identifier).path.split('/')[-1]
+            file_dict = {
+                'url': identifier,
+                'filename': filename,
+                'type': GEMDFileLink.typ
+            }
+            return FileLink.build(file_dict)
+        elif isinstance(identifier, str) and self._validate_filelink_url(identifier):
+            # It's an on-platform URL
+            hit = next((f for f in self.list() if identifier == f.url), None)
```
dleblanc, line 614: Why do you have to iterate with .list here? This looks inefficient.

dleblanc (commented):
I have some inline questions about the approach you're using.

src/citrine/resources/file_link.py
kroenlein: We should arguably be storing the versioned_url returned by the `files` routes, but the S3 resolution route on the backend literally expects a relative URL of the structure `/....../dataset/{uuid}/file/{uuid}/project/{uuid}`.  I'd like to move this full protocol over to where FileLinks are only storing those two pieces of information for platform files.  That will need to be a multi-PR multi-library strategy because we don't want to break anything that exists today.  This PR makes citrine-python flexible in what it consumes while still emitting what other tool expect.

[This parse](https://github.com/CitrineInformatics/platform-backend/blob/1cd99ce0cee1acc652d31027f400268f67305811/code/rest/clients/src/main/scala/io/citrine/rest/clients/data/api/impl/akka/FileServiceApiImpl.scala#L84-L88) feeds [this construction](https://github.com/CitrineInformatics/platform-backend/blob/1cd99ce0cee1acc652d31027f400268f67305811/code/rest/clients/src/main/scala/io/citrine/rest/clients/data/api/impl/akka/FileServiceApiImpl.scala#L70).  Note that nothing is done with datasetID even though the parse requires it.  If we got rid of that, we could just store that URL.

src/citrine/resources/file_link.py
kroenlein: No, I just failed to implement a consistent typing here because elsewhere I decided I wanted to avoid testing type in every method that leverages this one.

src/citrine/resources/file_link.py
kroenlein: There's no direct get-by-url, but doing the parse here makes sense.  I think I was trying to reduce the amount of URL twiddling that was everywhere in this module.  Will update.

kroenlein:
ITs:
https://jenkins.corp.citrine.io/job/nextgen-ITs/26144
https://jenkins.corp.citrine.io/job/nextgen-ITs/26145

src/citrine/resources/file_link.py
kroenlein: Added `FileLink` to the argument set.

src/citrine/resources/file_link.py
kroenlein: Reworked to use get, though annoyingly get was broken.  The files routes have different call structure than other calls in the API.

In particular see [files.get](https://docs.corp.citrine.io/shire/pippin/api/index.html#tag/Files/paths/~1projects~1{projectId}~1datasets~1{datasetId}~1files~1{fileId}/get) and [files.versions.get](https://docs.corp.citrine.io/shire/pippin/api/index.html#tag/Files/paths/~1projects~1{projectId}~1datasets~1{datasetId}~1files~1{fileId}~1versions~1{versionId}/get)

xperrylinn:
@kroenlein,

I tested the functionality using the line: 

` dataset.files.download(file_link=dataset.files.get("Recipe_filtered_out_non_agg.csv"), local_path="~/Downloads/")`

expecting the file to be downloaded to my Downloads folder, but it created new folders named `~` and `Downloads` placing the downloaded file into `citrine-python/~/Downloads/Recipe_filtered_out_non_agg.csv`. 

After looking at the documentation it's unclear if local path is relative or absolute, but it seems to be relative.

kroenlein:
@xperrylinn Finally got around to revisiting this, and the rework should resolve the issue you were having with expanding `~` ([docs](https://docs.python.org/3.10/library/os.path.html#os.path.expanduser)).  Can you test again?

tests/resources/test_file_link.py
```diff
@@ -92,25 +93,22 @@ def test_delete(collection, session):
...
+    # A remote URL is invalid
+    ext_invalid_url = f'http://www.citrine.io/develop/files/{file_id}/versions/{version_id}'
+    ext_invalid_file_link = collection.build(FileLinkDataFactory(url=ext_invalid_url))
+    with pytest.raises(ValueError):
+        collection.delete(ext_invalid_file_link)
 
-    mock_isfile.return_value = True
-    mock_stat.return_value = StatStub(st_size=22300)
-    mock_open.return_value.__enter__.return_value = 'Random file contents'
-    mock_boto3_client.return_value = FakeS3Client({'VersionId': '3'})
```
kroenlein, line 112: All of this was actually testing implementation details that got broken in the swap over to pathlib.

src/citrine/_utils/functions.py
```diff
@@ -117,15 +118,20 @@ def rewrite_s3_links_locally(url: str, s3_endpoint_url: str = None) -> str:
...
-        os.makedirs(directory)
-    with open(local_path, 'wb') as output_file:
-        output_file.write(content)
+    if isinstance(local_path, str):
+        if len(os.path.split(local_path)[-1]) == 0:
+            raise ValueError(f"A filename must be provided in the path ({local_path})")
+        local_path = Path(local_path)
+
+    # Resolve ~, .., and the like
+    local_path = local_path.expanduser().resolve()
```
dleblanc, line 129: That's handy!

dleblanc (approved):

PR #751, opened 2022-04-26
Using _cached_isinstance in property ser/deser

kroenlein:
# Citrine Python PR
https://citrine.atlassian.net/browse/PLA-9477

## Description 
This PR should substantially speed up end-to-end tests by using a type-caching version of is_instance.

### PR Type:


crleblanc (approved):
Ah, return of the LRU cache!  It'll be interesting to compare the callgraphs before and after using this.

It does seem odd to import a function with a leading underscore, but given that we own both citrine-python and gemd-python I'm happy to approve.

kroenlein:
> It does seem odd to import a function with a leading underscore, but given that we own both citrine-python and gemd-python I'm happy to approve.

Agree, so if this does what we want, we may want to change that to a public function in gemd-python.  But I think the short-term win is worth the faux pas.

Integration tests:
https://jenkins.corp.citrine.io/job/nextgen-ITs/26683
https://jenkins.corp.citrine.io/job/nextgen-ITs/26684

crleblanc (approved):
LGTM!

PR #753, opened 2022-05-02
[PLA-9255] Switching to v3 predictor endpoints.

anoto-moniz:
Predictor capabilities have been made available on their own endpoint: /api/v3/predictors. This PR adjusts the SDK to point at that new set of endpoints.

The biggest change is the shape of the data coming back from the endpoints. It still fits into the existing objects, but requires some changes to handling.

Additionally, some of the endpoint behavior has changed, which the SDK must account for in order to preserve backwards compatibility. Some fields have been removed in the new API, so corresponding fields (or methods) are now deprecated in the SDK. For example, the update endpoint no longer archives or trains. Instead, the SDK makes a separate call for archiving when appropriate, and always makes another call for training. We also deprecate the previous ability to set an `archived` property on update to trigger archiving, in favor of dedicated archive/restore methods.

Note that the v1 and v2 endpoints remain available unchanged, so users on older versions of the SDK won't experience any problems. Those will eventually go away, but not without thorough communication with the customer teams and lead time to upgrade, as well as tracking their usage on the server. 

### PR Type:


anoto-moniz:
There are some more small changes to the Predictor V3 API coming which will impact the SDK. As such, I'm going to close this PR until they're in and I've updated the SDK accordingly.

https://citrine.atlassian.net/browse/PLA-9628

PR #760, opened 2022-06-02
Branch docs update

jspeerless:
# Citrine Python PR

## Description 
Updating the documentation around branches and their relationship to design workflows, specifically with respect to archiving. Some confusion has arisen around this relationship which has resulted in bugs reported (https://citrine.atlassian.net/browse/PLA-9677).

### PR Type:


docs/source/workflows/design_workflows.rst
```diff
@@ -182,6 +187,10 @@ A branch has a name, along with any number of design workflows. A DesignWorkflow
 
 When you're done with a branch, it can be archived, removing it from the results of ``list`` and setting the ``archived`` flag.
 ``list_archived`` lists all archived branches in a project. An archived branch can be restored via its unique ID.
+Note that archiving branches are independent of archiving the design workflows contained within it.
+Archiving a branch, by removing it from listing methods, will hide the entire branch from default displays in the web UI.
```
anoto-moniz, line 191: This is a bit of a confusing sentence as written. I think dropping `by removing it from listing methods` will help a lot. And we mention above that archiving a branch will hide it from listing, so that information isn't lost.

docs/source/workflows/design_workflows.rst
```diff
@@ -182,6 +187,10 @@ A branch has a name, along with any number of design workflows. A DesignWorkflow
...
+Similarly, archiving a design workflow will result in its executions and relevant assets no longer being displayed within the branch.
+Yet archiving a branch will NOT change the archived status of the contained design workflows in the context of design workflow listing methods.
```
anoto-moniz, line 193: You can italicize using asterisks, which I think we should prefer over all caps for emphasis.

docs/source/workflows/design_workflows.rst
```diff
@@ -182,6 +187,10 @@ A branch has a name, along with any number of design workflows. A DesignWorkflow
 
 When you're done with a branch, it can be archived, removing it from the results of ``list`` and setting the ``archived`` flag.
 ``list_archived`` lists all archived branches in a project. An archived branch can be restored via its unique ID.
+Note that archiving branches are independent of archiving the design workflows contained within it.
```
anoto-moniz, line 190: Replace "archiving branches are" with "archiving branches is".

docs/source/workflows/design_workflows.rst
wrerwin, line 191: It may be worth mentioning here that branches are the only place where design workflows are displayed in the UI. One point that is not fully addressed here is that while design workflows are not archived when a parent branch is archived, the design workflows are no longer visible in the UI. This was a major point of confusion for me initially, and I think it would be helpful to call this behavior out explicitly.

claudiaviquez: +1 on Will's comment, and wanted to add that the reverse behavior is also confusing for (probably) the same reason - if I archive all the workflows in a branch, the branch will still exist but won't be shown on the UI

jspeerless: Let me know if the most recent version makes the relationship more clear. I would like to avoid being too specific about the behavior of our UI, given that it can change independently of our SDK.

jspeerless: Change made, with the comments from other teammates taken into account below. LMK if things are a bit clearer this time around.

wrerwin (commented):
The way this is documented clearly describes the behavior to me now.

anoto-moniz (dismissed):

anoto-moniz (approved):

PR #764, opened 2022-06-10
Added the Team resource list_resources() method and a test

crleblanc:
# Citrine Python PR

## Description 
As per PLA-9753 this PR implements the listing of resources for a team via the `list_resoures` method on the Team class.

### PR Type:


src/citrine/resources/team.py
```diff
@@ -13,7 +13,9 @@
 WRITE = "WRITE"
 READ = "READ"
 SHARE = "SHARE"
-TEAM_ACTIONS = Union[WRITE, READ, SHARE]
+CREATE = "CREATE"
+PUBLISH = "PUBLISH"
+TEAM_ACTIONS = Union[WRITE, READ, SHARE, CREATE, PUBLISH]
```
crleblanc, line 18: @zoesyang Can you confirm that these team actions should be added?  I've seen them mentioned elsewhere so I'm guessing they're missing from this file but would like to double check.

src/citrine/resources/team.py
```diff
@@ -224,6 +226,30 @@ def un_share(self, *, resource: Resource, target_team_id: Union[str, UUID]) -> b
         )
         return True
 
+    def list_resources(self, resource_type: str, action: str) -> List[str]:
```
crleblanc, line 229: I plan to replace 'str' with a better type here for resource_type and action.

kroenlein (commented):
Non-blocking question.  Also, you've prepped an integration / nextgen-devkit e2e for this?

src/citrine/__version__.py
```diff
@@ -1 +1 @@
-__version__ = '1.33.1'
+__version__ = '1.34.1'
```
kroenlein, line 1: Interesting version bump...

crleblanc: I'm left guessing what you mean.  As per semver.org:

Given a version number MAJOR.MINOR.PATCH, increment the:

    MAJOR version when you make incompatible API changes,
    MINOR version when you add functionality in a backwards compatible manner, and
    PATCH version when you make backwards compatible bug fixes.

It's new functionality that is backwards compatible, hence a minor version bump.

src/citrine/resources/team.py
```diff
@@ -224,6 +232,30 @@ def un_share(self, *, resource: Resource, target_team_id: Union[str, UUID]) -> b
...
+        -------
+        List[str]
+            The ids of the resources in this team where the owner has the specified action
+
+        """
+        query_params = {"domain": self._path(), "action": action}
+        return self.session.get_resource(f"/{resource_type}/authorized-ids",
+                                         params=query_params,
+                                         version="v3")['ids']
+
```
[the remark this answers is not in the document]
crleblanc: I see there was discussion about this approach on slack, but I didn't see the responses in the thread (ps: damn you Slack).  The API is `/api/v3/.teamId/authorized-ids` - hence the 1:1 mapping with teams.  I'm fine to close this PR and implement it the way you suggest, if this is the agreed approach.

PR #766, opened 2022-06-24
FakeCall should capture the endpoint version.

anoto-moniz:
As we add more endpoints not on v1 (see: predictors, design workflows, etc), it's becoming more useful to confirm the version being called is the correct one. However, at least at the moment, it would be onerous to force every test to do so. So when comparing two `FakeCall` objects, if both have a version, then it's included in the equality check. 

### PR Type:


tests/utils/session.py
```diff
@@ -10,19 +10,25 @@
...
     def __repr__(self):
         return 'FakeCall({})'.format(self)
 
     def __str__(self) -> str:
-        path = self.path
+        if self.version:
+            path = self.path[1:] if self.path[0] == '/' else self.path
+            path = '{}/{}'.format(self.version, path)
+        else:
+            path = self.path
```
kroenlein, line 28:
```suggestion
        path = self.path

        if self.version is not None:
            start = int(path.startswith('/'))  # 0 or 1
            path = f'{self.version}/{path[start:]}'
```

tests/utils/session.py
```diff
@@ -33,7 +39,8 @@ def __eq__(self, other) -> bool:
         return self.method == other.method and \
             self.path == other.path and \
             self.json == other.json and \
-            self.params == other.params
+            self.params == other.params and \
+            (not self.version or not other.version or self.version == other.version) # Allows users to check the URL version without forcing everyone to.
```
kroenlein, line 43:
```suggestion
            (self.version is None or other.version is None or self.version == other.version) # Allows users to check the URL version without forcing everyone to.
```
Many more things evaluate to False than `None`.

Are you sure it should pass if one case has a None and the other does not?

tests/utils/session.py
```diff
@@ -77,26 +84,26 @@ def delete_resource(self, path: str, **kwargs) -> dict:
         return self.checked_delete(path, **kwargs)
 
     def checked_get(self, path: str, **kwargs) -> dict:
-        self.calls.append(FakeCall('GET', path, params=kwargs.get('params')))
+        self.calls.append(FakeCall('GET', path, params=kwargs.get('params'), version=kwargs.get('version')))
```
kroenlein, line 87: At this point, it almost makes more sense to pass kwargs into the FakeCall constructor.

kroenlein (approved):
Nothing blocking, but points for consideration.

tests/utils/session.py
```diff
@@ -10,19 +10,25 @@
...
-        path = self.path
+        if self.version:
+            path = self.path[1:] if self.path[0] == '/' else self.path
+            path = '{}/{}'.format(self.version, path)
+        else:
+            path = self.path
+
         if self.params:
-            path = '{}?{}'.format(self.path, urlencode(self.params))
+            path = '{}?{}'.format(path, urlencode(self.params))
```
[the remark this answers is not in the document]
anoto-moniz: I was imitating the existing usage of `.format` (which I think was only one place, to be fair), but I also prefer f-strings.

tests/utils/session.py
anoto-moniz: Correct, and the only time they should be checked is if they're truth-y. Which is why the recommendation is to prefer truthiness checking over explicit `None` checking if at all possible.

And yes. Failing if only one is set means that every time someone performs a check against `FakeCall` they must specify the version, which is onerous since in many case they don't care. But it allows tests which do care to perform such a check.

kroenlein (approved):
Philosophical disagree on the `not` v. `is None` test, but that's not all that problematic since it's a test routine and not something user facing.

PR #769, opened 2022-07-07
Feature/pla 9753 expose cross team resource listing endpoints

crleblanc:
# Citrine Python PR

## Description 
As per PLA-9753 this PR gives users a way of listing published resources in a team, e.g. `team.dataset_ids.list_readable()`.  This returns a list of IDs.  This is not a great user experience, but our auth layer only records UIDs, so we would need additional APIs to hydrate these resources efficiently (e.g. getting a batch of resources).

There will be a PR for devkit to include an e2e for this functionality, once this PR is merged.

### PR Type:

### DIFF
--- a/src/citrine/_rest/collection.py
+++ b/src/citrine/_rest/collection.py
@@ -1,4 +1,3 @@
-import warnings
 from abc import abstractmethod
 from logging import getLogger
 from typing import Optional, Union, Generic, TypeVar, Iterable, Iterator
@@ -137,5 +136,5 @@ class Collection(Generic[ResourceType], Pageable):
             except(KeyError, ValueError) as e:
                 # TODO(PLA-9109): This is a patch to handle deprecated predictors client side
                 # Remove when predictors are migrated
-                warnings.warn(f"Building element skipped due to error: {e}", UserWarning)
+                logger.warning(f"Building element skipped due to error: {e}")
                 pass
--- a/src/citrine/builders/auto_configure.py
+++ b/src/citrine/builders/auto_configure.py
@@ -1,6 +1,6 @@
+from logging import getLogger
 from uuid import UUID
 from typing import Union, Optional, List
-import warnings
 
 from gemd.entity.link_by_uid import LinkByUID
 from gemd.enumeration.base_enumeration import BaseEnumeration
@@ -23,6 +23,8 @@ from citrine.resources.project import Project
 from citrine.resources.table_config import TableConfig, TableBuildAlgorithm
 
 from citrine.builders.scores import create_default_score
+
+logger = getLogger(__name__)
 
 
 class AutoConfigureStatus(BaseEnumeration):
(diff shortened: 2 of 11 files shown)


src/citrine/resources/team.py
```diff
@@ -33,6 +33,52 @@ def __str__(self):
...
+        return format_escaped_url(f'/teams/{str(self.team_id)}')
+
+    def _list_ids(self, action: str) -> List[str]:
+        query_params = {"domain": self._path(), "action": action}
+        return self.session.get_resource(f"/{self.resource_type}/authorized-ids",
+                                         params=query_params,
+                                         version=self._api_version)['ids']
+
+    def list_readable(self):
+        return self._list_ids(action="READ")
```
kroenlein, line 73: This should be the module constant.
```suggestion
        return self._list_ids(action=READ)
```

src/citrine/resources/team.py
```diff
@@ -33,6 +33,52 @@ def __str__(self):
...
+
+    def list_writeable(self):
+        return self._list_ids(action="WRITE")
```
kroenlein, line 76: This should be the module constant.
```suggestion
        return self._list_ids(action=WRITE)
```

src/citrine/resources/team.py
```diff
@@ -33,6 +33,52 @@ def __str__(self):
...
+
+    def list_shareable(self):
+        return self._list_ids(action="SHARE")
```
kroenlein, line 79: This should be the module constant.
```suggestion
        return self._list_ids(action=SHARE)
```

src/citrine/resources/team.py
```diff
@@ -33,6 +33,52 @@ def __str__(self):
...
+    def __init__(self,
+                 session: Session,
+                 team_id: Union[str, UUID],
+                 resource_type: str) -> None:
+        self.session = session
+        self.team_id = team_id
+        self.resource_type = resource_type
+
+    def _path(self) -> str:
+        return format_escaped_url(f'/teams/{str(self.team_id)}')
```
kroenlein, line 64: I believe the f-string automatically casts as `str` by default
```suggestion
        return format_escaped_url(f'/teams/{self.team_id}')
```

kroenlein (changes requested):
The build a collection then enumerate a set of methods doesn't quite align with my design aesthetic, but it's internally consistent and gets the job done.  I'll approve once you address the specific changes below.

( I'd probably have created one `resource_ids` method with the access and type arguments that take enumerations. )

kroenlein (approved):
Makes sense as written.  

Would make more sense to me grammatically if the syntax were:
```python
team.datasets.list_readable_ids()
```
but I appreciate that implementing that would be more complicated.

crleblanc:
> Makes sense as written.
> 
> Would make more sense to me grammatically if the syntax were:
> 
> ```python
> team.datasets.list_readable_ids()
> ```
> 
> but I appreciate that implementing that would be more complicated.

I agree that this would be much cleaner, but this is not the DatasetCollection instance we get elsewhere, e.g. project.datasets.  I suspect this could get confusing for users who would expect a fully hydrated resource collection but instead get IDs back.

I wonder if we'll have to revisit this in the future and make batch-get endpoints to facilitate a proper resource collection?  The user workflow would be smoother.

PR #770, opened 2022-07-08
Add get Team Member for a given User

kroenlein:
# Citrine Python PR
Test: https://jenkins.corp.citrine.io/job/nextgen-ITs/28729/

## Description 
This PR exposes the /api/v3/teams/:teamId/users/:userId (unexposed) route so that a user can access their permissions.  Also adds a convenience route `team.me()` to specifically make asking about one's own permissions straight-forward.

### PR Type:


tests/resources/test_user.py
```diff
@@ -127,8 +127,6 @@ def test_delete_user(collection, session):
 def test_get_me(collection, session):
     # Given
     user = UserDataFactory()
-    user["id"] = user["id"]
-    del user["id"]
```
btreecat, line 131: What do you reckon the original reason these were here was?

btreecat (dismissed):
Great work, thank you. One question that is for my own clarification.

src/citrine/resources/team.py
```diff
@@ -99,37 +101,81 @@ def list_members(self) -> List[TeamMember]:
                                             version=self._api_version)["users"]
         return [TeamMember(user=User.build(m), team=self, actions=m["actions"]) for m in members]
 
-    def remove_user(self, user_id: Union[str, UUID]) -> bool:
+    def get_member(self, user_id: Union[str, UUID, User]) -> TeamMember:
```
anoto-moniz, line 104: Informational question: are you thinking we should start to introduce this pattern across the SDK? Because there are a bunch of places we accept the UUID, but not the object itself.

src/citrine/resources/team.py
```diff
@@ -99,37 +101,81 @@ def list_members(self) -> List[TeamMember]:
...
+        Returns
+        -------
+        TeamMember
+            The requested team member
+
+        """
+        if isinstance(user_id, User):
+            user_id = user_id.uid
+        path = self._path() + format_escaped_url('/users/{user_id}', user_id=user_id)
+        member = self.session.get_resource(path=path, version=self._api_version)["user"]
```
anoto-moniz, line 124: Won't this raise a `KeyError` on a 404 (or other error) and obscure their mistake as a result?

src/citrine/resources/team.py
```diff
@@ -99,37 +101,81 @@ def list_members(self) -> List[TeamMember]:
                                             version=self._api_version)["users"]
         return [TeamMember(user=User.build(m), team=self, actions=m["actions"]) for m in members]
 
-    def remove_user(self, user_id: Union[str, UUID]) -> bool:
+    def get_member(self, user_id: Union[str, UUID, User]) -> TeamMember:
         """
-        Remove a User from a Team.
+        Get a particular member in the current team.
+
+        May require admin privileges depending on which user is being requested.
```
anoto-moniz, line 108: Informational question: do you know what determines if you need admin privileges or not?

src/citrine/resources/team.py
```diff
@@ -165,16 +217,21 @@ def update_user_action(self, user_id: Union[str, UUID], *,
...
         self.session.checked_put(self._path() + "/users", version=self._api_version,
                                  json={'id': str(user_id), "actions": actions})
         return True
 
-    def share(self, *,
+    def share(self,
+              *,
               resource: Resource,
-              target_team_id: Union[str, UUID]) -> bool:
+              target_team_id: Union[str, UUID, "Team"]) -> bool:
```
anoto-moniz, line 229: You document that the user can provide a `Team` object, but I don't see any handling for that input below.

src/citrine/resources/team.py
kroenlein: citrine-python raises a Unauthorized exception earlier in the stack if you don't get a 200.  The only case where this misbehaves should if the platform returned a 200 but the JSON didn't conform to the API.  If this is a concern, I'm happy to add additional care (though this follows the pattern established in `list_members`).

src/citrine/resources/team.py
kroenlein: I believe it's whether you are just asking about yourself, or asking about other users in the group.  Not confident in that answer.

src/citrine/resources/team.py
kroenlein: Oh, good catch.  I neglected to add the map-team-to-uid conditional on this one.

src/citrine/resources/team.py
kroenlein: There are also a number of places where we use this pattern.  I've been slowly migrating over to it when I touch files.

anoto-moniz: Gotcha, I'll start doing the same.

anoto-moniz (approved):

PR #779, opened 2022-08-12
convert batch_size into a package constant

StevenKauwe:
# Citrine Python PR

## Description 
The current hypothesis is that sometimes the objects we send to datasvc are so big that they cause a time-out during writing. @Kenneth Kroenlein One workaround for this is to reduce the batch-size configuration in citrine-python/gemd_resource.py from 50 to a lower value.
https://citrine.atlassian.net/browse/PLA-10156

To address the user pain point, we should change batch_size = 50 to a package constant BATCH_SIZE = 50 so that users can handle this case without editing citrine-python source code.

### PR Type:


kroenlein (approved):
LGTM!

Normally, I'd expect at least one new test with any PR, but this refactor is straightforward and doesn't really change any public functionality.

PR #780, opened 2022-08-31
PLA-10206: User configure auto ml

sfriedowitz:
# Citrine Python PR

Adds support to configure estimator types for AutoML.

### PR Type:


src/citrine/resources/predictor.py
```diff
@@ -81,9 +85,13 @@ def update(self, predictor: Predictor) -> Predictor:
             self.restore(predictor.uid)
         predictor._archived = None
 
-        # We never exposed saving a model without training, so we should
-        # continue to do it automatically.
-        return self._train(predictor.uid)
+        # If the initial response is invalid, just return it
```
sfriedowitz, line 88: We will never actually hit this branch until the API bug fix is in, but the logic is correct and tested for now.

src/citrine/informatics/predictors/auto_ml_predictor.py
```diff
@@ -54,10 +88,12 @@ def __init__(self,
                  output: Descriptor = None,
                  outputs: List[Descriptor] = None,
                  inputs: List[Descriptor],
+                 estimators: Optional[Set[AutoMLEstimator]] = None,
                  training_data: Optional[List[DataSource]] = None):
         self.name: str = name
         self.description: str = description
         self.inputs: List[Descriptor] = inputs
+        self.estimators: Set[AutoMLEstimator] = estimators or {AutoMLEstimator.RANDOM_FOREST}
```
kroenlein, line 96: This replaces an explicitly passed empty set with the default value.  Intended?

kroenlein (commented):
I didn't see a test verifying that an arbitrary string is not a valid `AutoMLEstimator`.  Intentional?

src/citrine/informatics/predictors/auto_ml_predictor.py
sfriedowitz: Hmm, passing an empty set is guaranteed to fail in the backend, just later on.

I suppose we could allow it here, but it will always result in an invalid module.

sfriedowitz: Ehh I think it makes sense to coerce to a valid default here.

kroenlein (approved):
LGTM.  Get an e2e in there when you get a chance.

PR #782, opened 2022-09-02
Update predictor to properly present archive status.

anoto-moniz:
In the near future, we'll be adding support for predictor versions. There will be a separation between a predictor version and the predictor root, which can be archived independently. However, the predictor root will be invisible to the user. Since this logic already exists in the backend, that means the value of "is_archived" may no longer match the expected value.

Until we expose that separation and make this explicit, we'll overwrite the archive status to present the root's status, which matches what the user previously expected. In addition, we point users at the new methods, which act on the predictor root.

### PR Type:


src/citrine/__version__.py
```diff
@@ -1 +1 @@
-__version__ = '1.38.0'
+__version__ = '1.38.1'
```
sfriedowitz, line 1: Think this should be a minor b/c adding new `archive_root` and `restore_root` methods.

src/citrine/resources/predictor.py
```diff
@@ -96,38 +128,69 @@ def update(self, predictor: Predictor) -> Predictor:
...
+        return any(uid == archived_pred.uid for archived_pred in self.list_archived())
+
+    @deprecated(deprecated_in="1.38.1", removed_in="2.0.0",
+                details="archive() is deprecated in favor of archive_root().")
+    def archive(self, uid: Union[UUID, str]) -> Predictor:
+        """[DEPRECATED] Archive a root predictor."""
+        self.archive_root(uid)
+        return self.get(uid)
+
+    @deprecated(deprecated_in="1.38.1", removed_in="2.0.0",
```
sfriedowitz, line 169: Update deprecated calls if changing version to a minor.

sfriedowitz (commented):
looks good, just a version comment.

src/citrine/__version__.py
anoto-moniz: That's a fair point.

src/citrine/resources/predictor.py
```diff
@@ -52,20 +54,50 @@ def _predictors_path(self, subpath: str, uid: Union[UUID, str] = None):
         path_template = self._path_template + (f"/{uid}" if uid else "") + f"/{subpath}"
         return format_escaped_url(path_template, project_id=self.project_id)
 
+    # Until we expose predictor versions in the SDK, and thus expose the
+    # difference between archving a predictor root and version, we should
+    # report the root's archive status.
```
kroenlein, line 59: Should this be labeled with a `TODO`?

src/citrine/resources/predictor.py
```diff
@@ -52,20 +54,50 @@ def _predictors_path(self, subpath: str, uid: Union[UUID, str] = None):
...
+    def _inject_archive_info(self, predictor, is_archived=None):
```
kroenlein, line 60:
```suggestion
    def _inject_archive_info(self, predictor: Predictor, is_archived: bool = None) -> Predictor:
```
It's internal, but typehints are still convenient

src/citrine/resources/predictor.py
```diff
@@ -52,20 +54,50 @@ def _predictors_path(self, subpath: str, uid: Union[UUID, str] = None):
...
+        if archived:
+            predictor.archived_by = predictor.updated_by
+            predictor.archive_time = predictor.update_time
+        else:
+            predictor.archived_by = None
+            predictor.archive_time = None
+        return predictor
+
+    # Same as above
+    def _build_collection_elements(self, collection, is_archived=None):
```
kroenlein, line 72:
```suggestion
    def _build_collection_elements(self, collection: PredictorCollection, is_archived: bool = None) -> Predictor:
```
and so on...

src/citrine/resources/predictor.py
```diff
@@ -96,38 +128,69 @@ def update(self, predictor: Predictor) -> Predictor:
...
+        """Archive a root predictor.
 
         uid: Union[UUID, str]
             Unique identifier of the predictor to archive.
 
         """
         path = self._predictors_path("archive", uid)
-        entity = self.session.put_resource(path, {}, version=self._api_version)
-        return self.build(entity)
+        self.session.put_resource(path, {}, version=self._api_version)
```
kroenlein, line 142: I don't necessarily agree with always returning an object, but you are changing the pattern here.

src/citrine/resources/predictor.py
```diff
@@ -96,38 +128,69 @@ def update(self, predictor: Predictor) -> Predictor:
...
+        """Restore an archived root predictor.
 
         uid: Union[UUID, str]
             Unique identifier of the predictor to restore.
 
         """
         path = self._predictors_path("restore", uid)
-        entity = self.session.put_resource(path, {}, version=self._api_version)
-        return self.build(entity)
+        self.session.put_resource(path, {}, version=self._api_version)
```
kroenlein, line 152: Same as above

tests/resources/test_predictor.py
```diff
@@ -50,48 +54,58 @@ def test_delete():
 
 
 def test_archive_and_restore(valid_label_fractions_predictor_data):
-    session = mock.Mock()
+    session = FakeSession()
```
kroenlein, line 57: Thank you for continuing this migration

tests/resources/test_predictor.py
```diff
@@ -50,48 +54,58 @@ def test_delete():
...
     pc = PredictorCollection(uuid.uuid4(), session)
     entity = deepcopy(valid_label_fractions_predictor_data)
     entity["metadata"]["archived"] = entity["metadata"]["created"]
-    session.put_resource.return_value = entity
+    session.set_responses(None, entity, paging_response(entity))
 
-    archived_predictor = pc.archive(uuid.uuid4())
+    with pytest.deprecated_call():
+        archived_predictor = pc.archive(uuid.uuid4())
```
kroenlein, line 64: It seems like you are missing the chance to also test the new method directly.  I understand that the test cascades, but the person following after you may not grasp that.

kroenlein (commented):
Generally good.  Some minor questions and tweaks.

src/citrine/resources/predictor.py
anoto-moniz: Sure.

src/citrine/resources/predictor.py
anoto-moniz: Fair enough.

tests/resources/test_predictor.py
anoto-moniz: Good point. I'll break them out.

src/citrine/resources/predictor.py
anoto-moniz: That's part of why it's a new method.

kroenlein (approved):
LGTM

PR #783, opened 2022-09-20
[PLA-10357] Drop 3.6 support

anoto-moniz:
Python 3.6 reached EOL last year, and some of the tooling we use to release it has begun complaining. We've communicated to the customer team this removal of support, and intend to give them further notice in the future. But as we distribute the package as Python wheels anyways, it shouldn't impact our customers negatively to drop 3.6 support.

### PR Type:


jspeerless (approved):
LGTM! Did a search and couldn't find other mention of our python versions supported.

kroenlein (approved):
Looks like which ever users were using 3.6 before have updated their deployments.

FYI, we had the specific python versions in the travis file because there were some support problems with previous versions of the containers.  I guess those have been resolved.

kroenlein (changes requested):
Please review the version numbers in requirements.txt and test_requirements.txt to bring them in line with the dropped support.

anoto-moniz:
> Please review the version numbers in requirements.txt and test_requirements.txt to bring them in line with the dropped support.

All of them work properly as is. If the minimum version of pandas is used, it will build from source, but that (and the rest) work properly. And since it's defined as a minimum version, users can ensure a higher version is installed if they're concerned about the time.

kroenlein (approved):
Thanks for verifying

PR #785, opened 2022-10-25
[PLA-9883] Expose predictor versioning.

anoto-moniz:
Adds support to the predictor collection for performing operations on specific predictor versions. Also automatically uses a new feature to cut a version upon successful predictor training.

This is split into three commits.
- One drops the patch we added to preserve the user's understanding of what archive means, since the ability to separately archive roots and versions confuses that. With the release of versions, this is no longer a concern.
- One adds support for versions to the predictor collection.
- One add support for predictor versions to collections which hit related endpoints.

### PR Type:


anoto-moniz:
Moved this back to a draft, as I need to add versions support to some other parts of the SDK. I'll set it back to ready and request new reviews once that's up.

tests/resources/test_predictor.py
```diff
@@ -79,46 +79,40 @@ def test_archive(valid_label_fractions_predictor_data):
...
         archived_predictor = pc.archive(pred_id)
 
     assert session.calls == [
         FakeCall(method='PUT', path=f"{predictors_path}/{pred_id}/archive", json={}),
-        FakeCall(method='GET', path=f"{predictors_path}/{pred_id}"),
-        FakeCall(method='GET', path=predictors_path, params={"filter": "archived eq 'true'", 'per_page': 20})
+        FakeCall(method='GET', path=f"{predictors_path}/{pred_id}")
     ]
 
-    assert archived_predictor.is_archived
```
genfx999, line 93: Was removing this intentional?

anoto-moniz:
Note we're going to hold off merging until UI is ready to release, which won't be for a bit. As such, I'm leaving it in draft, so if I need to make relevant updates/changes for them, you'll still be viewing it all at once.

genfx999 (approved):
A question about a removed assertion, but otherwise LGTM.  Also should let Ken get a look before merging.

tests/resources/test_predictor.py
anoto-moniz: Yes. Currently, `archived_predictor` is just the payload we told it to respond with on line 82. Previously, the value returned by `is_archived` would be impacted by processing on the response of a subsequent call, in order to present a value the user expected before they were exposed to versions.

genfx999 (dismissed):

genfx999 (approved):

PR #789, opened 2022-11-14
use data to created default design space dimension

StevenKauwe:
# Citrine Python PR

## Description 
DesignSpaceCollection.create_default() recently gained the ability to [optionally] automatically constrain the formulation-subspace of the returned ProductDesignSpace to the support of the reference predictor’s training data. This PR extends the current logic to add defaults for process parameters (https://citrine.atlassian.net/browse/PLA-9990).

### PR Type:


src/citrine/resources/design_space.py
```diff
@@ -69,7 +69,8 @@ def create_default(self,
                        predictor_id: UUID,
                        include_ingredient_fraction_constraints: bool = False,
                        include_label_fraction_constraints: bool = False,
-                       include_label_count_constraints: bool = False) -> DesignSpace:
+                       include_label_count_constraints: bool = False,
+                       include_parameter_constraints: bool = False) -> DesignSpace:
```
kroenlein, line 73: Why are we using `False` here instead of `None`?  Seems like a painful way to make later migration annoying.

kroenlein (approved):
Generally looks good.  You are consistent with previous behavior, but I would have used `None` as the default.  There also aren't tests of default behavior.

PR #791, opened 2022-11-23
[PLA-10556] Add support for status detail.

anoto-moniz:
These structured messages will eventually be used throughout the Orion part of the API, but for now are only applied to predictors and design spaces.

### PR Type:


src/citrine/_rest/ai_resource_metadata.py
```diff
@@ -29,11 +29,8 @@ class AIResourceMetadata():
 
     status = properties.Optional(properties.String(), 'status', serializable=False)
     """:Optional[str]: short description of the resource's status"""
-    status_info = properties.Optional(
-        properties.List(properties.String()),
-        'status_info',
-        serializable=False
-    )
+    status_info = properties.Optional(properties.List(properties.String()), 'status_info',
+                                      serializable=False)
```
kroenlein, line 33: How'd this change happen?  Some kind of auto-linting?

src/citrine/informatics/design_spaces/design_space.py
```diff
@@ -21,6 +24,27 @@ class DesignSpace(Module):
...
+        # TODO: This property should live in AIResourceMetadata. However, since not all modules
+        # which inherit from that class support status_detail, it must live here until they do.
+        """[DEPRECATED] :list[str]: human-readable explanations of the status."""
+        return [detail.msg for detail in self.status_detail]
+
+    @status_info.setter
+    def status_info(self, value):
+        # This is required to allow deserialization to work, since it tries to treat status_info as
+        # a property, as defined in AIResourceMetadata.
+        pass  # pragme: no cover
```
kroenlein, line 46:
```suggestion
        pass  # pragma: no cover
```

kroenlein, line 46: Why are we not skipping this implementation, or throwing a `RuntimeError`?  A no-op setter seems like bad mojo.

src/citrine/informatics/processors.py
```diff
@@ -22,6 +25,27 @@ class Processor(Module):
...
+        # TODO: This property should live in AIResourceMetadata. However, since not all modules
+        # which inherit from that class support status_detail, it must live here until they do.
+        """[DEPRECATED] :list[str]: human-readable explanations of the status."""
+        return [detail.msg for detail in self.status_detail]
+
+    @status_info.setter
+    def status_info(self, value):
+        # This is required to allow deserialization to work, since it tries to treat status_info as
+        # a property, as defined in AIResourceMetadata.
+        pass  # pragme: no cover
```
kroenlein, line 47: As above

src/citrine/informatics/design_spaces/design_space.py
anoto-moniz: Interesting! The linter doesn't alert on it anyways. I wonder why...

anoto-moniz: Backwards compatibility.

This was previously a property, which as a Python field, users were welcome to write to if they wanted. But as it had `serializable=False`, it wouldn't be sent to the backend. So of course, users _shouldn't_ write to it, but as it was previously allowed, we shouldn't break that capability.

Granted, you could make the argument we're changing its behavior. I wanted to do the least possible to not break the interface. But if you feel it's important for this behavior to stay the same, I can rig something up to make it work same as before if they set it.

kroenlein: Makes sense.  Maybe we should be raising a Warning, perhaps a DeprecationWarning?  That would at least be noisy.

src/citrine/informatics/design_spaces/design_space.py
```diff
@@ -21,6 +24,27 @@ class DesignSpace(Module):
...
+    @deprecated(deprecated_in="1.45.0", removed_in="2.0.0",
+                details="Please use the 'status_detail' property instead.'")
+    def status_info(self):
+        # TODO: This property should live in AIResourceMetadata. However, since not all modules
+        # which inherit from that class support status_detail, it must live here until they do.
+        """[DEPRECATED] :list[str]: human-readable explanations of the status."""
+        return [detail.msg for detail in self.status_detail]
+
+    @status_info.setter
+    def status_info(self, value):
```
kroenlein, line 43: Does this need a ` @deprecated` annotation as well?  I'm not certain on the scoping of that decorator.

src/citrine/informatics/processors.py
```diff
@@ -22,6 +25,27 @@ class Processor(Module):
...
+    @deprecated(deprecated_in="1.45.0", removed_in="2.0.0",
+                details="Please use the 'status_detail' property instead.'")
+    def status_info(self):
+        # TODO: This property should live in AIResourceMetadata. However, since not all modules
+        # which inherit from that class support status_detail, it must live here until they do.
+        """[DEPRECATED] :list[str]: human-readable explanations of the status."""
+        return [detail.msg for detail in self.status_detail]
+
+    @status_info.setter
+    def status_info(self, value):
```
kroenlein, line 44: Does this need a ` @deprecated` annotation as well?  I'm not certain on the scoping of that decorator.

src/citrine/resources/status_detail.py
```diff
@@ -0,0 +1,35 @@
...
+    INFO = "Info"
+    WARNING = "Warning"
+    ERROR = "Error"
+    UNKOWN = "Unknown"
+
+
+Self = TypeVar('Self', bound='StatusDetail')
+
+
+class StatusDetail(Serializable[Self]):
```
kroenlein, line 21: I don't see this `Self` pattern anywhere else in the library. Does it have a risk of namespace collision? Not super familiar with the implementation details of this chunk of our serialization stack. Closest matches look like they would call this the `StatusDetailType`
https://github.com/CitrineInformatics/citrine-python/blob/13e8eb1f123222863903f627ea82442e59c224a1/src/citrine/resources/module.py#L37

```suggestion
StatusDetailType = TypeVar('StatusDetailType', bound='StatusDetail')


class StatusDetail(Serializable[StatusDetailType]):
```

kroenlein (commented):
Some follow on comments.  Nothing blocking.

src/citrine/_rest/ai_resource_metadata.py
anoto-moniz: No, I probably did it when an earlier iteration made some changes to this file. Although maybe it was yelling at me, because I don't know why I'd mess with the formatting with no other change?

src/citrine/informatics/design_spaces/design_space.py
anoto-moniz: Agreed, it definitely should.

src/citrine/resources/status_detail.py
anoto-moniz: I copied it from src/citrine/_rest/resource.py, but I think your way makes more sense.

anoto-moniz:
@kroenlein When I went to add the deprecated annotation to the field, it ended up being always raised on deserialization, since design spaces inherit from a class which still needs to support `status_info`. Untangling that proved to be too complicated for now. Instead, I elected to add the `status_detail` field directly to that class, meaning it will appear on some AI Engine resources for which it will always be empty. That will change in the near future, as there's ongoing work to expose it on all of Orion's modules, at which time they will automatically start picking it up. And once they all do, we can deprecate `status_info` for all of them at once, eliminating this problem.

kroenlein (dismissed):
Sorry for opening a can of worms.  This looks solid now.

kroenlein (approved):
LGTM

kroenlein (approved):
Still LGTM

PR #802, opened 2023-01-09
Update docs about Data Model

kroenlein:
# Citrine Python PR
https://citrine.atlassian.net/browse/PLA-10695

## Description 
The primary goal of this PR is to update the citrine-python documentation to reflect how to update templates specifically and reflecting many QoL changes that have happened in code in general.

This PR also migrates `async_update` from ALPHA, which necessitates a minor bump.  It also removes duplicative code for `async_update` from the `GEMDResourceCollection`.

There should be no change in functionality.

### PR Type:


docs/source/formulations_example.rst
```diff
@@ -142,9 +142,7 @@ Make sure to set the bounds wide enough to encompass all anticipated use cases o
 
 .. code-block:: python
 
-    from citrine.resources.parameter_template import ParameterTemplate
-    from citrine.resources.property_template import PropertyTemplate
-    from gemd.entity.bounds import RealBounds
+    from gemd import ParameterTemplate, PropertyTemplate, RealBounds
```
lbryan-citrine, line 145: Why is this change the recommendation?

docs/source/data_entry.rst
```diff
@@ -39,11 +41,29 @@ For example:
 
     dataset.process_specs.register(ProcessSpec(...))
 
+Equivalent behavior is available through the type-agnostic ``gemd`` collection and directly from a dataset object:
```
lbryan-citrine, line 44: 👍

docs/source/formulations_example.rst
kroenlein: Because there is no additional functionality made available by importing the citrine-python objects and importing from the gemd-python base package significantly reduces the number of characters.  When someone is just learning the package, they shouldn't need to keep track all this complexity / package structure.

docs/source/data_entry.rst
```diff
@@ -99,10 +126,14 @@ The measurements of a material can be located with
...
 ---------------------------
-Runs and specs can be quickly modified in-place and persisted with ``upload``, but templates require more care.
+Runs and specs can be quickly modified in-place and persisted with ``register`` or ``register_all``, but templates require more care.
 In particular, changing the bounds or allowed names/labels of a template could invalidate existing data objects; thus every object on platform must be compared against the desired change.
+If there is no risk that an update could invalidate data (e.g., changing an object name), the template can be updated as per runs and specs.
+
+If such a risk exists (e.g., making bounds more restrictive), ``register`` and ``register_all`` will raise exceptions.
 To attempt such a template update, use :meth:`~citrine.resources.data_concepts.DataConceptsCollection.async_update`.
 If the update is invalid, then the reasons for failure are logged.
+As a convenience, the ``update`` method will first try ``register`` and then will fall-back to ``async_update`` with convenient arguments.
```
lbryan-citrine, line 136: It's unclear to me if `async_update` needs to be included in the high-level documentation at all, do the default arguments cover most use cases? Maybe adding a separate docs block for cases where things like timeout configuration would be more helpful.

lbryan-citrine (dismissed):
This is really helpful clarification, with a few callouts, thanks Ken!

docs/source/data_entry.rst
kroenlein: That's a fair critique, and something I'd considered.  It fell in there because it felt like I should include `sometimes register will work` and then felt like I should explain what `update` was doing.  Would you prefer I just say `templates can only be updated using update` or leave the `register` v. `update` distinction?

lbryan-citrine: I would bias towards recommending `update` for all use cases, not just templates, since the `register` vs `async_update` distinction is essentially an implementation detail.

kroenlein: It is, but `register_all` is a serious QoL improvement.  Orders of magnitude speedup in processing objects.  Understanding when you can use it seems important.

docs/source/formulations_example.rst
lbryan-citrine: I don't like the `gemd` imports for data objects because it breaks the mental model of "the `citrine` wraps the `gemd` lower-level api". My preference in general would be to create a more opinionated top-level package import structure for `citrine`, even if data objects are functional interchangeable between libraries. Happy to defer to your preference here though.

docs/source/data_entry.rst
lbryan-citrine: Ah, ok that is important. What about adding an equivalent `update_all` wrapper?

kroenlein: That'd be good, but would require adding a route to the API in order to really get the benefit.  

We could put in a method that filtered the templates from the runs/specs, but then you have a challenge of knowing which to execute first; if the template bounds were increased and the values also were broader, we need to load templates first; if the template bounds shrank and extreme values were dropped from the objects, we need to load the objects first.  Client would require knowledge of everything on platform and some business logic in order to draw those conclusions.

docs/source/formulations_example.rst
kroenlein: That's a good critique.  I'd really like to see citrine-python ergonomics improved generally.  The number of import statements working with citrine-python requires can be really painful.

In this case importing some the GEMD model from citrine-python and some from gemd-python seems worse to me than losing that hierarchical packaging perspective.  But I'll sleep on it.

docs/source/data_entry.rst
lbryan-citrine: Alright, given that context your existing documentation makes sense, thanks Ken.

jspeerless (commented):
I just have one (potentially dumb) question re importing directly from gemd: if one only installs citrine-python via pip, will the direct `gemd` imports still work? Or will a user also have to install gemd-python in their environment to perform imports in this way?

kroenlein:
> I just have one (potentially dumb) question re importing directly from gemd: if one only installs citrine-python via pip, will the direct `gemd` imports still work? Or will a user also have to install gemd-python in their environment to perform imports in this way?

`pip install citrine` installs gemd-python as well since it's a dependency.  Since it's installed in your local environment, import with  work even though you didn't install it directly.

jspeerless:
> > I just have one (potentially dumb) question re importing directly from gemd: if one only installs citrine-python via pip, will the direct `gemd` imports still work? Or will a user also have to install gemd-python in their environment to perform imports in this way?
> 
> `pip install citrine` installs gemd-python as well since it's a dependency. Since it's installed in your local environment, import with work even though you didn't install it directly.

Great! Looks good to me then!

jspeerless (approved):
Looks great!

lbryan-citrine (approved):
👍 re-approved

PR #804, opened 2023-01-25
Add functionality for generative design api

StevenKauwe:
# Citrine Python PR

## Description 

### PR Type:


StevenKauwe:
Taking a first guess at what citrine python code needs to be added (and where) for generative molecular design support.

src/citrine/informatics/executions/generative_design_execution.py
```diff
@@ -0,0 +1,90 @@
+# from deprecation import deprecated
```
bfolie, line 1: ?

src/citrine/resources/generative_design_execution.py
```diff
@@ -0,0 +1,112 @@
...
+        self.session: Session = session
+
+    def build(self, data: dict) -> GenerativeDesignExecution:
+        """Build an individual GenerativeDesignExecution."""
+        execution = GenerativeDesignExecution.build(data)  # this is inherited [Serialization]
+        execution._session = self.session
+        execution.project_id = self.project_id
+        return execution
+
+    # TODO: The execution_input should be a DesignExecutionInput(?), not a Score
```
bfolie, line 37: Yes, you'll need to create a Python class that holds the settings used to kick off the generative execution

bfolie:
This looks more or less correct. One thing I see missing is that you'll need a way for users to access the `GenerativeDesignExecutionCollection`. I expect that this would be through the project, so `Project` would have a property `def generative_design_executions` which returns a `GenerativeDesignExecutionCollection`. This allows a user who has selected a `project` to run `project.generative_design_executions.trigger(...)` to kick off the generative design

src/citrine/informatics/executions/generative_design_execution.py
StevenKauwe: Hahaha woops, I meant to delete that. I copied this structure from DesignExecution then updated it to match code found here: https://github.com/CitrineInformatics/platform-backend/blob/generative-design-task/code/orion/ml-config-api/src/main/scala/io/citrine/orion/api/endpoints/GenerativeDesignApi.scala#L69

src/citrine/informatics/executions/generative_design_execution.py
```diff
@@ -0,0 +1,111 @@
...
+
+        fetcher = partial(self._fetch_page, path=path, fetch_func=self._session.get_resource)
+
+        return self._paginator.paginate(page_fetcher=fetcher,
+                                        collection_builder=self._build_results,
+                                        page=page,
+                                        per_page=per_page)
+
+
+    # def result(
```
anoto-moniz, line 97: Combining what I see here with what you said in slack, I think you're trying to get a single result, but from an endpoint that isn't just `_path_template`.

You'll need to call `session.get_resource` directly. See the example in [report.py](https://github.com/CitrineInformatics/citrine-python/blob/32403f6139f1787223dc9fe7effd66d62f72a1a0/src/citrine/resources/report.py#L45).

src/citrine/resources/generative_design_execution.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def register(self, model: GenerativeDesignExecution) -> GenerativeDesignExecution:
+        """Cannot register an execution."""
+        raise NotImplementedError("Cannot register a GenerativeDesignExecution.")
+
+    def update(self, model: GenerativeDesignExecution) -> GenerativeDesignExecution:
+        """Cannot update an execution."""
+        raise NotImplementedError("Cannot update a GenerativeDesignExecution.")
+
+    def archive(self, uid: Union[UUID, str]):
```
anoto-moniz, line 55: `archive` and `restore` are not defined on `Collection`, so there's no need to overwrite them.

src/citrine/informatics/executions/generative_design_execution.py
```diff
@@ -0,0 +1,111 @@
...
+        )
+
+    @classmethod
+    def _build_results(
+        cls, subset_collection: Iterable[dict]
+    ) -> Iterable[GenerationExecutionResult]:
+        for generation_result in subset_collection:
+            yield GenerationExecutionResult.build(generation_result)
+
+    def results(
```
anoto-moniz, line 79: If you have a request method on a data class, it should be connected directly to that result. So `results` here only makes sense if there can be multiple results, and the listed results are only those connected to this execution.

src/citrine/informatics/executions/generative_design_execution.py
anoto-moniz: Similar to above, this `result` method should be sure to only return a result if the given ID is associated with the execution in which it's contained.

src/citrine/resources/generative_design_execution.py
```diff
@@ -0,0 +1,114 @@
...
+        Parameters
+        ----------
+        uid: UUID
+            Unique identifier of the execution to restore
+
+        """
+        raise NotImplementedError(
+            "Generative Design Executions cannot be restored")
+
+    def list(self, *,
```
anoto-moniz, line 79: This endpoint doesn't exist for generative design. The only way to list is with an execution ID, so this method needs to accept one.

tests/conftest.py
```diff
@@ -787,6 +787,31 @@ def design_execution_dict(generic_entity):
...
+@pytest.fixture
+def example_generation_results():
+    return {
+        "page": 2,
+        "per_page": 4,
+        "response": [{
+            "id": str(uuid.uuid4()),
+            "execution_id": str(uuid.uuid4()),
+            "result": {
+                "seed": str,
```
anoto-moniz, line 805: Found your test problem. These need to be values, not classes.

StevenKauwe: OMG lol. :D

src/citrine/resources/generative_design_execution.py
anoto-moniz: Nevermind. Somehow missed this in the API. My bad.

src/citrine/informatics/design_candidate.py
```diff
@@ -131,3 +131,63 @@ class DesignCandidate(Serializable["DesignCandidate"]):
...
+    fingerprint_similarity = properties.Float("fingerprint_similarity")
+    """The fingerprint similarity between the seed and the mutated molecule."""
+    fingerprint_type = properties.String("fingerprint_type")
+    """The fingerprint type used to calculate the fingerprint similarity."""
+
+    def __init__(self):
+        pass  # pragma: no cover
+
+
+class GenerationExecutionResult(Serializable["GenerationExecutionResult"]):
```
sfriedowitz, line 155: We could possibly combine this class and the above class into one for the sake of the SDK. It might be a bit simpler to have less nested items. The combined class could be a called a `GenerativeDesignResult` which contains the meta-data fields and the result fields.

src/citrine/informatics/design_candidate.py
```diff
@@ -131,3 +131,63 @@ class DesignCandidate(Serializable["DesignCandidate"]):
...
+
+    uid = properties.UUID('id')
+    execution_id = properties.UUID('execution_id')
+    result = properties.Object(GenerationResult, 'result')
+
+    def __init__(self):
+        pass  # pragma: no cover
+
+
+class GenerativeDesignExecutionInput(Serializable['GenerativeDesignExecutionInput']):
```
sfriedowitz, line 169: Suggest renaming this to `GenerativeDesignInput` to save a bit of wordiness.

src/citrine/resources/generative_design_execution.py
```diff
@@ -0,0 +1,114 @@
...
+shadow_classes_in_module(
+    citrine.informatics.executions.generative_design_execution, sys.modules[__name__]
+)
+
+
+# TODO: Double check the path.
+class GenerativeDesignExecutionCollection(Collection["GenerativeDesignExecution"]):
+    """A collection of GenerativeDesignExecutions."""
+
+    _path_template = '/projects/{project_id}/generative-design/executions'
```
sfriedowitz, line 24: I'm not sure if this is the correct path template? The base path for the API is `"api/v1/generative-design/executions"`, so I'm not sure if we need the project path explicitly. But @anoto-moniz  can confirm this.

sfriedowitz: Actually I thnk I am wrong here. i think the public URI has the `projects` ID in the path, and only the proxy is shown above from backend.

StevenKauwe: This was weird. When I had the "v1" in front of the expected path, my assertion was False. When I removed it (how you see) things worked again.

src/citrine/informatics/design_candidate.py
StevenKauwe: I created these two different classes because this matches the return type from backend. Maybe we can talk about how to combine.

sfriedowitz: It should be straightforward. The combined class will look something like

```
class GenerativeDesignResult():
    uid: UUID
    execution_id: UUID
    seed: str
    mutated: str
    ....
```

StevenKauwe: When the api is working, I'll play around with this!

src/citrine/jobs/waiting.py
```diff
@@ -126,14 +127,19 @@ def wait_while_validating(
                                         interval=interval)
 
 
+# Might need to add GenerativeDesignExecution here.
```
sfriedowitz, line 130: You could perhaps derive from a common base class like ExecutionCollection? Then you don't need an explicit Union type.

anoto-moniz: I wouldn't worry about it just for the purposes of type annotation.

sfriedowitz: Yeah agree, can just plop the new type it into the Union annotation there

src/citrine/informatics/generative_result.py
```diff
@@ -0,0 +1,57 @@
+from typing import List
```
sfriedowitz, line 1: I suggest we rename this file to `generative_design.py` since it contains generally data classes for this end point, not just the result.

src/citrine/informatics/executions/generative_design_execution.py
```diff
@@ -0,0 +1,114 @@
...
+from citrine._session import Session
+from citrine._utils.functions import format_escaped_url
+from citrine.informatics.generative_design import GenerationExecutionResult
+
+
+# TODO: This is not pretty, but I needed something to hold the status
+# When we build the GenerativeDesignExecution.
+# This and GenerationResult at `generative_design.py` are the same situation.
+# Is there a way to avoid this?
+class GenerativeExecutionStatus:
```
StevenKauwe, line 20: @anoto-moniz This part of the code is iffy. The status I get from the api is a dict {"minor": str, "info": list(str)}. So Without this "status object" I wasn't sure how to correctly build the GenerativeDesignExecution. Happy to figure out what the solution is here.

anoto-moniz: With Sean's status change yesterday, you'll want to update this to match the other executions.

src/citrine/resources/generative_design_execution.py
anoto-moniz: If you want your test to check the path version, then you need to pass the `version` parameter to the `FakeCall` constructor. This way, if the caller doesn't care about the version, they're not forced to specify it in the path.

We really need to document the `FakeSession` and `FakeCall` classes...

src/citrine/informatics/executions/generative_design_execution.py
anoto-moniz: And once you do, it should fix the broken tests.

StevenKauwe: Excelling This and a couple other small things on the way!

src/citrine/informatics/executions/generative_design_execution.py
bfolie: Still need to remove

tests/resources/test_generative_design_execution.py
```diff
@@ -0,0 +1,133 @@
...
+    build_data["id"] = str(generative_design_execution_id)
+
+    # When
+    execution: GenerativeDesignExecution = collection.build(build_data)
+
+    # Then
+    assert execution.uid == generative_design_execution_id
+    assert execution.project_id == collection.project_id
+    assert execution._session == collection.session
+    assert execution.in_progress() and not execution.succeeded() and not execution.failed()
```
bfolie, line 52: Why generate a new uid to assign to `build_data`? Why not just take the uid that comes with `generative_design_execution_dict` and assert that that uid makes it on to the `GenerativeDesignExecution` object?

src/citrine/informatics/executions/generative_design_execution.py
```diff
@@ -0,0 +1,101 @@
...
+                                        collection_builder=self._build_results,
+                                        page=page,
+                                        per_page=per_page)
+
+    def result(
+        self,
+        *,
+        result_id: UUID,
+    ) -> GenerativeDesignResult:
+        """Fetch the Design Candidates for the particular execution, paginated."""
```
sfriedowitz, line 98: This docstring is incorrect.

src/citrine/informatics/generative_design.py
```diff
@@ -0,0 +1,83 @@
+from typing import List, TypeVar
+from citrine._serialization import properties
+from citrine._serialization.serializable import Serializable
+from gemd.enumeration.base_enumeration import BaseEnumeration
+
+
+Self = TypeVar('Self', bound='Serializable')
```
sfriedowitz, line 7: What's going on here with the type var? is this necessary?

src/citrine/resources/generative_design_execution.py
```diff
@@ -0,0 +1,91 @@
...
+from citrine._rest.collection import Collection
+from citrine._utils.functions import shadow_classes_in_module
+from citrine._session import Session
+import citrine.informatics.executions.generative_design_execution
+from citrine.informatics.executions.generative_design_execution import GenerativeDesignExecution
+from citrine.informatics.generative_design import GenerativeDesignInput
+from citrine.resources.response import Response
+
+
+shadow_classes_in_module(
```
sfriedowitz, line 15: Is this necessary?

tests/resources/test_generative_design_execution.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def test_trigger_execution(collection: GenerativeDesignExecutionCollection, generative_design_execution_dict, session):
+    # Given
+    session.set_response(generative_design_execution_dict)
+    design_execuption_input = GenerativeDesignInput(
+        seeds = ["CC(O)=O"],
+        fingerprint_type = FingerprintType.ECFP4,
+        min_fingerprint_similarity = 0.5,
+        mutation_per_seed = 2
```
bfolie, line 62: We don't enforce style in the tests directory, but there should not be spaces around the equals sign

tests/resources/test_generative_design_execution.py
bfolie: For example, the following test passes and I believe tests the same thing:
```
    execution: GenerativeDesignExecution = collection.build(generative_design_execution_dict)

    assert str(execution.uid) == generative_design_execution_dict["id"]
    assert execution.project_id == collection.project_id
    assert execution._session == collection.session
    assert execution.in_progress() and not execution.succeeded() and not execution.failed()
```

tests/resources/test_generative_design_execution.py
```diff
@@ -0,0 +1,133 @@
...
+    expected_path = '/projects/{}/generative-design/executions/{}/results'.format(
+        generative_design_execution.project_id,
+        generative_design_execution.uid,
+    )
+    assert session.last_call == FakeCall(method='GET', path=expected_path, params={"per_page": 4})
+
+
+def test_generative_design_execution_result(generative_design_execution: GenerativeDesignExecution, session, example_generation_results):
+    # Given
+    session.set_response(example_generation_results)
```
bfolie, line 102: It strikes me as odd that `example_generation_results` is being used as the expected response for testing both the `results` and `result` function. The former returns a paginated list of results whereas the latter returns a single generative design result. So don't they have different json responses?

tests/resources/test_generative_design_execution.py
StevenKauwe: I was trying to use the pattern found here. I'm not sure if there is a "gotcha" they were avoiding that we don't need to worry about when that workflow_id is included. I happy making it simpler by removing the `generative_design_execution_id = uuid.uuid4()` though.

https://github.com/CitrineInformatics/citrine-python/blob/main/tests/resources/test_predictor_evaluation_executions.py#L47

tests/resources/test_generative_design_execution.py
StevenKauwe: Ah. I am grabbing the `id` from the first result in the `example_generation_results`. Thats why it is used in both locations.

bfolie: But why is it the response in both cases? That would imply that both `/results` and `/result/{id}` return json of the same shape. Do they?

StevenKauwe: I see. Let me look at the backend API to double check what it returns.

src/citrine/informatics/executions/generative_design_execution.py
sfriedowitz: @StevenKauwe this comment stil needs to be addressed

src/citrine/informatics/generative_design.py
```diff
@@ -0,0 +1,83 @@
...
+__all__ = ['FingerprintType']
+
+
+class FingerprintType(BaseEnumeration):
+    """[ALPHA] Fingerprint type used in Generative Design.
```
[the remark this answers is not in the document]
StevenKauwe: For now I'm okay not including this information and letting the names be a "good enough" starting point if they wanna go down that rabbit hole. But I will fix the text a bit. :D

bfolie (commented):
I ran everything against the development environment and it's all working correctly, so I think that this is just about done. But I see one lingering issue -- there's no documentation of the arguments in `GenerativeDesignInput`. This is especially important because they don't seem to behave intuitively. I ran design using two seeds and with `mutation_per_seed = 15`. I would have expected this to result in 30 results, but I only got 6 (3 from each seed). What's the expected behavior here?

bfolie (approved):

PR #806, opened 2023-02-08
remove processors

bfolie:
# Citrine Python PR

## Description 
Removes the entire concept of processors. They are no longer persistent assets and can not be specified as part of a design workflow. Instead, the backend selects a default.

Adds the optional argument `max_candidates` to DW execution triggering, in case users want to run a short DW execution (which they previously did by defining a processor with a limited number of max candidates).

### PR Type:


src/citrine/informatics/executions/design_execution.py
```diff
@@ -64,6 +64,8 @@ class DesignExecution(Resource['DesignExecution'], Pageable, AsynchronousObject)
 
     score = properties.Object(Score, 'score')
     """:Score: score by which this execution was evaluated"""
+    max_candidates = properties.Optional(properties.Integer, 'max_candidates')
```
anoto-moniz, line 67: I don't think we need to store this as part of the design execution unless users ask for it. Part of my hesitation is it would require us to persist it on the backend, which feels like overkill for an execution parameter which just limits the number of results.

tests/utils/fakes/fake_execution_collection.py
```diff
@@ -9,9 +10,10 @@
 
 class FakeDesignExecutionCollection(DesignExecutionCollection):
 
-    def trigger(self, execution_input: Score) -> DesignExecution:
+    def trigger(self, execution_input: Score, max_candidates: Optional[int] = None) -> DesignExecution:
         execution = DesignExecution()
         execution.score = execution_input
+        execution.max_candidates = max_candidates
```
anoto-moniz, line 16: To remove `max_candidates` from the response, it would also be removed from here.

src/citrine/informatics/executions/design_execution.py
bfolie: Agreed; I'll take it out

PR #807, opened 2023-02-09
 Remove v2 deprecations

bfolie:
# Citrine Python PR

## Description 
Remove all deprecated code that is slated for EoL by version 2.

### PR Type:


src/citrine/_rest/engine_resource.py
```diff
@@ -55,38 +54,6 @@ def is_archived(self):
...
-    def status_info(self):
-        """[DEPRECATED] :list[str]: human-readable explanations of the status."""
-        return [detail.msg for detail in self.status_detail]
-
-    @property
-    @deprecated(deprecated_in="1.31.0", removed_in="2.0.0",
-                details="Please use the 'is_archived' property instead.'")
-    def archived(self):
-        """[DEPRECATED] whether the resource is archived."""
-        return self.is_archived
```
bfolie, line 70: Note that design spaces, which are modules and not engine resources, still use `.archived`

src/citrine/_rest/paginator.py
```diff
@@ -61,12 +55,8 @@ def paginate(self,
         # making 'search_params' key of outermost dict for keyword expansion by page_fetcher func
         search_params = {} if search_params is None else {'search_params': search_params}
 
-        if page is not None:
-            warnings.warn("The page parameter is deprecated, default is automatic pagination",
-                          DeprecationWarning)
-
         first_entity = None
-        page_idx = page
+        page_idx = 1
```
bfolie, line 59: Always start with page 1 and retrieve everything

src/citrine/_session.py
```diff
@@ -35,23 +34,13 @@ class Session(requests.Session):
 
     def __init__(self,
                  refresh_token: str = None,
-                 legacy_scheme: Optional[str] = None,
+                 scheme: str = None,
                  host: str = None,
-                 port: Optional[str] = None,
-                 *,
-                 scheme: str = None):
+                 port: Optional[str] = None):
```
bfolie, line 39: I'm tempted to enforce keyword args here, but worry about breaking scripts that start with `Citrine(api_key, scheme, host, port)`

tests/resources/test_predictor.py
```diff
@@ -70,127 +65,6 @@ def test_delete():
...
-        archived_predictor = pc.archive(uuid.uuid4())
-    assert archived_predictor.is_archived
-
-    del entity["metadata"]["archived"]
-    session.set_responses(None, entity, paging_response())
-    
-    with pytest.deprecated_call():
-        restored_predictor = pc.restore(uuid.uuid4())
-
-    assert not restored_predictor.is_archived
```
bfolie, line 90: I'm confused as to how archiving works now and how it should be tested. There are some lines in the `update` method of `resources/predictor.py` that are no longer covered by tests and pertain to the `._archived` property.

bfolie:
@anoto-moniz the only thing missing is some coverage of the `update` method, in particular around archiving and automatic post-update training.

src/citrine/_session.py
anoto-moniz: Given that using it positionally already emitted a warning (see below), we're fine to enforce it as a keyword.

anoto-moniz: Host should be the same thing, as indicated by that deprecation warning.

src/citrine/citrine.py
```diff
@@ -28,22 +28,12 @@ class Citrine:
 
     def __init__(self,
                  api_key: str = None,
-                 legacy_scheme: Optional[str] = None,
+                 scheme: str = None,
```
anoto-moniz, line 31: Same as in the session.

src/citrine/informatics/predictors/simple_ml_predictor.py
```diff
@@ -1,69 +0,0 @@
-from typing import List, Optional
```
anoto-moniz, line 1: Sweeeeeeeeeeeeeeeeeeeet.

src/citrine/resources/dataset.py
```diff
@@ -276,7 +275,7 @@ def delete_contents(
             *,
             timeout: float = 2 * 60,
             polling_delay: float = 1.0,
-            prompt_to_confirm: bool = None,
+            prompt_to_confirm: bool = True,
```
anoto-moniz, line 278: Given that it previously defaulted to `None`, would it make more sense to make it default to `False` now? Or will users usually want to be prompted?

src/citrine/resources/report.py
```diff
@@ -29,13 +26,10 @@ def __init__(self, project_id: UUID, session: Session):
         self.session = session
 
     def get(self,
-            module_id: UUID = None,
-            *,
             predictor_id: Union[UUID, str] = None,
```
anoto-moniz, line 29: I think `predictor_id` should be kept keyword-only. It might not be obvious to a caller that first argument to `get` isn't the report ID, but forcing them to name it as such should do so.

src/citrine/resources/predictor.py
```diff
@@ -397,115 +367,3 @@ def create_default(self,
                 "prefer_valid": prefer_valid}
         data = self.session.post_resource(path, json=body, version=self._api_version)
         return self.build(Predictor.wrap_instance(data["instance"]))
-
-    def convert_to_graph(self,
```
anoto-moniz, line 401: I'm not sure we want to remove access to this endpoint. We still have customers with unarchived simple ML predictors in unarchived design workflows. I count about 150 such predictors across 5 customers.

So while the SDK can drop support for simple ML predictors, we should still allow them to convert for the time being.

tests/resources/test_predictor.py
anoto-moniz: The `_archived` property was introduced for backwards compatibility in archiving. Instead of setting `archived` on the predictor and updating it, users can now call the `archive`/`restore` methods. Since setting `archived` emitted a deprecation warning, we can now remove it (as you already did), along with the `_archived` property, and no longer do any archiving in the `update` method.

anoto-moniz:
This is great! Thanks so much for the effort here. This major version bump is gonna make maintaining the SDK so much nicer.

Just a few comments. Lemme know if you wanna discuss.

src/citrine/resources/dataset.py
bfolie: Now that I think about it, the warning says that you will soon need to call the method with `prompt_to_confirm` explicitly set. So perhaps there shouldn't be a default

src/citrine/resources/design_workflow.py
```diff
@@ -43,15 +44,37 @@ def register(self, model: DesignWorkflow) -> DesignWorkflow:
...
-        model_copy = deepcopy(model)
-        model_copy.branch_id = self.branch_id
-        return super().register(model_copy)
+        # Importing locally to avoid circular dependency
+        from citrine.resources.branch import Branch, BranchCollection
+
+        if self.branch_id is None:
+            # There are a number of contexts in which hitting design workflow endpoints without a
+            # branch ID is valid, so only this particular usage is deprecated.
+            msg = ('Creating a design workflow without a branch is deprecated as of 1.19.0 and '
```
anoto-moniz, line 53: When you merged main, this was pulled in. Given the deprecation warning, we don't need to allow users to create a design workflow without a branch anymore, although we could.

anoto-moniz (approved):

PR #817, opened 2023-02-17
Bugfix/Score serialization

bfolie:
# Citrine Python PR

## Description 
Although setting the name and description is deprecated, we still need those fields on the backend. That should be fixed, but for now silently set defaults.

### PR Type:

### DIFF
--- a/src/citrine/informatics/scores.py
+++ b/src/citrine/informatics/scores.py
@@ -17,6 +17,9 @@ class Score(PolymorphicSerializable['Score']):
 
     """
 
+    _name = properties.String('name')
+    _description = properties.String('description')
+
     @classmethod
     def get_type(cls, data):
         """Return the subtype."""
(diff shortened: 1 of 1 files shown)


src/citrine/informatics/scores.py
```diff
@@ -17,6 +17,9 @@ class Score(PolymorphicSerializable['Score']):
 
     """
 
+    _name = properties.String('name')
+    _description = properties.String('description')
```
anoto-moniz, line 21: If they're only needed for deserialization, we can mark them as `serializable=False` so we won't even send anything to the backend when POSTed.

anoto-moniz (approved):

src/citrine/informatics/scores.py
bfolie: They're only needed for serialization, so I marked `deserializable=False`

PR #826, opened 2023-03-08
Allow user to update predictor without retraining, default case is to retrain

genfx999:
# Citrine Python PR

## Description 
From original support ticket, a request from DS to allow changing name of predictor (from API) without automatically retraining.  The training was initiated from the predictorCollection.update in all cases, this update allows training to be omitted if desired.

### PR Type:


anoto-moniz:
Do we actually want this behavior? From the linked ticket, I don't know that we do. And if so, why wouldn't we also add that option to `register`?

anoto-moniz (changes requested):
Do we actually want this behavior? From the linked ticket, I don't know that we do. And if so, why wouldn't we also add that option to `register`?

PR #827, opened 2023-03-08
[PLA-9323] Add ingredient ratio constraint.

anoto-moniz:
This constraint is used on a design space, and allows the user to constrain an ingredient or label based on its ratio to others.

### PR Type:


src/citrine/informatics/constraints/ingredient_ratio_constraint.py
```diff
@@ -0,0 +1,107 @@
...
+        """Set the label and its multiplier in the numerator."""
+        self._labels = self._numerator_validate(value, "Label")
+
+    def _numerator_read(self, num_dict):
+        if num_dict:
+            return tuple(num_dict.items())[0]
+        else:
+            return None
+
+    def _numerator_validate(self, value, input_type):
```
sfriedowitz, line 93: The backend is also doing this validation logic, correct? I guess it's good to do it early, but perhaps not strictly necessary to have it client-side this things will eventually fall out of sync.

sfriedowitz: _If_ we want to do more client-side validation, we may also want to validate that the basis is (1) has > 0 entries and (2) the basis multipliers are positive.

sfriedowitz (commented):
one comment, but looks pretty good

src/citrine/informatics/constraints/ingredient_ratio_constraint.py
anoto-moniz: Only sort of. The main purpose is to ensure they only provide a single key-value pair, since that's what we want to support in the client right now. Since there's no datatype in Python to let me restrict the length of a sequence (that I know of), I enforce it here.

I added the multiplier piece while I was here, but it could be removed to let the backend respond if we really want to. I'd just rather be strict in the client and relax those constraints later rather than the other way around.

sfriedowitz: Sounds good, I won't complain about it.

src/citrine/__version__.py
```diff
@@ -1 +1 @@
-__version__ = '2.3.0'
+__version__ = '2.4.0'
```
sfriedowitz, line 1: Thimnk you need a new verison bump

sfriedowitz (approved):

PR #830, opened 2023-03-14
PLA-12095: Deprecated FormulationDescriptor on table data source

sfriedowitz:
# Citrine Python PR

`formulation_descriptor` is no longer a field on a `GemTableDataSource`. Deprecate its usage, update various docs, and make package-wide test updates to use the only two valid keys for formulations.

Its very possible I missed some doc updates, but this is a definite improvement.

### PR Type:


src/citrine/informatics/descriptors.py
```diff
@@ -252,3 +253,11 @@ def __str__(self):
 
     def __repr__(self):
         return "FormulationDescriptor(key={})".format(self.key)
+
+    @staticmethod
+    def hierarchical() -> "FormulationDescriptor":
```
sfriedowitz, line 258: I'm not sure what we want to call this for the SDK?

scripts/run_pytest.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env sh
```
sfriedowitz, line 1: I find this is helpful for my local development

src/citrine/informatics/descriptors.py
bfolie: @jspeerless thoughts on user-facing language?

src/citrine/informatics/descriptors.py
```diff
@@ -252,3 +253,13 @@ def __str__(self):
...
+        """The hierarchical formulation descriptor with key 'Formulation'."""
+        return FormulationDescriptor("Formulation")
+
+    @staticmethod
+    def flat() -> "FormulationDescriptor":
```
bfolie, line 263: Is this (using static methods to create specific instantiations of a class) Python idiomatic? It feels so much like Scala that it makes me wary. Maybe I'm over-thinking it

docs/source/workflows/descriptors.rst
```diff
@@ -62,9 +62,18 @@ The Molecular Structure Descriptor has no parameters other than a name.
...
-:class:`~citrine.informatics.descriptors.FormulationDescriptor` is used to represent variables that contain information
-about mixtures of other materials.
-The Formulation Descriptor has no parameters other than a name.
+:class:`~citrine.informatics.descriptors.FormulationDescriptor`
+is used to represent variables that contain information about mixtures of other materials.
+The Formulation Descriptor has no parameters other than a name,
+of which the two allowed values are 'Formulation' and 'Flat Formulation'.
+The key 'Formulation' should be used when referring to mixtures found directly in the training data,
+for which the ingredients may be other mixtures themselves.
+The key 'Flat Formulation' should be used when referring to mixtures containing only raw ingredients,
```
bfolie, line 71: In a one-level mixing problem the mixture only contains raw ingredients, but we still want to use "Flat Formulation"

src/citrine/informatics/descriptors.py
sfriedowitz: Looks like the googles says we should use a `classmethod` for alternate constructors: https://realpython.com/python-multiple-constructors/#defining-multiple-class-constructors:~:text=Remove%20ads-,Providing%20Multiple%20Constructors%20With%20%40classmethod%20in%20Python,-A%20powerful%20technique

src/citrine/informatics/descriptors.py
sfriedowitz: James said he likes "hierarchical" here

tests/informatics/test_predictors.py
```diff
@@ -343,7 +343,7 @@ def test_auto_ml_multiple_outputs(auto_ml_multiple_outputs):
 def test_ing_to_formulation_initialization(ing_to_formulation_predictor):
     """Make sure the correct fields go to the correct places for an ingredients to formulation predictor."""
     assert ing_to_formulation_predictor.name == 'Ingredients to formulation predictor'
-    assert ing_to_formulation_predictor.output.key == 'formulation'
+    assert ing_to_formulation_predictor.output.key == 'Formulation'
```
kroenlein, line 346: Can this set of tests happen with a constant / enum instead of a hard-coded string?

kroenlein, line 346: Can this be a constant / enum instead of having that string appear several times in tests?

scripts/run_pytest.sh
kroenlein: From a formal perspective, this should be a second PR. In this case, I won't block on this.

Why are you putting pytest in the script but not flake8, derp, or validate-version-bump.sh?

Is the directory modification necessary? The test dir should probably be tests not ..

My current git hook that could probably use an update:
```
if [ "`git rev-parse --abbrev-ref HEAD`" == "main" ];
     then echo "On main branch";
     exit 1;
fi                                 &&
scripts/validate-version-bump.sh   &&
flake8 src                         &&
pytest --quiet --cov=src --cov-report term-missing:skip-covered   \
  --cov-config=tox.ini --no-cov-on-fail --cov-fail-under=100 -x tests
```

kroenlein (dismissed):
Primary work looks fine, though I didn't dig through to make sure you have 100 % doc coverage.

A couple of non-blocking questions.

scripts/run_pytest.sh
sfriedowitz: >Why are you putting pytest in the script but not flake8, derp, or validate-version-bump.sh?

Because it is quite easy to remember the command `flake8 src`, and harder to type out the pytest command with specific coverage flags.

> Is the directory modification necessary

Yes, so the script can be invoked regardless of where your terminal is currently located.

> The test dir should probably be tests not ..

Copied directly from the previous `.travis.yml` file. We can change to `tests` if desired.

>My current git hook that could probably use an update:

Probably good to update that to call this script directly after this PR 🙃

tests/informatics/test_predictors.py
sfriedowitz: Sure, I can make an enum of the two allowed keys.

scripts/run_pytest.sh
sfriedowitz: I do see a bug in my directory modification, which I will fix now.

sfriedowitz: Fixed in this revision.

src/citrine/informatics/descriptors.py
```diff
@@ -235,20 +250,36 @@ class FormulationDescriptor(Serializable['FormulationDescriptor'], Descriptor):
...
     """
 
-    typ = properties.String('type', default='Formulation', deserializable=False)
+    typ = properties.String(
+        'type', default=FormulationKey.HIERARCHICAL.value, deserializable=False
+    )
+
+    def __init__(self, key: Union[FormulationKey, str]):
+        if not isinstance(key, FormulationKey):
+            key = FormulationKey.get_enum(key)
```
bfolie, line 265: This will throw an error if `key` is not one of the two options, yes?

sfriedowitz: Yep

sfriedowitz: I think we are already forcing them down a path of failed assets by using other keys so that additional client-side validation is fine.

bfolie: fair enough

bfolie (approved):

PR #835, opened 2023-03-21
[PLA-12207] Add BranchCollection.get_by_root_id.

anoto-moniz:
Allows the caller to get the latest version of a branch by passing the root branch ID. This allows a user to go between the UI and SDK more easily when necessary.

### PR Type:


src/citrine/resources/branch.py
```diff
@@ -100,6 +101,33 @@ def build(self, data: dict) -> Branch:
...
+
+        Returns
+        -------
+        Branch
+            The latest version of the branch.
+
+        """
+        params = {"root": str(branch_root_id), "version": "latest"}
+        branch = next(self._list_with_params(per_page=1, **params), None)
+        if branch:
```
kroenlein, line 121:
```suggestion
        if branch is None:
```
Probably equivalent in practice, but explicit None comparison is preferred in style guides.

tests/resources/test_branch.py
```diff
@@ -406,3 +407,35 @@ def test_experiment_data_source_no_project_id(session):
...
+    root_id = uuid.uuid4()
+
+    # When
+    branch = collection.get_by_root_id(branch_root_id=root_id)
+
+    # Then
+    assert session.calls == [FakeCall(
+        method='GET',
+        path=branch_path,
+        params={'page': 1, 'per_page': 1, 'root': str(root_id), 'version': 'latest'}
```
kroenlein, line 425: Should this just be testing for `root` and `version` rather than the full call structure?  I feel like those are the only bits that actually matter, and `page` and `per_page` are internal implementation details.

src/citrine/exceptions.py
```diff
@@ -82,7 +84,25 @@ def __init__(self, path: str, response: Optional[Response] = None):
...
 
-    pass
+    @staticmethod
+    def build(*, message: str, method: str, path: str, params: dict = {}):
+        """
+        Construct a NotFound response from parameters.
+
+        This allows raising NotFound locally to produce consistent behavior even if the server
+        doesn't return a 404.
+        """
```
kroenlein, line 94: Public method doesn't document it's arguments / return value

kroenlein (approved):
Minor, non-blocking style concerns.

src/citrine/resources/branch.py
anoto-moniz: Use of truth values is idiomatic Python, and I don't see PEP 8 taking a stance one way or another about their use.

tests/resources/test_branch.py
anoto-moniz: Only `root` and `version` are relevant, yes. But the `FakeCall` construct compares the whole payload, not just present keys, so they must be included. And since this is a unit test, including internal details isn't a problem (although it should of course be avoided when possible).

kroenlein (approved):
Approving following previous non-blocking approval.

kroenlein (approved):

PR #840, opened 2023-03-31
Make path construction & escaping central

kroenlein:
# Citrine Python PR

## Description 
This moves some longer URL construction logic from the file_link.py into collection.py, and then refactors the methods that I could find to take advantage.  There is still some URL construction happening in the various resource files, but that'd take another refactoring push because those are either a) in Resources, not Collections or b) changing something early in the Collection path.

[Jenkins job](https://jenkins.corp.citrine.io/job/nextgen-ITs/36618/)

### PR Type:


src/citrine/_rest/collection.py
```diff
@@ -39,15 +40,37 @@ def _put_resource_ref(self, subpath: str, uid: Union[UUID, str]):
...
         else:
-            return format_escaped_url(self._path_template + subpath,
-                                      **self.__dict__)
+            base = urlparse(self._path_template)
+        path = base.path.split('/')
+
+        if uid is not None:
+            path.append("{uid}")
+
+        if action is None:
```
anoto-moniz, line 57: If the `action` parameter defaults to `[]`, then this first clause wouldn't be necessary.

src/citrine/_rest/collection.py
```diff
@@ -1,6 +1,7 @@
 from abc import abstractmethod
 from logging import getLogger
-from typing import Optional, Union, Generic, TypeVar, Iterable, Iterator
+from typing import Optional, Union, Generic, TypeVar, Iterable, Iterator, Sequence, Dict
```
anoto-moniz, line 3: `typing.Sequence` [is deprecated](https://docs.python.org/3/library/typing.html#typing.Sequence). Use `collections.abc.Sequence` instead.

src/citrine/_rest/collection.py
```diff
@@ -39,15 +40,37 @@ def _put_resource_ref(self, subpath: str, uid: Union[UUID, str]):
...
+
+        if action is None:
+            action = []
+        elif isinstance(action, str):
+            action = [action]
+        else:
+            action = list(action)
+        path.extend(["{}"] * len(action))
+
+        if query_terms is None:
```
anoto-moniz, line 65: Same as `action` (except a dict, of course).

src/citrine/_rest/collection.py
```diff
@@ -39,15 +40,37 @@ def _put_resource_ref(self, subpath: str, uid: Union[UUID, str]):
...
+            query_terms = {}
+        query_args = [x for k_and_v in query_terms.items() for x in k_and_v]
```
anoto-moniz, line 67: Instead of this and your join below, you can use `urllib.parse.urlencode`: [https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode)

src/citrine/_rest/collection.py
```diff
@@ -39,15 +40,37 @@ def _put_resource_ref(self, subpath: str, uid: Union[UUID, str]):
...
-                                      **self.__dict__)
+            base = urlparse(self._dataset_agnostic_path_template)
         else:
-            return format_escaped_url(self._path_template + subpath,
-                                      **self.__dict__)
+            base = urlparse(self._path_template)
+        path = base.path.split('/')
+
+        if uid is not None:
+            path.append("{uid}")
```
anoto-moniz, line 55: What's the reason for inserting a format string and having `format_escaped_url` populate it, rather than just appending the value inline here?

anoto-moniz: Ah, nevermind, it's because of how you construct `path`. Makes sense.

src/citrine/_rest/pageable.py
```diff
@@ -1,5 +1,5 @@
 from logging import getLogger
-from typing import Optional, Iterable, Dict, Tuple, Callable, Union
+from typing import Optional, Iterable, Dict, Tuple, Callable, Union, Sequence
```
anoto-moniz, line 2: See comment above.

src/citrine/_rest/collection.py
kroenlein: Yes, thanks!

kroenlein: It's deprecated as of 3.9, but we support 3.7.

src/citrine/_rest/collection.py
kroenlein: Done.

src/citrine/_rest/collection.py
kroenlein: Makes sense, though it'd be slightly better if I moved that into a argument to `format_escaped_url` -- we're mixing responsibilities.

Done.

anoto-moniz (approved):

PR #843, opened 2023-04-12
[PLA-12391] Support retraining stale predictors.

anoto-moniz:
Our releases sometimes mean the platform cannot load trained predictors. We call these predictors stale. The predictor configuration is still in tact and usable, so our solution is to silently retrain it when it's used in a design execution.

The modify candidates feature allows users to use an old, trained predictor directly. As such, it also sometimes requires retraining, but outside the context of a design execution. As such, this exposes a route to check if a predictor is stale, and to retrain such a predictor.

### PR Type:


genfx999 (approved):

src/citrine/resources/predictor.py
```diff
@@ -147,6 +153,22 @@ def convert_to_graph(self,
...
+        path = self._construct_path(uid, version, "is-stale")
+        response = self.session.get_resource(path, version=self._api_version)
+        return response["is_stale"]
+
+    def retrain_stale(self,
+                      uid: Union[UUID, str],
+                      *,
+                      version: Union[int, str] = MOST_RECENT_VER) -> Predictor:
+        path = self._construct_path(uid, version, "retrain-stale")
+        entity = self.session.put_resource(path, {}, version=self._api_version)
```
kroenlein, line 169: It's weird that we have to manually put in the API version.  Seems like we should have cleaned this up to defaultly use the `self._api_version`.

kroenlein (approved):
Looks good.  Hope the path clean-ups helped.

PR #844, opened 2023-04-17
PLA-12433: Only graph predictor usages on platform

sfriedowitz:
# Citrine Python PR

## Description 
Platform only supports graph predictors. Eliminating usages of other predictor types in the context of a graph. Defines a new `PredictorNode` hierarchy for the individual components nested within a graph.

ITs running off this branch: https://jenkins.corp.citrine.io/job/nextgen-ITs/37203/

### PR Type:


src/citrine/informatics/predictors/node.py
```diff
@@ -0,0 +1,61 @@
...
+        warnings.warn(
+            f"The field `training_data` on single predictor nodes is deprecated "
+            "and will be removed in version 3.0.0. Include training data for all "
+            "sub-predictors on the parent GraphPredictor. Existing training data "
+            "on this predictor will be moved to the parent GraphPredictor upon registration.",
+            DeprecationWarning
+        )
+
+
+class PredictorNode(PolymorphicSerializable["PredictorNode"]):
```
sfriedowitz, line 20: It is not possible for the node types to acquire meta-data/version information because they cannot be registered standalone. As such, I have removed those mix-in classes from the inheritance chain which would indicate a break in the fields present on the types. 

They can be added back for the sake of backwards compat if we deem necessary.

src/citrine/informatics/predictors/predictor.py
```diff
@@ -1,125 +0,0 @@
...
-from citrine.informatics.data_sources import DataSource
-from citrine.resources.report import ReportResource
-from citrine.informatics.predictors.single_predict_request import SinglePredictRequest
-from citrine.informatics.predictors.single_prediction import SinglePrediction
-from citrine._utils.functions import format_escaped_url
-
-__all__ = ['Predictor']
-
-
-class Predictor(PolymorphicSerializable['Predictor'], AsynchronousObject):
```
sfriedowitz, line 18: The `GraphPredictor` assumes all fields from this interface, since there is only one type.

We can change the inheritance chain back to maintain backwards compat if we deem. necessary.

src/citrine/informatics/predictors/graph_predictor.py
```diff
@@ -1,25 +1,35 @@
...
+    The GraphPredictor is the only predictor that can be registered on the Citrine Platform
+    and carries along its meta-data regarding versioning, platform identifiers, and updates.
 
     Parameters
     ----------
     name: str
         name of the configuration
     description: str
         the description of the predictor
-    predictors: List[Union[UUID, Predictor]]
```
sfriedowitz, line 21: Removing the `UUID` signature for graphs, which is technically a break.

tests/resources/test_descriptors.py
```diff
@@ -50,6 +50,21 @@ def test_from_predictor_responses():
...
+    graph_results = descriptors.from_predictor_responses(predictor=graph, inputs=[MolecularStructureDescriptor(col)])
+    assert graph_results == [
+        RealDescriptor(
+            key=r['descriptor_key'],
+            lower_bound=r['lower_bound'],
+            upper_bound=r['upper_bound'],
+            units=r['units']
+        ) for r in response_json['responses']
+    ]
+
```
bfolie, line 67: What new logic is this testing?

tests/serialization/test_predictors.py
```diff
@@ -49,39 +50,45 @@ def test_graph_serialization(valid_graph_predictor_data):
...
 
 def test_auto_ml_serialization(valid_auto_ml_predictor_data):
     """"Ensure that a serialized AutoMLPredictor looks sane."""
-    predictor_serialization_check(valid_auto_ml_predictor_data, AutoMLPredictor)
+    predictor_node_serialization_check(valid_auto_ml_predictor_data, AutoMLPredictor)
 
 
-def test_invalid_predictor_type(invalid_predictor_data):
+def test_invalid_graph(invalid_graph_predictor_data):
+    """Ensures we raise proper exception when horribly malformed data is used."""
```
bfolie, line 86:
```suggestion
    """Ensures we raise proper exception when malformed data is used."""
```

I hope that it doesn't need to be horribly malformed in order to raise an exception :p

src/citrine/informatics/predictors/graph_predictor.py
bfolie: Where did we come down on the discussion of minor vs. major version bump?

src/citrine/informatics/predictors/graph_predictor.py
```diff
@@ -31,40 +41,75 @@ class GraphPredictor(VersionedEngineResource['GraphPredictor'], Predictor):
...
+            project_id=self._project_id,
+            predictor_id=str(self.uid),
+            version=self.version
+        )
+
+    def predict(self, predict_request: SinglePredictRequest) -> SinglePrediction:
+        """Make a one-off prediction with this predictor."""
+        path = self._path() + '/predict'
+        res = self._session.post_resource(path, predict_request.dump(), version=self._api_version)
+        return SinglePrediction.build(res)
```
bfolie, line 115: Is this new or was it moved from somewhere else?

src/citrine/informatics/predictors/graph_predictor.py
sfriedowitz: We didn't land on anything.

I suggested y'all have a look at this draft, which is the "ideal" state in terms of changes. And see if we can walk any of these changes back until we feel comfortable doing this under a minor instead of a major.

src/citrine/informatics/predictors/graph_predictor.py
sfriedowitz: This is all copied from the `Predictor` super class, which has been absorbed into the `GraphPredictor` class.

tests/resources/test_descriptors.py
sfriedowitz: The fact that the `from_predictor_responses` endpoint accepts a `Union[PredictorNode, GraphPredictor]` type, and has a conditional checking how to extract the JSON payload.

src/citrine/informatics/predictors/graph_predictor.py
bfolie: I really doubt anyone is doing this, but we could keep this as `List[Union[UUID, Predictor]]`, and if there are any UUIDs filter them out and emit a deprecation warning

src/citrine/informatics/predictors/node.py
bfolie: These are all read-only fields, right? So the potential break is that someone could have a script in which they request, say the `updated_by` field of a predictor node, and that field is no longer available. Is that correct?

If it's impossible to register non-graph predictor nodes, then none of those fields are meaningful anyway. We'd want to pull the associated value from the graph predictor. Not that that's easy to do...

sfriedowitz: >If it's impossible to register non-graph predictor nodes, then none of those fields are meaningful anyway

These fields are only populated when you register a predictor. They are generated by the platform upon DB entry. Given that all predictor nodes are contained within the Graph, I do not think it is possible for these fields to be populated on the node classes.

So for example, lets say a user creates a local `AutoMLPredictor` and tries to access the `updated_by` field. Before these changes, that would result in a value of `None` being returned. After these changes, it would become `'AutoMLPredictor' object has no attribute 'updated_by'`

I suppose one option is to just leave the inheritance of those types and the fields can never be populated since the platform will never fill them in.

src/citrine/informatics/predictors/graph_predictor.py
sfriedowitz: That works. I can make that change.

src/citrine/informatics/predictors/node.py
bfolie: Let's just leave the inheritance. It maintains strict backwards compatibility and is very easy to remove when we go to version 3

sfriedowitz: The inheritance fucks with the serialization paths for the data, so this is quite tedious to figure out.

src/citrine/informatics/predictors/predictor.py
```diff
@@ -1,125 +1,11 @@
...
 
 
-class Predictor(PolymorphicSerializable['Predictor'], AsynchronousObject):
-    """Module that describes the ability to compute/predict properties of materials.
-
-    Abstract type that returns the proper type given a serialized dict. Subtype
-    based on the 'type' value of the passed in dict.
-
-    """
+class Predictor(AsynchronousObject):
```
sfriedowitz, line 6: I added back this type as pretty much a dummy object, in case anyone is importing it within their scripts. The majority of fields have been moved over to Graph Predictor.

src/citrine/informatics/predictors/node.py
```diff
@@ -0,0 +1,177 @@
...
+        typ = type_dict.get(data['type'])
+        if typ is not None:
+            return typ
+        else:
+            raise ValueError(
+                '{} is not a valid predictor node type. '
+                'Must be in {}.'.format(data['type'], type_dict.keys())
+            )
+
+    @property
```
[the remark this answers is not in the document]
bfolie: That'll do for now. Hopefully we'll get rid of this before too long

bfolie:
Did you test out the e2e tests using this branch? Given the magnitude of the changes it seems prudent

sfriedowitz:
I've ran a few. Here is a new one after my most recent changes: https://jenkins.corp.citrine.io/job/nextgen-ITs/37261/

bfolie (approved):

PR #845, opened 2023-04-19
[PLA-12466] Fix ingredient ratio constraint interface.

anoto-moniz:
To match the UI, users should not be able to provide multipliers for the basis labels or ingredients. The client still needs to provide them (as 1) to the backend, though.

### PR Type:


jamie-heller (dismissed):
LGTM!

anoto-moniz:
> I don't know this interface well enough to review. Can you link the API docs? It looks like your calls have less data in them than there used to be.

Unfortunately, I can't link to the docs, because the design space portion of the API just accepts a JSON blob. And this impacts the handling of a specific constraint type.

The backend API hasn't changed, though. The intention is to further restrict the clients in order to keep data in line with the way we're going to present this feature in the UI. If by my calls having less data you mean that the user's "basis" multipliers are ignored, then yes, that's correct.

genfx999 (approved):

PR #846, opened 2023-04-20
[PLA-12483] Predictor archive/restore are not implmented.

anoto-moniz:
Predictor versioning and the separation between archiving/restoring versions and roots meant the standard archive/restore should no longer be used. They were deprecated, but when we bumped to 2.0, they wer removed, which uncovered the base collection archive/restore, despite them no longer working as expected.

So instead, we explicitly raise a NotImplementedError with a descriptive error message.

### PR Type:


kroenlein (dismissed):
I'm a little torn on the philosophical question of is this a change in API.  But thing that threw and exception before throws an exception, so... 🤷

kroenlein (approved):

PR #850, opened 2023-04-28
Rework relationship between citrine-python & gemd-python's serialization layers

kroenlein:
# Citrine Python PR

## Description 
This PR creates a cleaner surface separating gemd-python & citrine-python's serialization behaviors.

* It updates the GEMD-derived objects to use the implicit typ system introduced in gemd-python 1.13.0
* It makes the MRO consistently choose serialization, then citrine-python objects, then gemd-objects
* It updates the _serialization tools to respect mro (method resolution order) and to search anywhere in the class structure for a method instead of just one layer deep.
* It adds an `use_init` boolean to Property objects to flag those fields that require an `__init__` routine to be run for proper initialization.
* It corrects a number of what looked like previously uncovered bugs.

### PR Type:

- [X]  I have communicated the downstream consequences of the PR to others.


src/citrine/_rest/resource.py
```diff
@@ -46,6 +48,49 @@ def access_control_dict(self) -> dict:
         }
 
 
+GEMDSelf = TypeVar('GEMDSelf', bound='GEMDResource')
+
+
+class GEMDResource(Resource[GEMDSelf]):
```
kroenlein, line 54: All GEMD objects / DictSerializables go through this one Resource to maintain a consistent interface layer.

src/citrine/_rest/resource.py
```diff
@@ -46,6 +48,49 @@ def access_control_dict(self) -> dict:
...
+            key = next(k for k in data if k != "context")
+            idx = make_index([_inflate(x) for x in data["context"] + [data[key]]])
+            lst = [idx[k] for k in idx]
+            substitute_objects(lst, idx, inplace=True)
+
+            root = _inflate(data[key])
+            if root in idx:  # It was a link
+                return idx[root]
+            else:  # It was an object, but it won't be densely linked
+                return idx[root.to_link()]
```
kroenlein, line 72: Support Material History-like serializations.  Handles both the return structure from the `material_history` route as well as the GEMD denormalized format.

src/citrine/_rest/resource.py
```diff
@@ -46,6 +48,49 @@ def access_control_dict(self) -> dict:
...
+        else:
+            if data.get("type") is not None:
+                if not issubclass(cls, DictSerializable.class_mapping.get(data.get("type"))):
+                    raise ValueError(f"{cls.__name__} passed a {data.get('type')} dictionary.")
```
kroenlein, line 76: This is new behavior, and I'm open to dropping it.  It essentially says that `ProcessSpec.build` expects that you're actually passing it a `ProcessSpec`, as opposed to e.g. a `MaterialRun`.  The alternative is to propagate Polymorphic behavior for build to all concrete GEMD objects.

src/citrine/_serialization/properties.py
```diff
@@ -22,22 +22,50 @@
...
 
     def __init__(self,
                  serialization_path: typing.Optional[str] = None,
+                 *,
                  serializable: bool = True,
                  deserializable: bool = True,
                  default: typing.Optional[DeserializedType] = None,
-                 override: bool = False
+                 override: bool = False,
+                 use_init: bool = False
```
kroenlein, line 56: New parameter.  When de-serializing an object, use klass(args) instead of __new__.  Necessary for cases where, for example, a private field is set to None in the __init__ routine.

src/citrine/_serialization/properties.py
```diff
@@ -22,22 +22,50 @@
...
                  ):
         self.serialization_path = serialization_path
-        self._key: str = '__' + str(uuid.uuid4())  # Make this object key human readable
+        if override:
+            self._key: None = None
+        else:
+            self._key: str = '__' + str(uuid.uuid4())  # Make this object key human-readable
```
kroenlein, line 62: With the properties clean up, `override` became redundant with the None-ish value of `_key`.

src/citrine/_serialization/properties.py
```diff
@@ -101,40 +135,54 @@ def deserialize_from_dict(self, data: dict) -> DeserializedType:
...
-        base_class = _get_base_class(obj, self.serialization_path)
+        property_name, base_class = _get_key_and_base_class(self, type(obj))
         if issubclass(type(value), self.underlying_types):
             value_to_set = value
         else:
             # if value is not an underlying type, set its deserialized version.
             value_to_set = self.deserialize(value, base_class=base_class)
 
-        if base_class is not None and self.override:
-            getattr(base_class, self.serialization_path).fset(obj, value_to_set)
```
kroenlein, line 137: This was a latent bug.  `serialization_path` happened to equal all property_names for the cases where something was overridden, but that was definitely not guaranteed.

src/citrine/_serialization/properties.py
```diff
@@ -175,22 +229,21 @@ def _set_elements(self, value: typing.Union[SerializedType, DeserializedType]):
         """
 
 
-def _get_base_class(obj: object, key: str) -> type:
+@lru_cache(maxsize=1024 * 1024)
```
kroenlein, line 232: This caching drops test execution share from 28 % to 6 %.

src/citrine/_serialization/properties.py
```diff
@@ -175,22 +229,21 @@ def _set_elements(self, value: typing.Union[SerializedType, DeserializedType]):
...
+def _get_key_and_base_class(prop: Property, klass: typing.Any) -> \
+        typing.Tuple[typing.Optional[str], typing.Optional[str]]:
```
kroenlein, line 234: Previously, we were incorrectly searching for serialization_path.  This first finds the actual field key and then resolves the MRO base class (instead of demanding that the property exist along exactly one inheritance path).

src/citrine/_serialization/properties.py
```diff
@@ -612,28 +686,37 @@ def _serialize(self, value: typing.Any) -> str:
...
 
 
+@lru_cache(maxsize=1024 * 1024)
+def _fields_map(klass: typing.Type) -> typing.Dict[str, Property]:
+    """Compute the properties-relevant fields in a given class."""
+    return {
+        k: v
+        for x in reversed(klass.__mro__)  # Classes at the front trump
+        for k, v in x.__dict__.items() if isinstance(v, Property)
+    }
```
kroenlein, line 696: Uses a reverse MRO instead of piecemeal pulling things in from two most immediate layers.  Allows for proper method inheritance instead of just a shallow inheritance.

src/citrine/_serialization/properties.py
```diff
@@ -612,28 +686,37 @@ def _serialize(self, value: typing.Any) -> str:
         return self.klass.get_value(value)
 
 
+@lru_cache(maxsize=1024 * 1024)
```
kroenlein, line 689: Caching drops this routine's testing share from 30 % to 7.5 %.

src/citrine/_serialization/properties.py
```diff
@@ -716,16 +819,27 @@ class LinkOrElse(PropertyCollection[typing.Union[Serializable, LinkByUID], dict]
...
+                 use_init: bool = False
+                 ):
+        super().__init__(
+            serialization_path=serialization_path,
+            serializable=serializable,
+            deserializable=deserializable,
+            default=default,
+            override=override,
+            use_init=use_init)
+        self.klass = klass
```
kroenlein, line 838: Modify LinkOrElse to actually keep track of what Else means.

src/citrine/_serialization/properties.py
```diff
@@ -738,12 +852,9 @@ def _serialize(self, value: typing.Any) -> dict:
...
     def _deserialize(self, value: dict):
-        if 'type' in value and value['type'] == LinkByUID.typ:
-            if 'scope' in value and 'id' in value:
-                value.pop('type')
-                return LinkByUID(**value)
-            else:
-                raise ValueError("LinkByUID dictionary must have both scope and id fields")
+        if 'type' in value:
+            target = DictSerializable.class_mapping[value['type']]
+            return target.build(value)
```
kroenlein, line 857: No need for LinkOrElse to know how to decode a LinkByUID.

src/citrine/resources/api_error.py
```diff
@@ -4,7 +4,7 @@
 from gemd.entity.dict_serializable import DictSerializable
 
 
-class ValidationError(DictSerializable):
+class ValidationError(DictSerializable, typ="validation_error"):
```
kroenlein, line 7: DictSerializable is being modified to make objects w/o typ abstract according to ABC.  There is no type field in the object that comes in.  Might be reasonable to just migrate this to a Serializable[] object.

src/citrine/resources/api_error.py
```diff
@@ -15,7 +15,7 @@ def __init__(self, failure_message: Optional[str] = None, property: Optional[str
         self.failure_id = failure_id
 
 
-class ApiError(DictSerializable):
+class ApiError(DictSerializable, typ="api_error"):
```
kroenlein, line 18: DictSerializable is being modified to make objects w/o typ abstract according to ABC.  There is no type field in the object that comes in.  Might be reasonable to just migrate this to a Serializable[] object.

tests/_serialization/_data.py
```diff
@@ -86,6 +86,5 @@ def _deserialize(self, value):
     (properties.Boolean, 1.0),
     (properties.Boolean, 'asdf'),
     (properties.UUID, 'wrong-number-of-chars'),
-    (properties.UUID, uuid.uuid4()),
```
kroenlein, line 89: Modified to be valid, since deserialization routines aren't bothered by previously deserialized values anymore.

tests/_serialization/test_container_properties.py
```diff
@@ -65,12 +65,12 @@ def test_mapping_property_list_of_pairs_multiple():
...
-    dummy_set = properties.Set(type(properties.Float()))
-    link_or_else = properties.LinkOrElse()
-    map_collection_key = properties.Mapping(properties.Optional(properties.String), properties.Integer)
-    specified_mixed_list = properties.SpecifiedMixedList([properties.Integer(default=100)])
+    dummy_map = properties.Mapping(properties.Float(), properties.String, "dummy_map")
+    dummy_list = properties.List(properties.Float, "dummy_list")
+    dummy_set = properties.Set(type(properties.Float()), "dummy_map")
+    link_or_else = properties.LinkOrElse(serialization_path="link_or_else")
+    map_collection_key = properties.Mapping(properties.Optional(properties.String), properties.Integer, "map_collection_key")
+    specified_mixed_list = properties.SpecifiedMixedList([properties.Integer(default=100)], "specified_mixed_list")
```
kroenlein, line 73: serialization_path changed to required argument for top-level properties.  Doesn't make sense to have a top-level property without it.

src/citrine/resources/data_concepts.py
```diff
@@ -27,7 +29,23 @@
...
+class DataConceptsMeta(DictSerializableMeta):
+    """Data Concepts metaclass for handling serialization."""
+
+    def __init__(cls, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        resolved = next((b.typ for b in cls.__bases__ if getattr(b, "typ", None) is not None),
+                        None)
+        if resolved is not None:
+            cls._typ_stash = resolved
+        cls.typ = String("type")
```
kroenlein, line 41: We have to do weird things here because GEMD.typ is a class-value, rather than an object value, property, or classmethod.  Ideally, it'd be a classmethod+property, but Python 3.12 dropped support for that concept.

src/citrine/resources/data_concepts.py
```diff
@@ -27,7 +29,23 @@
...
 
 
-class DataConcepts(DictSerializable, PolymorphicSerializable['DataConcepts'], ABC):
+class DataConceptsMeta(DictSerializableMeta):
+    """Data Concepts metaclass for handling serialization."""
+
+    def __init__(cls, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        resolved = next((b.typ for b in cls.__bases__ if getattr(b, "typ", None) is not None),
+                        None)
```
kroenlein, line 38: Once we can import a gemd-python version that has abstract + concrete classes, this can be simplified.  That is expected to be released next cycle.

src/citrine/_rest/resource.py
```diff
@@ -46,6 +48,49 @@ def access_control_dict(self) -> dict:
...
+GEMDSelf = TypeVar('GEMDSelf', bound='GEMDResource')
+
+
+class GEMDResource(Resource[GEMDSelf]):
+    """A reference to a resource by UID."""
+
+    @classmethod
+    def build(cls, data: dict) -> GEMDSelf:
+        """Convert a raw, nested dictionary into Objects."""
+        if "context" in data and len(data) == 2:
```
bfolie, line 60: Which situations does this apply to?

src/citrine/_serialization/properties.py
```diff
@@ -22,22 +22,50 @@
...
+    serialization_path: str
+        The location in the serialized structure where the value of interest can be
+        found.  The value can be nested in the structure, where separate keys are
+        joined by `.`.  Required in a top level property, but not necessary if
+        a property is wrapped by a container.
+    serializable: bool
+        Include the field when serializing an object to a dictionary; default: True
+    deserializable: bool
+        Look for the field when deserializing a dictionary into an object; default: True
+    override: bool
```
bfolie, line 42: Missing docstring for `default`

src/citrine/_serialization/properties.py
```diff
@@ -52,26 +80,32 @@ def serialized_types(self) -> typing.Union[SerializedType, typing.Tuple[Serializ
...
+                f'{self.underlying_types}{base_name}'
+            )
         return self._serialize(value)
 
     def deserialize(self, value: SerializedType,
-                    base_class: Optional[type] = None) -> DeserializedType:
+                    base_class: typing.Optional[type] = None) -> DeserializedType:
         if not isinstance(value, self.serialized_types):
+            if isinstance(value, self.underlying_types):
+                return value  # Don't worry if it was already deserialized
```
bfolie, line 103: In what situation are we trying to deserialize an already-deserialized object?

src/citrine/_rest/resource.py
kroenlein: https://github.com/CitrineInformatics/gemd-python/blob/b625aee87d02768d918ccb4088a97b43fa15c894/gemd/json/gemd_json.py#L32
gemd-python makes a material history with a `object` key for the root object of the serialized graph
https://github.com/CitrineInformatics/citrine-python/blob/9be3cded48ea64f754d1897513100fd2230d0c52/src/citrine/resources/material_run.py#L141
platform returns a material history with a `root` key for a LinkByUID to the root object.

src/citrine/resources/api_error.py
bfolie: It's a smell to have this citrine-python Error object inheriting its type from GEMD

src/citrine/_serialization/properties.py
kroenlein: The `flatten` utility from gemd-python only partly serializes its objects when dissecting them, the validate_templates method in citrine-python does something similar, and many tests use build combined with Objects in the incoming values stream.

Maybe not ideal, but correcting this would make this PR substantially larger.  Not a new behavior.

src/citrine/resources/api_error.py
kroenlein: Probably better to rebuild from scratch.  This seemed like the highest-velocity solution.

src/citrine/_serialization/properties.py
```diff
@@ -101,40 +135,54 @@ def deserialize_from_dict(self, data: dict) -> DeserializedType:
...
                 value = next_value
-        base_class = _get_base_class(data, self.serialization_path)
-        return self.deserialize(value, base_class=base_class)
+        return self.deserialize(value, base_class=None)  # always a dict
 
     def serialize_to_dict(self, data: dict, value: DeserializedType) -> dict:
         if self.serialization_path is None:
             raise ValueError('No serialization path set!')
-        else:
-            base_class = _get_base_class(data, self.serialization_path)
```
bfolie, line 111: Why is resolving the base class no longer necessary along this code path?

src/citrine/_serialization/properties.py
bfolie: Why 1024*1024?

src/citrine/_serialization/properties.py
```diff
@@ -655,11 +738,31 @@ def _deserialize(self, data: dict) -> typing.Any:
...
-                setattr(instance, property_name, value)
+                values[property_name] = field.deserialize_from_dict(data)
+            elif field.default is not None:
+                values[property_name] = field.default
+
+        if len(init_props) > 0:
+            try:
+                instance = self.klass(**{k: v for k, v in values.items() if k in init_props})
+            except TypeError as e:
+                if "required" in str(e) and "argument" in str(e):
```
bfolie, line 755: Is there a way to make this based on the specific error class, as opposed to string-matching on the message?

src/citrine/_serialization/properties.py
bfolie: Why not?

src/citrine/_serialization/properties.py
kroenlein: https://github.com/CitrineInformatics/citrine-python/blob/9be3cded48ea64f754d1897513100fd2230d0c52/src/citrine/_serialization/properties.py#L666-L686

There's no way to get to that line where `data` is not a `dict`.

src/citrine/_serialization/properties.py
kroenlein: Error condition is that you've failed to pass a required argument to __init__.  This fires off `TypeError` with messges of the form 
`TypeError: CLASS.__init__() missing 1 required positional argument: 'arg_name'`
or 
`TypeError: CLASS.__init__() missing 1 required keyword-only argument: 'arg_name'`
I thought just catching a TypeError would catch too many other potential issues.  There's no tighter error type to filter by.

src/citrine/_serialization/properties.py
kroenlein: Arbitrarily large while not very large.  A tighter estimate would be total number of property objects in all the files, but I figured trying to guess at that wasn't worth the effort.

src/citrine/_serialization/properties.py
bfolie: What I'm wondering is if we could change those from being TypeErrors to being some other type and then recognize that?

src/citrine/_serialization/properties.py
kroenlein: Because it already needs to know to ask `DictSerializable` to decode the object, and `DictSerializable` knows how to decode LinkByUIDs.

src/citrine/_serialization/properties.py
kroenlein: We'd need to scan the __init__ argument list and compare what we're about to pass it.  This is actually not terrible because python provides [https://docs.python.org/3.10/library/inspect.html#inspect.getfullargspec] and [https://docs.python.org/3.10/library/inspect.html#inspect.signature], but I, again, trying to maintain a bit of velocity.

kroenlein: Test implemented with [signature](https://docs.python.org/3.10/library/inspect.html#inspect.signature).

src/citrine/resources/api_error.py
kroenlein: Classes rebuilt as per _serialization.

tests/resources/test_gemd_resource.py
```diff
@@ -120,6 +145,51 @@ def test_register(gemd_collection):
             assert pair[1] == registered.uids[pair[0]]
 
 
+def test_gemd_register(gemd_collection):
+    """Check that register routes to the correct collections"""
```
bfolie, line 149: This is also checking that register appends a uid to the returned object, but not the original object, yes?

kroenlein: I believe the intended behavior in GEMD object registration is to edit the incoming objects such that they also contain new identifiers.  For objects with no UIDs, the act of serialization (all client-side) requires that they have at least one UID and CitrineIDs are assigned client side.  That alone would modify the objects.

This may be inconsistent with other behaviors in the library, and is certainly inconsistent with a `no side effects` design paradigm, but gemd-python doesn't follow a `no side effects` policy around serialization.

tests/serialization/test_material_run.py
```diff
@@ -153,6 +155,38 @@ def test_cake():
...
+                if dct1[k] != dct2[k]:
+                    print(k)
+            blah = dct1["ingredients"][0]._dict_for_compare()
+            labh = dct2["ingredients"][0]._dict_for_compare()
+            assert left.process == right.process
+        if left != right:
+            print(left.name)
+            print(left)
+            print(right)
+            break
```
bfolie, line 191: outdated debugging?

tests/serialization/test_material_run.py
```diff
@@ -153,6 +155,38 @@ def test_cake():
     method attempts to traverse the object tree, and requires an overhaul of build().
```
bfolie, line 155: It looks like this docstring is no longer valid

tests/_serialization/test_object_serialization.py
```diff
@@ -61,3 +61,90 @@ def test_bad_object_serde():
...
+    with pytest.raises(AttributeError, match="overridden_value"):
+        OverrideTestClass.build(data={"overridden_value": "string"})
+
+    obj = OverrideTestClass()
+    with pytest.raises(AttributeError, match="overridden_value"):
+        obj.overridden_value = "1"
+    with pytest.raises(AttributeError, match="overridden_option"):
+        obj.overridden_option = "1"
+    with pytest.raises(AttributeError, match="overridden_value"):
+        _ = obj.overridden_value
```
bfolie, line 81: What's going on here?

src/citrine/_serialization/properties.py
bfolie: It just strikes me as a strange way to specify. If you want the cache size to be about a million, why not just say `maxsize=1e6`?

tests/_serialization/test_object_serialization.py
kroenlein: Linters complain when a object getter is called in null context.  This tells the linter you know what you are asking for without creating a new symbol.

src/citrine/_serialization/properties.py
kroenlein: A bias towards base 2, but there's no compelling reason.

tests/_serialization/test_object_serialization.py
bfolie: Sorry, I should have been more precise. Why does calling the getting throw an Attribute Error?

kroenlein: Sorry, yes, very different question.  In a perfect world, this would be a compile time error because what's getting caught is that you've asked to override a method that doesn't appear in a parent class.  However, when the Property is defined, it doesn't actually know its parentage, and even if it did, it wouldn't know to which class variable you are going to assign it.  The easiest solution then is to throw an error at runtime when you can actually introspect enough to resolve an MRO.

src/citrine/_serialization/properties.py
kroenlein: Benchmarking showed saturation around 128, so I changed both caches down to 1024.  If you have strong opinions, I can change that to `2**10` or `10**3` (both are which are int)

bfolie:
Let me know once you've run the e2es and then I'll approve

kroenlein:
Dev e2es:
https://jenkins.corp.citrine.io/job/nextgen-ITs/37703

Stage e2es:
https://jenkins.corp.citrine.io/job/nextgen-ITs/37704

bfolie (approved):
Looks good, but @anoto-moniz should weigh in as well

src/citrine/resources/condition_template.py
```diff
@@ -53,7 +48,7 @@ def __init__(self,
                  ):
         if uids is None:
             uids = dict()
-        DataConcepts.__init__(self, GEMDConditionTemplate.typ)
+        DataConcepts.__init__(self)
```
anoto-moniz, line 51: This section is a bit odd, though it predates this commit. I see why we invoke `__init__` directly on separate classes, but why not use super to ensure we respect the parent's MRO and anything in between?

`super(AttributeTemplate, self).__init__()`

kroenlein: Dunno if I've ever seen that pattern preferred, but I don't see any harm.

tests/_serialization/test_simple_properties.py
```diff
@@ -240,13 +240,13 @@ def test_linkorelse_deserialize_requires_serializable():
 
 def test_linkorelse_deserialize_requires_scope_and_id():
     loe = LinkOrElse()
-    with pytest.raises(ValueError):
+    with pytest.raises(TypeError):
```
anoto-moniz, line 243: This appears to be an interface break. If so, we need to either force it to a `ValueError` for the time being and make the switch come 3.0, or make sure we very clearly and proactively (read: not just release notes) communicate when it happens, so people can update their scripts if need be.

tests/resources/test_object_setters.py
```diff
@@ -53,7 +53,7 @@ def test_object_pointer_serde():
 def test_object_validation():
     """Test that an object pointing to another object is validated."""
     meas = MeasurementRun("A measurement")
-    with pytest.raises(TypeError):
+    with pytest.raises(ValueError):
```
anoto-moniz, line 56: Same thing here as with LOE above.

src/citrine/_serialization/properties.py
```diff
@@ -22,22 +23,52 @@
 
 
 class Property(typing.Generic[DeserializedType, SerializedType]):
+    """
+    Basic operating unit of a serialization layer.
+
+    By defining a Property as a class variable, access methods are defined and
+    the
```
anoto-moniz, line 30: ???

tests/resources/test_object_setters.py
kroenlein: The problem here is that gemd-python (correctly) throws TypeErrors when passing arguments of the wrong type, and _serialization throws ValueErrors.  This change is reflective of the new MRO.  

It's even a little grosser than that, since the error message thrown is actually misleading.
```
ValueError: <class 'citrine.resources.measurement_run.MeasurementRun'> <Measurement run 'A measurement'> is not one of valid types: (<class 'dict'>, <class 'NoneType'>) for MaterialRun:process
```

tests/_serialization/test_simple_properties.py
kroenlein: Ironically, solvable with the try-except block that @bfolie asked me to replace with the signature test, albeit in this case the exception in thrown by a dependency.  Thinking about how to address this, given the harder issue below, but it's worth noting that this case should really be a TypeError because missing __init__ arguments are TypeErrors in the Python core libraries.

anoto-moniz: It makes sense to me this should be a `TypeError`, but that means a user's `try-except` won't catch it any more. So as annoying as it is, I do think we should catch it, and make a note to change it at the 3.0 release.

tests/resources/test_object_setters.py
anoto-moniz: Ugh, I've seen that error message and it's awful.

Sounds like the underlying issue is that incongruence, and it will likely be more complicated to preserve as is now. Is `deserialize` an operation users will often call, or is it more an internally used method?

We should move `_serialization` to `TypeError` when we go to 3.0.

PR #852, opened 2023-05-05
Deprecate dehydrated design spaces.

anoto-moniz:
Dehydration is a reuse strategy which we don't really support any more. In fact, using it will result in an error when registering or updating a design space. However, since it's still documented, we should document it and do the right thing for now, but drop it in the next major release.

### PR Type:


kroenlein (approved):
Tracks.  Thanks for the clean-up!

PR #856, opened 2023-05-19
add param and updated test, docs, and version

StevenKauwe:
# Citrine Python PR

## Description 

### PR Type:


src/citrine/informatics/generative_design.py
```diff
@@ -128,9 +136,11 @@ def __init__(
         min_fingerprint_similarity: float,
         mutation_per_seed: int,
         structure_exclusions: Optional[List[StructureExclusion]],
+        min_substructure_counts: Dict[str, int],
```
anoto-moniz, line 139: @kylecsnow Do customers always need to provide `min_substructure_counts`? If not, let's have it default to `{}` for convenience.

src/citrine/informatics/generative_design.py
```diff
@@ -128,9 +136,11 @@ def __init__(
         min_fingerprint_similarity: float,
         mutation_per_seed: int,
         structure_exclusions: Optional[List[StructureExclusion]],
```
anoto-moniz, line 138: If we need to make other tweaks, let's fix this. Looks like the type-hinting should drop `Optional` and specify a default value of `[]`.

anoto-moniz (dismissed):
A couple tiny nitpicks, but that's it. I'm not going to bock merging based on them.

src/citrine/informatics/generative_design.py
kylecsnow: @anoto-moniz Mostly because I don't frequently find myself writing highly clean code at the citrine-python standard of quality, and don't often consider the nuances between using something like `Optional` vs. other alternatives, I want to better understand:

Why we wouldn't want to use `Optional` here?  On the pymithril side of this Kaai, Sean, and I came to decide to use `Optional` for both `structure_exclusions` and `min_substructure_counts` arguments (ex: comment thread https://github.com/CitrineInformatics/pymithril/pull/16#discussion_r1182590037)

src/citrine/informatics/generative_design.py
kylecsnow: @anoto-moniz This should be an optional argument generally.  But I've got the same question about this as in my other comment reply, should this instead be `min_substructure_counts: Optional[Dict[str, int]],`? 
(please see other comment thread)

src/citrine/informatics/generative_design.py
anoto-moniz: They're correct that you've got to be careful when setting containers as default values, as the article they link details. The thing is that it's only relevant if you're modifying the collection you've been given, which is to say almost never. Hence why I tend to still default towards using them.

Their way is definitely safer, though, in that the issue cannot occur. So you'd set this argument to default to `None`, and then in the method body assign it `[]` if it's `None` and the value otherwise. You can do that easily with something like

`self.structure_exclusions = structure_exclusions if structure_exclusions else []`

Or you can use a slick Python idiom to reduce repetition:

`self.structure_exclusions = structure_exclusions or []`

kylecsnow: Makes sense, thanks!  I went with the latter suggestion and did the same for `min_substructure_counts`:
`self.min_substructure_counts = min_substructure_counts or {}`).

kylecsnow (commented):
If everything else looks good @anoto-moniz, I think we're ready to merge?

anoto-moniz:
> If everything else looks good @anoto-moniz, I think we're ready to merge?

Sorry to be a paint, but if an argument is `Optional`, it should default to `None`. Right now, callers have to specify `structure_exclusions` and `min_substructure_counts`. After that, yes, I'll be set to approve.

anoto-moniz (approved):
Looks good! Thanks for taking this across the finish line, Kyle.

PR #858, opened 2023-05-26
PLA-12608: Featurized training data

sfriedowitz:
# Citrine Python PR

Adds support for `featurized-training-data` endpoint for a predictor version. To support this, we implement the classes for hierarchical design materials/candidates.

### PR Type:


src/citrine/informatics/design_candidate.py
```diff
@@ -124,6 +126,14 @@ def __init__(self, *, smiles: str):
 class DesignMaterial(Serializable["DesignMaterial"]):
     """Description of the material that was designed, as a set of DesignVariables."""
 
+    material_id = properties.UUID('identifiers.id')
```
sfriedowitz, line 129: The backend populates these fields already. They were just being ignored before.

src/citrine/resources/predictor.py
```diff
@@ -125,6 +126,17 @@ def get(self,
...
 
+    def get_featurized_training_data(
+            self,
+            uid: Union[UUID, str],
+            *,
+            version: Union[int, str] = MOST_RECENT_VER
+    ) -> List[HierarchicalDesignMaterial]:
+        version_path = self._construct_path(uid, version)
+        full_path = f"{version_path}/featurized-training-data"
+        payload = self.session.get_resource(full_path, version=self._api_version)
```
sfriedowitz, line 137: Based on how the backend streams the response out, it is not trivial to chunk the incoming JSON per element in the stream (see more details [here](https://doc.akka.io/docs/akka-http/current/routing-dsl/source-streaming-support.html#customising-response-rendering-mode)).

Because of that, I am simply loading the full JSON payload into memory and then parsing into the design material objects. We'll see if this ends up causing issues on bigger sample sizes, and can adjust accordingly then

src/citrine/informatics/design_candidate.py
```diff
@@ -132,8 +142,28 @@ def __init__(self, *, values: dict):
...
 
+class HierarchicalDesignMaterial(Serializable["HierarchicalDesignMaterial"]):
+    """Description of a designed material as a set of connected individual materials.
+
+    A hierarchical material contains a root material, containing identifiers and variables,
+    along any of sub-materials appearing in the history of the root.
+    Connections between the root and sub-materials are contained in the mixtures dictionary
+    that associates each material (by Citrine ID) with the ingredients that comprise it.
+    """
+
```
sfriedowitz, line 153: It was a mistake to call this JSON field "terminal" in the backend, but its too late for that now. Hence why I am renaming it here.

PR #860, opened 2023-05-31
sample from search space api

StevenKauwe:
# Citrine Python PR

## Description 

### PR Type:


src/citrine/resources/project.py
```diff
@@ -150,6 +152,13 @@ def generative_design_executions(self) -> GenerativeDesignExecutionCollection:
         """Return a collection representing all visible generative design executions."""
         return GenerativeDesignExecutionCollection(project_id=self.uid, session=self.session)
 
+    @property
+    def sample_design_space_executions(self) -> SampleDesignSpaceExecutionCollection:
```
anoto-moniz, line 156: Since this requires a design space ID, it doesn't belong as a property on the `ProjectCollection`. It would fit on a `DesignSpace` object, similar to how we have [reports on predictors](https://github.com/CitrineInformatics/citrine-python/blob/main/src/citrine/informatics/predictors/graph_predictor.py#L120).

tests/resources/test_generative_design_execution.py
```diff
@@ -51,7 +51,7 @@ def test_build_new_execution(collection, generative_design_execution_dict):
 def test_trigger_execution(collection: GenerativeDesignExecutionCollection, generative_design_execution_dict, session):
     # Given
     session.set_response(generative_design_execution_dict)
-    design_execuption_input = GenerativeDesignInput(
+    design_execution_input = GenerativeDesignInput(
```
anoto-moniz, line 54: 😂

src/citrine/informatics/design_candidate.py
```diff
@@ -202,3 +203,20 @@ class HierarchicalDesignCandidate(Serializable["HierarchicalDesignCandidate"]):
...
+
+    uid = properties.UUID('id')
+    """:UUID: unique external Citrine ID of the material"""
+    execution_uid = properties.UUID('id')
+    """:UUID: unique external Citrine ID of the execution"""
+    material = properties.Object(HierarchicalDesignMaterial, "material")
+    """:HierarchicalDesignMaterial: the material returned by the design workflow"""
+
+    def __init__(self):
+        pass  # pragma: no cover
```
anoto-moniz, line 222: Due to a tweak I made a little while ago, I'd expect this `__init__` is not necessary. If you're getting errors after deleting it, I can take a look.

src/citrine/informatics/executions/sample_design_space_execution.py
```diff
@@ -0,0 +1,101 @@
...
+):
+    """The execution of a Sample Design Space task.
+
+    Possible statuses are INPROGRESS, SUCCEEDED, and FAILED.
+    Design executions also have a ``status_description`` field with more information.
+    """
+
+    _paginator: Paginator = Paginator()
+    _collection_key = 'response'
+    _in_progress_statuses = ["INPROGRESS"]
```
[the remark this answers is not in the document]
anoto-moniz: Hmmmmmmmmmm...but this is actually an execution, so that doesn't make the most sense. Maybe we should have a base `Execution` type instead, since I see you're just doing the same thing [DesignExecution](https://github.com/CitrineInformatics/citrine-python/blob/main/src/citrine/informatics/executions/design_execution.py#L21) does.

src/citrine/informatics/executions/sample_design_space_execution.py
```diff
@@ -0,0 +1,101 @@
...
+from citrine.informatics.design_candidate import SampleSearchSpaceResultCandidate
+
+
+class SampleDesignSpaceExecution(
+    Resource["SampleDesignSpaceExecution"], Pageable, AsynchronousObject
+):
+    """The execution of a Sample Design Space task.
+
+    Possible statuses are INPROGRESS, SUCCEEDED, and FAILED.
+    Design executions also have a ``status_description`` field with more information.
```
anoto-moniz, line 21: Update name.

src/citrine/informatics/executions/sample_design_space_execution.py
```diff
@@ -0,0 +1,101 @@
...
+    design_space_id: Optional[UUID] = None
+
+    uid: UUID = properties.UUID('id', serializable=False)
+    """:UUID: Unique identifier of the execution"""
+    status = properties.Optional(properties.String(), 'status', serializable=False)
+    """:Optional[str]: short description of the execution's status"""
+    status_description = properties.Optional(
+        properties.String(), 'status_description', serializable=False)
+    """:Optional[str]: more detailed description of the execution's status"""
+    status_info = properties.Optional(
```
anoto-moniz, line 40: I'd suggest dropping the `Optional`, since it being missing is functionally the same as it being an empty list.

src/citrine/informatics/executions/sample_design_space_execution.py
```diff
@@ -0,0 +1,101 @@
...
+    update_time = properties.Optional(properties.Datetime, 'update_time', serializable=False)
+    """:Optional[datetime]: date and time at which the resource was most recently updated,
+    if it has been updated"""
+
+    def __init__(self):
+        """The SampleDesignSpace executions are not directly instantiated by the user."""
+        pass  # pragma: no cover
+
+    def __str__(self):
+        return '<SampleDesignSpaceExecution {!r}>'.format(str(self.uid))
```
anoto-moniz, line 63: nitpick: use an f-string.

src/citrine/informatics/executions/sample_design_space_execution.py
```diff
@@ -0,0 +1,101 @@
...
+    _in_progress_statuses = ["INPROGRESS"]
+    _succeeded_statuses = ["SUCCEEDED"]
+    _failed_statuses = ["FAILED"]
+    _session: Optional[Session] = None
+    project_id: Optional[UUID] = None
+    design_space_id: Optional[UUID] = None
+
+    uid: UUID = properties.UUID('id', serializable=False)
+    """:UUID: Unique identifier of the execution"""
+    status = properties.Optional(properties.String(), 'status', serializable=False)
```
anoto-moniz, line 35: Are there situations where the status won't be defined on the response payload? Or cases where this object will be built as an input payload?

src/citrine/informatics/executions/sample_design_space_execution.py
```diff
@@ -0,0 +1,101 @@
...
+    """:Optional[str]: short description of the execution's status"""
+    status_description = properties.Optional(
+        properties.String(), 'status_description', serializable=False)
+    """:Optional[str]: more detailed description of the execution's status"""
+    status_info = properties.Optional(
+        properties.List(properties.String()),
+        'status_info',
+        serializable=False
+    )
+
```
anoto-moniz, line 45: tnitpick: This space makes it look like the next docstring is associated with `created_by`.

src/citrine/informatics/executions/sample_design_space_execution.py
```diff
@@ -0,0 +1,101 @@
...
+    """:Optional[List[str]]: human-readable explanations of the status"""
+    created_by = properties.Optional(properties.UUID, 'created_by', serializable=False)
```
anoto-moniz, line 47: Same for all these `Optional` fields. If there's no situation where they will be missing (or `null`), then they don't need to be `Optional`.

src/citrine/informatics/executions/sample_design_space_execution.py
```diff
@@ -0,0 +1,101 @@
...
+    updated_by = properties.Optional(properties.UUID, 'updated_by', serializable=False)
+    """:Optional[UUID]: id of the user who most recently updated the resource,
+    if it has been updated"""
+    create_time = properties.Optional(properties.Datetime, 'create_time', serializable=False)
+    """:Optional[datetime]: date and time at which the resource was created"""
+    update_time = properties.Optional(properties.Datetime, 'update_time', serializable=False)
+    """:Optional[datetime]: date and time at which the resource was most recently updated,
+    if it has been updated"""
+
+    def __init__(self):
```
anoto-moniz, line 58: Shouldn't be needed any more.

src/citrine/resources/sample_design_space_execution.py
```diff
@@ -0,0 +1,84 @@
...
+    def register(self, model: SampleDesignSpaceExecution) -> SampleDesignSpaceExecution:
+        """Cannot register an execution."""
+        raise NotImplementedError("Cannot register a SampleDesignSpaceExecution.")
+
+    def update(self, model: SampleDesignSpaceExecution) -> SampleDesignSpaceExecution:
+        """Cannot update an execution."""
+        raise NotImplementedError("Cannot update a SampleDesignSpaceExecution.")
+
+    def list(self, *,
+             page: Optional[int] = None,
```
anoto-moniz, line 51: `page` is deprecated, and should be removed. It probably should have been removed from everything in our last major version bump.

tests/conftest.py
```diff
@@ -706,6 +713,20 @@ def example_candidates(example_design_material):
     }
 
 
+@pytest.fixture()
+def example_sample_design_space_response(example_hierarchical_design_material):
+    return {
+        "page": 1,
```
anoto-moniz, line 719: Drop page

src/citrine/informatics/design_spaces/design_space.py
```diff
@@ -35,3 +37,10 @@ def get_type(cls, data) -> Type[Serializable]:
             'FormulationDesignSpace': FormulationDesignSpace,
             'DataSourceDesignSpace': DataSourceDesignSpace
         }[data['config']['type']]
+
+    @property
+    def sample_design_space_executions(self):
```
StevenKauwe, line 42: I wasn't sure the pattern to use here. On one hand, just calling sample would be awesome. But that kind of hides what is going on. So I named it this way.

src/citrine/informatics/design_spaces/sample_design_space.py
```diff
@@ -0,0 +1,18 @@
+from citrine._serialization import properties
+from citrine._serialization.serializable import Serializable
+
+
+class SampleDesignSpaceInput(Serializable['SampleDesignSpaceInput']):
```
StevenKauwe, line 5: Moved to its own file to remove circular dependency

src/citrine/informatics/executions/base_execution.py
```diff
@@ -0,0 +1,62 @@
...
+from citrine._rest.paginator import Paginator
+from citrine._serialization import properties
+from citrine._session import Session
+from citrine.resources.status_detail import StatusDetail
+
+
+from deprecation import deprecated
+
+
+class Execution(Pageable, AsynchronousObject, ABC):
```
StevenKauwe, line 16: Added a base execution class. This is used for `design_execution`, `generative_design_execution`, `predictor_evaluation_execution` (this one doesn't NEED pagable, but also doesn't hurt?) and `sample_design_space_execution`.

StevenKauwe: Also added `EXECUTING` to the `in_progress_statuses`, since we used that for the sample_design_space on the backend.

StevenKauwe: I named the file `base_execution.py`. I'm wondering if that should just be called `Execution`. I was on the fence. (maybe there is a different better name).

src/citrine/informatics/executions/sample_design_space_execution.py
StevenKauwe: I missed this comment on my last push. But I have a local update to the doc string as
```
    """The execution of a Sample Design Space task.

    Possible statuses are EXECUTING, SUCCEEDED, and FAILED.
    Sample Design Space executions also have a ``status_detail`` field with more information.

    """
```

Was this what you were pointing out?

src/citrine/informatics/design_candidate.py
```diff
@@ -159,9 +159,6 @@ class HierarchicalDesignMaterial(Serializable["HierarchicalDesignMaterial"]):
     mixtures = properties.Mapping(properties.UUID, properties.Object(Mixture), 'mixtures')
     """:Dict[UUID, Mixture]: mapping from Citrine ID to components the material is composed of"""
 
-    def __init__(self):
```
StevenKauwe, line 161: I removed all the __init__ in this file that didn't do anything. Not sure if you wanted all of em removed. 🤷  But tests still pass. 😆

tests/conftest.py
```diff
@@ -661,10 +661,18 @@ def example_holdout_result_dict(example_holdout_evaluator_dict, example_rmse_met
...
 
 
-
 @pytest.fixture
 def sample_design_space_execution_dict(generic_entity):
     ret = generic_entity.copy()
+    ret.update(
+        {
+            "design_space_id": str(uuid.uuid4()),
+            "status": {
```
StevenKauwe, line 670: It looks like, until now, we have always unpacked the `Status` response in the backend. We could consider making a class for this. But I figured I could solve it in the `_pre_build` method for now.

StevenKauwe (commented):
Made updates to include base class. Made a test script, and the end point does give results!!! 🙌  We know it runs. Now we just need to double check that the code looks the way we want.

src/citrine/informatics/executions/base_execution.py
anoto-moniz: And yeah, I think just calling the file `execution.py` would be fine.

src/citrine/informatics/executions/sample_design_space_execution.py
```diff
@@ -2,73 +2,37 @@
...
             '/projects/{project_id}/design-spaces/{design_space_id}/sample/',
             project_id=self.project_id,
             design_space_id=self.design_space_id,
         )
 
+    @classmethod
+    def _pre_build(cls, data: dict) -> dict:
+        """Run data modification before building."""
+        status = data.get('status')
+        update_dict = {"status": status["minor"], "status_detail": status["detail"]}
```
anoto-moniz, line 32: To maintain consistency, the `status` field should be the major status. Use `status_description` for the minor status.

src/citrine/informatics/design_spaces/design_space.py
anoto-moniz: We do both in different classes, so it's your call.

src/citrine/informatics/design_spaces/sample_design_space.py
anoto-moniz: What circular dependency?

StevenKauwe: This was introduced when moving the execution collection from projects.py to design_space.py.

src/citrine/informatics/executions/sample_design_space_execution.py
```diff
@@ -28,9 +28,20 @@ def _path(self):
     @classmethod
     def _pre_build(cls, data: dict) -> dict:
         """Run data modification before building."""
-        status = data.get('status')
-        update_dict = {"status": status["minor"], "status_detail": status["detail"]}
-        data.update(update_dict)
+        status_dict = data['status']
+
+        # Set major status
+        if "major" not in status_dict and status_dict["minor"] in ["EXECUTING", "VALIDATING"]:
```
anoto-moniz, line 34: I think we always want to set `status_dict['major']` based on the value of `status_dict['minor']` if the server doesn't. That way, no matter the status, the interface matches the other executions.

src/citrine/informatics/executions/sample_design_space_execution.py
```diff
@@ -30,17 +30,34 @@ def _pre_build(cls, data: dict) -> dict:
...
+            'INVALID': 'FAILED',
+            'CANCELLED': 'FAILED',
+            'RESOURCE_EXCEEDED': 'FAILED',
+            'ERROR': 'FAILED',
+        }
+
+        if "major" not in status_dict:
+            if status_dict["minor"] in major_status_mapping:
+                status_dict["major"] = major_status_mapping[status_dict["minor"]]
+        else:
```
anoto-moniz, line 52: This `else` should be indented one level to do what the error message says.

src/citrine/informatics/executions/sample_design_space_execution.py
```diff
@@ -30,17 +30,34 @@ def _pre_build(cls, data: dict) -> dict:
...
+            raise ValueError(f"Unknown minor status: {status_dict['minor']}")  # pragma: no cover
```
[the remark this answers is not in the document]
StevenKauwe: I saw other places with ValueError using no cover. So I just kinda went with it. It is super easy to make a test for this. I have no idea why my brain decided this was the right move.. 😅  I can add that when fixing else from above.

anoto-moniz (approved):

PR #861, opened 2023-06-02
Add class migration / deprecation meta

kroenlein:
# Citrine Python PR

## Description 
This PR adds a metaclass for marking a class that has been moved or renamed as deprecated, removing the need for
the `shadow_classes_in_module` method and removing an implicit import procedure.  `shadow_classes_in_module` is also therefore marked as deprecated.

It also adds a `raise_error` argument to `_poll_for_job_completion`, since raising an exception isn't always the correct choice when a job finishes with a less-than-ideal state.  That probably should have been another PR, but I was a little sloppy.

### PR Type:
- [X ] Maintenance (non-breaking change to assist developers)


StevenKauwe (approved):
This is my first time encountering deprecation for classes that have been move to new packages, but the code and logic LGTM.

PR #864, opened 2023-06-05
Add --only-binary pandas to keep testing performant

kroenlein:
# Citrine Python PR

## Description 
* Add --only-binary to travis install so that tests under 3.11 run in reasonable time. Pandas was take 15 minutes to build a local version that was compatible with all pytz 2019.3
* Boost maximum versions of libraries so actual latest libraries can be tested
* Update no-warnings test to remove deprecation warning, see https://docs.pytest.org/en/latest/how-to/capture-warnings.html#additional-use-cases-of-warnings-in-tests

### PR Type:


setup.py
```diff
@@ -24,24 +24,24 @@
       install_requires=[
           "requests>=2.31.0,<3",
           "pyjwt>=2,<3",
-          "arrow>=0.15.4,<0.16",
+          "arrow>=0.15.4,<2",
```
anoto-moniz, line 27: Did the interface not change at all in 1.0?

anoto-moniz (approved):

setup.py
kroenlein: It changed, just not in a way that impacts us:
https://github.com/arrow-py/arrow/issues/832

PR #868, opened 2023-06-16
[PLA-12745] Use design space v3 endpoints.

anoto-moniz:
Update the design space endpoints to use the new v3 design space endpoints. This means an update in the structure and parsing and deprecating some fields, although the interface stays the same.

### PR Type:


anoto-moniz:
e2e run: https://jenkins.corp.citrine.io/view/e2es/job/nextgen-ITs/39261/

src/citrine/informatics/design_spaces/formulation_design_space.py
```diff
@@ -37,26 +36,23 @@ class FormulationDesignSpace(Resource['FormulationDesignSpace'], DesignSpace, AI
...
 
-    _resource_type = ResourceTypeEnum.MODULE
-
     formulation_descriptor = properties.Object(
         FormulationDescriptor,
-        'config.formulation_descriptor'
+        'data.instance.formulation_descriptor'
     )
-    ingredients = properties.Set(properties.String, 'config.ingredients')
+    ingredients = properties.Set(properties.String, 'data.instance.ingredients')
```
StevenKauwe, line 43: Where in the backend does this correspond to? `data.instance` -> `OrionDesignSpace` (Formulation in this case)?

src/citrine/informatics/modules.py
```diff
@@ -30,6 +30,4 @@ class Module(PolymorphicSerializable['Module'], AsynchronousObject):
     def get_type(cls, data) -> Type['Module']:
         """Return the subtype."""
         from citrine.informatics.design_spaces import DesignSpace
-        return {
```
StevenKauwe, line 33: Bleh, this must have felt nice to update.

anoto-moniz: I'm so excited for 3.0, when we can rip it all out completely! =D

src/citrine/resources/design_space.py
```diff
@@ -55,33 +56,91 @@ def _validate_write_request(self, design_space: DesignSpace):
...
+        self._verify_write_request(design_space)
+        hydrated_ds = self._hydrate_design_space(design_space)
+        updated_ds = super().update(hydrated_ds)
+
+        # The /api/v3/design-spaces endpoint switched archiving from a field on the update payload
+        # to their own endpoints. To maintain backwards compatibility, all design spaces have an
+        # _archived field set by the archived property. It will be archived if True, and restored
+        # if False. It defaults to None, which does nothing. The value is reset afterwards.
+        if design_space._archived is True:
+            self.archive(design_space.uid)
```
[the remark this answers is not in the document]
anoto-moniz: Philosophically, it enforces separation of concerns. The update endpoint can focus on changes to the data itself, whereas archive/restore can focus on the archive state.
It also allows the client to archive without needing to pass a copy of the resource. The client need only know the resource's ID.
And it's a natural way to pass that information to the backend. Previously, it was part of the payload. But in the new setup, since archive status is metadata, it's not part of the input payload. We could allow it to be specified as a query parameter, but that feels clunky as a way to direct action (which is part of the reason `validate` is now it's own endpoint).

src/citrine/resources/design_space.py
```diff
@@ -55,33 +56,91 @@ def _validate_write_request(self, design_space: DesignSpace):
                 raise ValueError(msg.format(self._enumerated_cell_limit, width * length))
         return
```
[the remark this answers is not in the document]
anoto-moniz: Nope, no idea. It should be removed, as it's redundant.

src/citrine/informatics/design_spaces/formulation_design_space.py
anoto-moniz: Correct. If it's a top-level DS, then it's just the new structure. If it's a subspace of a product DS, then when the product DS is deserialized it wraps its children in the scaffolding of the top-level object so the polymorphic serializable can do its thing.

StevenKauwe (approved):
LGTM 🚀

PR #869, opened 2023-06-21
Update ingestion to return status

kroenlein:
# Citrine Python PR

## Description 
This PR updates the [ALPHA] ingest endpoint support, trying to bring consistency to error handling across the various contexts.  It also now correctly returns the ingestion status in failure scenarios (as opposed to JobFailureError), since that is where the actionable feedback resides.

### PR Type:


anoto-moniz (approved):
LGTM

src/citrine/resources/ingestion.py
```diff
@@ -200,6 +217,70 @@ def status(self) -> IngestionStatus:
         return IngestionStatus.build(self.session.get_resource(path=path))
 
 
+class FailedIngestion(Ingestion):
```
StevenKauwe, line 220: I'm not immediately understanding the FailedIngestion class and why it inherits from Ingestion. 😅  Everything looks good and logical, just hoping to get an explanation!

kroenlein: The shapes of the responses under different types of errors are different -- the errors returned when a file upload files are structured differently than when there's a missing ingredient, for example.  

The `FailedIngestion` is essentially written to unify the interfaces so that no matter how your ingest failed, you end up with a consistent object to query.  When you get a failure early in the process, this class makes assumptions how to enrich that to look like the later failures.

PR #873, opened 2023-07-10
PLA-12831: Integer support

sfriedowitz:
# Citrine Python PR

Adds support for integer descriptor, integer range constraint, and integer dimension. Also deprecated `units` on integer descriptor, as they cannot be created on platform.

### PR Type:


docs/source/workflows/generative_design.rst
```diff
@@ -1,4 +1,4 @@
-.. generative_design_execution:
+.. _generative_design:
```
sfriedowitz, line 1: It felt weird that this file was given its own special section outside of the AI Engine documentation, so I am moving it within that heading.

docs/source/data_extraction.rst
```diff
@@ -287,3 +289,43 @@ There are several ways to define columns, depending on the type of the attribute
 * Collections of values
 
  * :class:`~citrine.gemtables.columns.ConcatColumn`: for concatenating the results of a list- or set-valued result, such as is returned by :class:`~citrine.gemtables.variables.IngredientLabelsSetInOutput`
+
+Compatibility with AI Engine
```
sfriedowitz, line 293: @ventura-rivera asked us to write this documentation

tests/informatics/test_dimensions.py
```diff
@@ -12,6 +14,13 @@ def continuous_dimension() -> ContinuousDimension:
     return ContinuousDimension(alpha, lower_bound=3, upper_bound=33)
 
 
+@pytest.fixture
+def integer_dimension() -> IntegerDimension:
```
bfolie, line 18: Is this used anywhere?

src/citrine/informatics/constraints/integer_range_constraint.py
```diff
@@ -0,0 +1,48 @@
...
+    descriptor_key: str
+        the key corresponding to a descriptor
+    lower_bound: int
+        the minimum value in the range
+    upper_bound: int
+        the maximum value in the range
+    lower_inclusive: bool
+        if True, will include the lower bound value in the range (default: true)
+    upper_inclusive: bool
+        if True, will include the max value in the range (default: true)
```
bfolie, line 24: Do we need to expose inclusive options? After all, the integer constraint [a, b] with exclusive bounds can also be written as [a+1, b-1] with inclusive bounds

tests/informatics/test_constraints.py
```diff
@@ -18,6 +18,17 @@ def scalar_range_constraint() -> ScalarRangeConstraint:
     )
 
 
+@pytest.fixture
```
bfolie, line 21: No need to define this as a fixture if it's only used in one test

sfriedowitz: I'm following the pattern used by all other tests in this file.

src/citrine/informatics/constraints/integer_range_constraint.py
sfriedowitz: Hmm, idk why I ever added that in the backend. I think it was already there in mithril. Should probably just remove it.

sfriedowitz: Im just gonna remove from backend and here.

docs/source/data_extraction.rst
```diff
@@ -287,3 +289,43 @@ There are several ways to define columns, depending on the type of the attribute
...
+found on the attribute template referenced by a :ref:`GEM Table variable <variables>`.
+The key of the descriptor is derived from concatenation of the `headers` field of the table variable.
+An exception to this is for the :class:`~citrine.informatics.descriptors.FormulationDescriptor`
+which follows the special rule described below.
+The mappings from variables in a GEM Table to descriptors are as follows:
+
+- :class:`~gemd.entity.bounds.RealBounds` are converted to a :class:`~citrine.informatics.descriptors.RealDescriptor`
+- :class:`~gemd.entity.bounds.IntegerBounds` are converted to an :class:`~citrine.informatics.descriptors.IntegerDescriptor`
+- :class:`~gemd.entity.bounds.CategoricalBounds` are converted to a :class:`~citrine.informatics.descriptors.CategoricalDescriptor`
+- :class:`~gemd.entity.bounds.CompositionBounds` are converted to a :class:`~citrine.informatics.descriptors.ChemicalFormulaDescriptor`
```
ventura-rivera, line 312: if the following statement is true, we should at it to the end of this line:
`or CategoricalDescriptor when the quantity keys are not atomic symbols`

ventura-rivera (changes requested):
Looks great! Just one request on the docs.

docs/source/data_extraction.rst
sfriedowitz: That is not how it works, though. A `CompositionBounds` always produces a `ChemicalFormulaDescriptor`.

In order for a `NominalComposition` to work with a `CategoricalDescriptor` in the AIE, you'd have to pair it with a `CategoriclaBounds`. Is that even possible?

tests/informatics/test_constraints.py
bfolie: It's a bad pattern because it splits the test logic into two places without providing any reproducibility benefits.

Looking at the history, when this file was first created each fixture was used for multiple tests. But that's no longer the case (except for the ingredient ratio constraint)

sfriedowitz: Sure, its probably a bad pattern. But i'd rather maintain style consistency with the rest of the file

docs/source/data_extraction.rst
sfriedowitz: It looks like a `NominalComposition` actually cant be used with a `CategoricalBounds`, meaning these mappings won't work in practice. Looks like I'll have to remove the option to treat a `NominalComposition` as a categorical in cases where the keys _are not_ valid atomic symbols.

ventura-rivera: I am not sure how it works, but the thing that confuses me about how this works is this part a little further down:
```
- :class:`~citrine.informatics.descriptors.CategoricalDescriptor`: values of type :class:`~gemd.entity.NominalCategorical` and :class:`~gemd.entity.DiscreteCategorical`,
  or values of type :class:`~gemd.entity.NominalComposition` when the quantity keys are **not** atomic symbols
- :class:`~citrine.informatics.descriptors.ChemicalFormulaDescriptor`: values of type :class:`~gemd.entity.EmpiricalFormula`,
  or values of type :class:`~gemd.entity.NominalComposition` when **all** quantity keys are valid atomic symbols
```

It says that NominalComposition can be compatible with CategoricalDescriptor or ChemicalFormulaDescriptor depending on quantity keys being atomic symbols.

ventura-rivera: ok I gotcha 👌

sfriedowitz: Yep, sorry about that. I'll update the docs here too.

The logic will be that a `NominalComposition` can only be used if the keys are all valid atomic symbols.

ventura-rivera: Great thanks for documenting that! Just want to acknowledge, we have had customers want to use a NominalComposition value as a RealDescriptor, but that can be a feature request for the future 🙂

sfriedowitz: > we have had customers want to use a NominalComposition value as a RealDescriptor

How would this be possible? I don't see a clear way in which we treat that value type as a real number?

ventura-rivera: Nominal Compositions are mappings of component to quantity. When a user constructs a gemtable, they can use a `ComponentQuantityColumn` with the component name as the column name, and the quantity of that component for the given attribute will be filled into the cell values

sfriedowitz: I see, that's a bit trickier one, as we'd have to generate a set of real descriptors per value, I guess.

bfolie (approved):

ventura-rivera (approved):
thanks for the change! lgtm!

PR #876, opened 2023-07-13
PLA-12042: Support for hierarchical search spaces

sfriedowitz:
# Citrine Python PR

## Description 

### PR Type:


.travis.yml
```diff
@@ -1,36 +1,36 @@
 language: python
```
sfriedowitz, line 1: I don't know how this got formatted, and literally copying the file back verbatim from `main` doesn't seem to change it

src/citrine/informatics/design_spaces/hierarchical_design_space.py
```diff
@@ -0,0 +1,215 @@
...
+            subspaces: Optional[List[MaterialNodeDefinition]] = None,
+            data_sources: Optional[List[DataSource]] = None
+    ):
+        self.name: str = name
+        self.description: str = description
+        self.root: MaterialNodeDefinition = root
+        self.subspaces: List[MaterialNodeDefinition] = subspaces or list()
+        self.data_sources: List[DataSource] = data_sources or list()
+
+    def _post_dump(self, data: dict) -> dict:
```
sfriedowitz, line 175: All of this pre-build and post-dump stuff is necessary because the `MaterialNodeDefinition` class contains a `FormulationDesignSpace`. Because of that, we have to wrap that JSON in the `DesignSpace.wrap_instance` format such that it can be built properly.

src/citrine/resources/design_space.py
```diff
@@ -206,6 +227,54 @@ def create_default(self,
         data = self.session.post_resource(path, json=payload, version=self._api_version)
         return self.build(DesignSpace.wrap_instance(data["instance"]))
```
bfolie, line 228: Have you confirmed that this works as expected with a hierarchical design space?

src/citrine/informatics/design_spaces/hierarchical_design_space.py
```diff
@@ -0,0 +1,215 @@
...
+    referencing other sub-nodes, allowing for the linkage of complex material history shapes
+    in the resulting candidates.
+
+    Every node also contains a set of :class:`~citrine.informatics.dimensions.Dimension`s
+    used to define any attributes (i.e., properties, processing parameters)
+    that will appear on the materials produced by that node.
+
+    :class:`~citrine.informatics.data_sources.DataSource`s can be included on the configuration
+    to allow for design over "known" materials. The Citrine Platform will look up
+    the any ingredient names from formulation subspaces on the design space nodes
```
bfolie, line 129:
```suggestion
    the ingredient names from formulation subspaces on the design space nodes
```

bfolie (commented):
Have you tested out the various methods (create default, convert, build) on a live environment? The tests all look correct, but given the de/serialization complexity involved here it would be prudent to do a non-mock test

src/citrine/informatics/design_spaces/hierarchical_design_space.py
```diff
@@ -0,0 +1,215 @@
...
+        self.name = name
+        self.scope: Optional[str] = scope
+        self.attributes = attributes or list()
+        self.formulation_subspace: Optional[FormulationDesignSpace] = formulation_subspace
+        self.template_link: Optional[TemplateLink] = template_link
+        self.display_name: Optional[str] = display_name
+
+    def __repr__(self):
+        display_name = self.display_name or self.name
+        return "<MaterialNodeDefinition {!r}>".format(display_name)
```
kroenlein, line 103:
```suggestion
        return f"<MaterialNodeDefinition {display_name!r}>"
```
Not required, but I prefer f-strings from a legibility perspective.

src/citrine/informatics/design_spaces/hierarchical_design_space.py
```diff
@@ -0,0 +1,215 @@
...
+
+    """
+
+    root = properties.Object(MaterialNodeDefinition, "data.instance.root")
+    subspaces = properties.List(
+        properties.Object(MaterialNodeDefinition), "data.instance.subspaces", default=list()
+    )
+    data_sources = properties.List(
+        properties.Object(DataSource), "data.instance.data_sources", default=list()
+    )
```
kroenlein, line 155: As an FYI, by setting the default value to a particular list, all class instances that use that default value.  You actually share a single mutable object between multiple objects, and edits to one containing object will happen to all.  This is probably best read as a bug in our serialization stack.  

Just so you know.

src/citrine/resources/design_space.py
```diff
@@ -206,6 +227,54 @@ def create_default(self,
...
+        display_name: Optional[str]
+            Optional display name to include on the root material node
+
+        Returns
+        -------
+        A :class:`~citrine.informatics.design_spaces.HierarchicalDesignSpace`
+        with a single material node obtained from the input search space.
+
+        """
+        path = f"projects/{self.project_id}/design-spaces/{uid}/convert-hierarchical"
```
kroenlein, line 268: The bound terms should really be run through percent encoding.

from citrine._utils.functions import format_escaped_url

kroenlein (commented):
Mostly fine, but you should escape the terms you're binding to that URL.

src/citrine/informatics/design_spaces/hierarchical_design_space.py
sfriedowitz: I think this is just a bug in our serialization library? We do it all over the place in here.

src/citrine/resources/design_space.py
sfriedowitz: Yep! I've been running all of this code to test the serde, and it does indeed work.

sfriedowitz:
>Have you tested out the various methods (create default, convert, build) on a live environment?

Yep! It all works. Its obnoxious that our serde library is so bad that we even need that logic. But it does indeed work.

src/citrine/informatics/design_spaces/hierarchical_design_space.py
sfriedowitz: We don't even need a default. The backend has these fields as non optional but with empty list defaults, so the field will always be included

bfolie (approved):

PR #882, opened 2023-07-29
Consistent error handling and typing across ingest

kroenlein:
# Citrine Python PR

## Description 
This PR brings the errors & exceptions reported by the various ingestion endpoints into a consistent framework that mimics how things are handled on the back end.  A e2e PR will follow this one, so that the Ingestion ITs are consistent with what is seen outside the Product API.

### PR Type:


src/citrine/resources/status_detail.py
```diff
@@ -22,7 +22,7 @@ class StatusDetail(Serializable[StatusDetailType]):
     """A status message and its level."""
 
     msg = properties.String("msg")
-    level = properties.String("level")
+    level = properties.Enumeration(StatusLevelEnum, "level")
```
anoto-moniz, line 25: This is an interface break. We'll need to deprecate it as a `String` and make it an `Enumeration` in 3.0.

anoto-moniz (changes requested):
This involves a handful of interface breaks in the form of parsing strings into enums. I like the change, but we need to deprecate the `str` form until we can remove it in 3.0, at which point we can re-type it. And in the mean time, we can introduce a separate field which is the enum type, and will become a synonym in 3.0.

src/citrine/resources/ingestion.py
```diff
@@ -339,17 +467,13 @@ def build_from_file_links(self,
             response = self.session.post_resource(path=self._get_path(), json=req)
         except BadRequest as e:
             if e.api_error is not None:
-                errors = e.api_error.validation_errors
-                if len(errors) == 0:
-                    errors = [ValidationError.build({"failure_message": e.api_error.message,
-                                                     "failure_id": "failure_id"})
-                              ]
+                if len(e.api_error.validation_errors) != 0:
```
anoto-moniz, line 470: Nitpick, since it's come up before: "For sequences, (strings, lists, tuples), use the fact that empty sequences are false:"

https://peps.python.org/pep-0008/

kroenlein:
> This involves a handful of interface breaks in the form of parsing strings into enums. I like the change, but we need to deprecate the `str` form until we can remove it in 3.0, at which point we can re-type it. And in the mean time, we can introduce a separate field which is the enum type, and will become a synonym in 3.0.

Those interfaces should all have been marked Alpha, which means no deprecation cycle.

src/citrine/resources/ingestion.py
kroenlein: Okay, I recalled that PEP completely backwards.  Weird.

tests/resources/test_workflow.py
```diff
@@ -28,7 +28,7 @@ def failed_design_workflow_data(basic_design_workflow_data):
         'status': 'FAILED',
         'status_description': 'ERROR',
         'status_detail': [
-            {'level': 'WARN', 'msg': 'Something is wrong'},
+            {'level': 'WARNING', 'msg': 'Something is wrong'},
```
kroenlein, line 31: This is still fixing a bug, since the backend does not return the string `WARN`, regardless of casing.

src/citrine/resources/status_detail.py
kroenlein: I don't know if I agree, but it's tangential to the purpose of this PR, so I'm removing it for the sake of expediency.  

(The route a user would have to go down in order to have code break feels like they've already violated our API guarantees.  But formally, yes, formerly valid data is no longer valid, and a user could have called the level setter.)

kroenlein:
> This involves a handful of interface breaks in the form of parsing strings into enums. I like the change, but we need to deprecate the `str` form until we can remove it in 3.0, at which point we can re-type it. And in the mean time, we can introduce a separate field which is the enum type, and will become a synonym in 3.0.

I've removed the update to status_detail since that was incidental to the PR.  All other changes are to Alpha endpoints, and thus are not subject to a deprecation cycle.

anoto-moniz (approved):

PR #887, opened 2023-09-18
PLA-13279: Change template name field on TemplateLink

sfriedowitz:
# Citrine Python PR

## Description 
The UI requires these new fields, so we updated the data structure internally. Given that this is not used in practice, I am proposing we do not worry about a backwards compatible change here and "lie" a bit on our versioning.

### PR Type:


src/citrine/informatics/design_spaces/hierarchical_design_space.py
```diff
@@ -25,26 +25,30 @@ class TemplateLink(Serializable["TemplateLink"]):
...
-    template_name = properties.Optional(properties.String, "name")
+    material_template_name = properties.Optional(properties.String, "material_template_name")
+    process_template_name = properties.Optional(properties.String, "process_template_name")
 
     def __init__(
             self,
             *,
             material_template: UUID,
             process_template: UUID,
-            template_name: Optional[str] = None
```
sfriedowitz, line 43: @jspeerless do you care that this is a breaking change? Nobody is really using this yet

kroenlein: I'm fine with either option (minor bias toward deprecation cycle), but it'd be pretty straight forward to support both keyword arguments, issuing deprecation warnings on `template_name` usage and feeding them both into the material_template_name field.

sfriedowitz: ill support both with an actual deprecation.

PR #901, opened 2023-11-13
as_admin listing for users and teams

robbtraister:
# Citrine Python PR

## Description 
In order to restrict the list of users to only those on shared teams, we must expose an admin variant of the users list for user management. This feature already exists for teams, but was not previously exposed in the SDK.

This PR adds a toggle to include the `?as_admin=true` query param, allowing an admin to request the entire user list (the backend still applies logic to ensure the requesting user is actually an admin).

While adding this implementation, I also discovered a discrepancy in kwarg naming between `search_params` and `additional_params` which would prevent this PR from working, so I fixed that.

### PR Type:

### DIFF
--- a/src/citrine/_rest/admin_collection.py
+++ b/src/citrine/_rest/admin_collection.py
@@ -1,0 +1,39 @@
+from typing import Iterator
+
+from citrine._rest.collection import Collection, ResourceType
+
+
+class AdminCollection(Collection[ResourceType]):
+    def list(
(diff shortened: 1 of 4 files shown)


anoto-moniz (dismissed):
Seems legit to me. Thanks Robb!

robbtraister:
thanks for your assistance on the closure (ahem, `partial`)

anoto-moniz (approved):

PR #902, opened 2023-12-19
[PLA-12982] Deprecate branch access by version ID, in favor of root ID and version

anoto-moniz:
# Citrine Python PR

## Description 
The branch version ID isn't exposed in the UI, so its use in the SDK is confusing to users, especially since there's nothing to distinguish it from the branch root ID. Additionally, there are some places in the SDK where we require the branch root ID. Also, the predictor and table collections use a root ID plus a version number. As such, the simplest interface from a user perspective is to do the same for branches.

With the branch collection, this is easy. We just deprecate referring to a branch by its version ID, in favor of using the branch root ID and version number (or the string `latest` in some places). Since the backend isn't changing, this will incur an extra GET request for most requests, as many of the APIs still operate on the version ID, so the client must look it up. But in practice, it's so small as to be negligible.

But we also have to adapt the `DesignWorkflow` object to work on the root ID and version from a user perspective, while maintaining the version ID behind the scenes. This is slightly trickier due to the interface and possible interactions (it becomes simpler when we can remove `branch_id` in 3.0). With no changes to the backend, this will incur two additional requests on some calls: one to retrieve the branch ID on request, and one to retrieve the root ID and version on response. But again, this usually doesn't result in a noticeable slowdown.

The one exception is if they're listing all the design workflows. Since it needs to look up each one's root ID and version, even though it takes minimal time for each, they add up. To alleviate this, we could tweak the backend to add branch root ID and branch version to the design workflow payload. But for now, the slowdown doesn't seem significant enough to overrule ease of use.

### PR Type:


anoto-moniz:
Full e2e run against development-e2e: https://jenkins.corp.citrine.io/job/nextgen-ITs/43111/

src/citrine/resources/branch.py
```diff
@@ -142,9 +222,12 @@ def list(self, *, per_page: int = 20) -> Iterator[Branch]:
         Returns
         -------
         Iterator[Branch]
-            Active branches in this collection.
+            All branches in this collection.
 
         """
+        warnings.warn("Beginning in the 3.0 release, this method will only list unarchived "
```
anoto-moniz, line 228: That it listed all branches conflicted with the documentation, and also doesn't match the approach taken in our other collections.

src/citrine/resources/design_workflow.py
```diff
@@ -18,10 +19,34 @@ class DesignWorkflowCollection(Collection[DesignWorkflow]):
...
+                 branch_root_id: Optional[UUID] = None,
+                 branch_version: Optional[int] = None):
         self.project_id: UUID = project_id
         self.session: Session = session
-        self.branch_id: UUID = branch_id
+        self.branch_root_id: UUID = branch_root_id
+        self.branch_version: int = branch_version
+
+    def _resolve_branch_root_and_version(self, workflow):
+        from citrine.resources.branch import BranchCollection
```
anoto-moniz, line 33: Since branches can also list the design workflows on them, importing this at the top of the file results in a circular dependency.

src/citrine/resources/experiment_datasource.py
```diff
@@ -106,7 +107,8 @@ def build(self, data: dict) -> ExperimentDataSource:
 
     def list(self, *,
              per_page: int = 100,
-             branch_id: Optional[UUID] = None,
+             branch_id: Optional[Union[UUID, str]] = None,
+             branch_version_id: Optional[Union[UUID, str]] = None,
```
anoto-moniz, line 111: The naming of this parameter was misleading, so much so that the documentation was wrong, although its use was correct.

kroenlein (approved):
That's a lot work work 🙇.  Generally looks good, but I have not done as detailed a cross check as I might.  Ping me if you'd prefer a v. detailed check.

requirements.txt
```diff
@@ -1,4 +1,4 @@
-gemd==1.15.0
+gemd==1.16.0
```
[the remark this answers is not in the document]
anoto-moniz: Correct.

requirements.txt
```diff
@@ -10,3 +10,4 @@ botocore==1.20.95
 deprecation==2.1.0
 urllib3==1.26.18
 tqdm==4.62.3
+pint==0.18
```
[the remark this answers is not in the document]
anoto-moniz: Since the constraint on pint wasn't present in gemd 1.16.0, a clean install (such as our tests do) results in 0.23 being downloaded, and thus errors. And when I tried pinning it at 0.20, I also got errors, I think because they dropped 3.7 support.

So this will go away once I do https://citrine.atlassian.net/browse/PLA-13742, which I will be starting on today.

src/citrine/informatics/workflows/design_workflow.py
```diff
@@ -62,3 +71,36 @@ def design_executions(self) -> DesignExecutionCollection:
...
+                details="Please use the branch_root_id and branch_version instead.")
+    def branch_id(self):
+        """[deprecated] Retrieve the version ID of the branch this workflow is on."""
+        return self._branch_id
+
+    @branch_id.setter
+    @deprecated(deprecated_in="2.42.0", removed_in="3.0.0",
+                details="Please set the branch_root_id and branch_version instead.")
+    def branch_id(self, value):
+        self._branch_id = value
```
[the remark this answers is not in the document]
anoto-moniz: Hmmmmmmmm...I think this is an oversight on my part. They should still be able to set `branch_id` and have the exact same impact as before. So yes, setting it should blank `branch_root_id` and `branch_version`. And when registering or updating, we should use it if the root and version aren't set, and emit a deprecation warning.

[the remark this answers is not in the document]
anoto-moniz: Ah, I see where I got crossed up. It respects it for `update`, but not for `register`.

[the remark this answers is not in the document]
anoto-moniz: Ah, actually. I still think setting `branch_id` should mark the root ID and version as `None`. But note that it's actually handled properly in the collection. That is, when updating, it's used if `branch_root_id` and `branch_version` are `None`. And when registering, the version ID should come from the collection, not the payload.

The issue is that I mistakenly broke that element of the collection. I removed the `branch_id` field of the collection, instead of deprecating it. So I need to add it back, then handle it properly in the collection methods.

src/citrine/resources/branch.py
```diff
@@ -1,18 +1,25 @@
...
 from citrine._session import Session
+from citrine._utils.functions import migrate_deprecated_argument
 from citrine.exceptions import NotFound
 from citrine.resources.data_version_update import BranchDataUpdate, NextBranchVersionRequest
 from citrine.resources.design_workflow import DesignWorkflowCollection
 from citrine.resources.experiment_datasource import (ExperimentDataSourceCollection,
                                                      ExperimentDataSource)
 
 
+LATEST_VER = "latest"  # Refers to the most recently created branch version.
```
kroenlein, line 20: I'm presuming this is intended to be importable as a keyword.  If this is wrong, we should add a leading underscore.

There's a similar constant defined for "latest" in predictor.py.  Are these programmatically tied on the back end?

tests/resources/test_branch.py
```diff
@@ -9,8 +9,9 @@
 from citrine.exceptions import NotFound
 from citrine.resources.data_version_update import NextBranchVersionRequest, DataVersionUpdate, BranchDataUpdate
 from citrine.resources.branch import Branch, BranchCollection
```
kroenlein, line 11:
```suggestion
from citrine.resources.branch import Branch, BranchCollection, LATEST_VER
```
Remove hard-coded string.  Several suggestions to follow.

tests/resources/test_branch.py
```diff
@@ -426,16 +563,277 @@ def test_get_by_root_id(session, collection, branch_path):
...
-    with pytest.raises(NotFound) as exc:
-        collection.get_by_root_id(branch_root_id=root_id)
+    with pytest.deprecated_call():
+        with pytest.raises(NotFound) as exc:
+            collection.get_by_root_id(branch_root_id=root_id)
 
     # Then
     assert str(root_id) in str(exc)
     assert "branch root" in str(exc).lower()
     assert "latest" in str(exc).lower()
```
kroenlein, line 579:
```suggestion
    assert LATEST_VER.lower() in str(exc).lower()
```

tests/resources/test_experiment_datasource.py
```diff
@@ -91,25 +91,49 @@ def test_build(collection, erds_dict):
...
+    version_id = uuid.uuid4()
 
+    session.set_response({"response": []})
+
+    list(collection.list())
+    list(collection.list(branch_version_id=version_id))
+    list(collection.list(version=4))
+    list(collection.list(version="latest"))
+    list(collection.list(branch_version_id=version_id, version=12))
+    list(collection.list(branch_version_id=version_id, version="latest"))
```
kroenlein, line 103:
```suggestion
    list(collection.list(branch_version_id=version_id, version=LATEST_VER))
```

tests/resources/test_experiment_datasource.py
```diff
@@ -91,25 +91,49 @@ def test_build(collection, erds_dict):
...
+
+    assert session.calls == [
+        FakeCall(method='GET', path=erds_base_path, params={'per_page': 100, 'page': 1}),
+        FakeCall(method='GET', path=erds_base_path, params={'per_page': 100, "branch": str(version_id), 'page': 1}),
+        FakeCall(method='GET', path=erds_base_path, params={'per_page': 100, "version": 4, 'page': 1}),
+        FakeCall(method='GET', path=erds_base_path, params={'per_page': 100, "version": "latest", 'page': 1}),
+        FakeCall(method='GET', path=erds_base_path, params={'per_page': 100, "branch": str(version_id), "version": 12, 'page': 1}),
+        FakeCall(method='GET', path=erds_base_path, params={'per_page': 100, "branch": str(version_id), "version": "latest", 'page': 1})
```
kroenlein, line 111:
```suggestion
        FakeCall(method='GET', path=erds_base_path, params={'per_page': 100, "branch": str(version_id), "version": LATEST_VER, 'page': 1})
```

tests/resources/test_experiment_datasource.py
```diff
@@ -91,25 +91,49 @@ def test_build(collection, erds_dict):
...
     branch_id = uuid.uuid4()
 
     session.set_response({"response": []})
     
     list(collection.list())
-    list(collection.list(branch_id=branch_id))
+    with pytest.deprecated_call():
+        list(collection.list(branch_id=branch_id))
     list(collection.list(version=4))
     list(collection.list(version="latest"))
```
kroenlein, line 124:
```suggestion
    list(collection.list(version=LATEST_VER))
```

tests/resources/test_experiment_datasource.py
```diff
@@ -91,25 +91,49 @@ def test_build(collection, erds_dict):
...
-    list(collection.list(branch_id=branch_id, version=12))
-    list(collection.list(branch_id=branch_id, version="latest"))
+    with pytest.deprecated_call():
+        list(collection.list(branch_id=branch_id, version=12))
+    with pytest.deprecated_call():
+        list(collection.list(branch_id=branch_id, version="latest"))
```
kroenlein, line 128:
```suggestion
        list(collection.list(branch_id=branch_id, version=LATEST_VER))
```

tests/resources/test_experiment_datasource.py
```diff
@@ -91,25 +91,49 @@ def test_build(collection, erds_dict):
...
 
     assert session.calls == [
         FakeCall(method='GET', path=erds_base_path, params={'per_page': 100, 'page': 1}),
-        FakeCall(method='GET', path=erds_base_path, params={'per_page': 100, "branch": branch_id, 'page': 1}),
+        FakeCall(method='GET', path=erds_base_path, params={'per_page': 100, "branch": str(branch_id), 'page': 1}),
         FakeCall(method='GET', path=erds_base_path, params={'per_page': 100, "version": 4, 'page': 1}),
         FakeCall(method='GET', path=erds_base_path, params={'per_page': 100, "version": "latest", 'page': 1}),
```
kroenlein, line 134:
```suggestion
        FakeCall(method='GET', path=erds_base_path, params={'per_page': 100, "version": LATEST_VER, 'page': 1}),
```

tests/resources/test_experiment_datasource.py
```diff
@@ -91,25 +91,49 @@ def test_build(collection, erds_dict):
...
-        FakeCall(method='GET', path=erds_base_path, params={'per_page': 100, "branch": branch_id, "version": 12, 'page': 1}),
-        FakeCall(method='GET', path=erds_base_path, params={'per_page': 100, "branch": branch_id, "version": "latest", 'page': 1})
+        FakeCall(method='GET', path=erds_base_path, params={'per_page': 100, "branch": str(branch_id), "version": 12, 'page': 1}),
+        FakeCall(method='GET', path=erds_base_path, params={'per_page': 100, "branch": str(branch_id), "version": "latest", 'page': 1})
```
kroenlein, line 136:
```suggestion
        FakeCall(method='GET', path=erds_base_path, params={'per_page': 100, "branch": str(branch_id), "version": LATEST_VER, 'page': 1})
```

src/citrine/resources/branch.py
```diff
@@ -100,38 +108,110 @@ def build(self, data: dict) -> Branch:
...
+        if uid:
+            warnings.warn("Retrieving a branch by its version ID using .get() is deprecated. "
+                          "Please use .get_by_version_id().",
+                          DeprecationWarning)
+
+            if root_id:
+                raise ValueError("Please provide exactly one: the version ID or root ID.")
+            return self.get_by_version_id(version_id=uid)
+        elif root_id:
+            version = version or LATEST_VER
```
kroenlein, line 177:
```suggestion
            if version is None:
                version = LATEST_VER
```
The original behaves incorrectly with empty string.

src/citrine/resources/branch.py
```diff
@@ -167,43 +250,109 @@ def list_archived(self, *, per_page: int = 20) -> Iterator[Branch]:
...
+            The version of the branch. If provided, must either be a positive integer, or "latest".
+            Defaults to "latest".
 
         """
+        if uid:
+            warnings.warn("Archiving a branch by its version ID is deprecated. "
+                          "Please use its root ID and version number.",
+                          DeprecationWarning)
+        elif root_id:
+            version = version or LATEST_VER
```
kroenlein, line 305:
```suggestion
            if version is None:
                version = LATEST_VER
```

src/citrine/resources/branch.py
```diff
@@ -167,43 +250,109 @@ def list_archived(self, *, per_page: int = 20) -> Iterator[Branch]:
...
+            The version of the branch. If provided, must either be a positive integer, or "latest".
+            Defaults to "latest".
 
         """
+        if uid:
+            warnings.warn("Restoring a branch by its version ID is deprecated. "
+                          "Please use its root ID and version number.",
+                          DeprecationWarning)
+        elif root_id:
+            version = version or LATEST_VER
```
kroenlein, line 341:
```suggestion
            if version is None:
                version = LATEST_VER
```

src/citrine/resources/branch.py
```diff
@@ -231,9 +387,26 @@ def update_data(self,
...
+                warnings.warn("Updating a branch's data by its version ID is deprecated. "
+                              "Please use its root ID and version number.",
+                              DeprecationWarning)
+                branch = self.get_by_version_id(version_id=branch)
+            root_id = branch.root_id
+            version = branch.version
+        elif not root_id:
+            raise ValueError("Please provide exactly one: the version ID or root ID.")
+
+        version = version or LATEST_VER
```
kroenlein, line 408:
```suggestion
        if version is None:
            version = LATEST_VER
```

src/citrine/resources/branch.py
```diff
@@ -244,32 +417,57 @@ def update_data(self,
...
 
         """
+        if uid:
+            warnings.warn("Retrieving data updates for a branch by its version ID is deprecated. "
+                          "Please use its root ID and version number.",
+                          DeprecationWarning)
+            if root_id:
+                raise ValueError("Please provide exactly one: the version ID or root ID.")
+        elif root_id:
+            version = version or LATEST_VER
```
kroenlein, line 458:
```suggestion
            if version is None:
                version = LATEST_VER
```

src/citrine/resources/design_workflow.py
```diff
@@ -18,10 +20,44 @@ class DesignWorkflowCollection(Collection[DesignWorkflow]):
     _resource = DesignWorkflow
     _api_version = "v2"
 
-    def __init__(self, project_id: UUID, session: Session, branch_id: UUID = None):
+    def __init__(self,
+                 project_id: UUID,
+                 session: Session,
+                 branch_id: Optional[UUID] = None,
+                 branch_root_id: Optional[UUID] = None,
+                 branch_version: Optional[int] = None):
```
kroenlein, line 28: It doesn't make sense that these would all be positional arguments.  Allowing three is already outside [repo coding style](https://github.com/CitrineInformatics/citrine-python/blob/main/CONTRIBUTING.md#codestyle).  If you're migrating, we should probably do something like:
```suggestion
    def __init__(self,
                 project_id: UUID,
                 session: Session,
                 positional_branch_id: Optional[UUID] = None,
                 *,
                 branch_id: Optional[UUID] = None,
                 branch_root_id: Optional[UUID] = None,
                 branch_version: Optional[int] = None):
```
and throw a deprecation warning when the `positional_branch_id` argument is not None.  Alternatively, if you intend to support positional arguments for `branch_root_id` & `branch_version` and `branch_id` and `branch_root_id` are functionally interchangeable during the migration,
```suggestion
    def __init__(self,
                 project_id: UUID,
                 session: Session,
                 branch_root_id: Optional[UUID] = None,
                 branch_version: Optional[int] = None):
                 *,
                 branch_id: Optional[UUID] = None,
```
and deprecation is issued when branch_id is not None.

src/citrine/resources/design_workflow.py
```diff
@@ -18,10 +20,44 @@ class DesignWorkflowCollection(Collection[DesignWorkflow]):
...
         self.project_id: UUID = project_id
         self.session: Session = session
-        self.branch_id: UUID = branch_id
+
+        if branch_id:
```
kroenlein, line 32:
```suggestion
        if branch_id is not None:
```

src/citrine/resources/design_workflow.py
```diff
@@ -18,10 +20,44 @@ class DesignWorkflowCollection(Collection[DesignWorkflow]):
...
+            warnings.warn("Constructing the design workflow interface by its parent branch's "
+                          "version ID is deprecated. Please use its root ID and version number.",
+                          DeprecationWarning)
+            if branch_root_id:
```
kroenlein, line 36:
```suggestion
            if branch_root_id is not None:
```

src/citrine/resources/design_workflow.py
```diff
@@ -42,23 +78,30 @@ def register(self, model: DesignWorkflow) -> DesignWorkflow:
...
 
         """
-        if self.branch_id is None:
-            # There are a number of contexts in which hitting design workflow endpoints without a
-            # branch ID is valid, so only this particular usage is deprecated.
-            msg = ('A design workflow must be created with a branch. Please use'
-                   'branch.design_workflows.register() instead of '
-                   'project.design_workflows.register().')
-            raise RuntimeError(msg)
+        if self.branch_root_id is None or self.branch_version is None:
```
kroenlein, line 81: Does is ever make sense to have one set and the other not?

kroenlein (approved):
One significant question about a method signature and some falsity check best practice questions.

src/citrine/resources/branch.py
anoto-moniz: Yes, it is intended to be public, like the similar one in predictors. But no, they're not tied together on the backend.

tests/resources/test_branch.py
anoto-moniz: This would be bad practice. It would mean if we screwed up the string in `branch.LATEST_VER`, the tests wouldn't catch it.

That being said, it's not the worst idea to define an independent constant in the test file.

src/citrine/resources/branch.py
anoto-moniz: No it doesn't. If the user provides an empty string, we should treat that the same as `None`.

tests/resources/test_branch.py
anoto-moniz: This would be bad practice. It would mean if we screwed up the string in branch.LATEST_VER, the tests wouldn't catch it.

That being said, it's not the worst idea to define an independent constant in the test file.

src/citrine/resources/design_workflow.py
anoto-moniz: I'm not sure if any of our other collection constructors have keyword-only parameters. But maybe they should?

In any case, sure, I'll make the new args kw-only. But since we're deprecating `branch_id`, I'm going to leave it as is.

src/citrine/resources/design_workflow.py
anoto-moniz: This just depends on how aggressively we want to raise the deprecation warning. My thought is that since other things will catch it and fail downstream, I'm not concerned. Plus, if users are constructing their own `DesignWorkflowCollection`, they're already using the SDK in an unorthodox way.

src/citrine/resources/design_workflow.py
anoto-moniz: Similar to above.

src/citrine/resources/design_workflow.py
anoto-moniz: No. Programmatically, it can happen, but in the context of the collection, we always want them both set.

kroenlein (approved):
I would have preferred we were importing the constant into the tests rather than hard-coded, but that's personal preference and what's here works fine.

anoto-moniz:
> I would have preferred we were importing the constant into the tests rather than hard-coded, but that's personal preference and what's here works fine.

See my comment here: https://github.com/CitrineInformatics/citrine-python/pull/902#discussion_r1434009315

PR #904, opened 2023-12-21
[PLA-13149] Bump dependency versions.

anoto-moniz:
# Citrine Python PR

## Description 
Bumped all dependencies to their latest version, with a few exceptions.

pytest-flake8 and pydocstyle don't appear to be in use, so they were dropped instead.

botocore only needs to be in our requirements file if we need to pin it separately from boto3, which we currently do not.

boto3 1.34.5 imposes some restrictions on urllib3. If Python is < 3.10, it must be pre-2.0. But even if Python is more recent, it can only support up to (but not including) urllib3 2.1. We've set the pins accordingly.

The latest version of Sphinx available for Python 3.8 is 7.1.2. So that version specific pin was made, and it was pinned at 7.2.6 for everything else.

### PR Type:


anoto-moniz:
https://jenkins.corp.citrine.io/job/nextgen-ITs/43125/

anoto-moniz:
https://jenkins.corp.citrine.io/job/nextgen-ITs/43135/

anoto-moniz:
nextgen IT runs:
- development-e2e: https://jenkins.corp.citrine.io/job/nextgen-ITs/43189/
- stage: https://jenkins.corp.citrine.io/job/nextgen-ITs/43190/

test_requirements.txt
```diff
@@ -1,12 +1,17 @@
...
+pytz==2023.3.post1
+requests-mock==1.11.0
+sphinx-rtd-theme==2.0.0
+sphinxcontrib-apidoc==0.4.0
+
+pandas==2.0.3; python_version < "3.9"
+sphinx==7.1.2; python_version < "3.9"
+
+pandas==2.1.4; python_version >= "3.9"
+sphinx==7.2.6; python_version >= "3.9"
```
kroenlein, line 17: I hadn't seen that before.  Solves problems.

setup.py
```diff
@@ -24,25 +24,24 @@
       install_requires=[
           "requests>=2.31.0,<3",
           "pyjwt>=2,<3",
-          "arrow>=0.15.4,<2",
-          "gemd>=1.15.0,<2",
-          "boto3>=1.17.93,<2",
-          "botocore>=1.20.95,<2",
+          "arrow>=1.3.0,<2",
+          "gemd>=1.16.7,<2",
```
kroenlein, line 28: Is this just an instinctive set-to-latest, or is there a feature-based motivation for choosing this one?

src/citrine/resources/ingestion.py
```diff
@@ -408,7 +410,7 @@ def status(self) -> IngestionStatus:
         """
         if self.raise_errors:
             raise JobFailureError(
-                message=f"Ingestion creation failed: self.errors",
+                message=f"Ingestion creation failed: {self.errors}",
```
kroenlein, line 413: Good catch.

src/citrine/resources/ingestion.py
```diff
@@ -1,14 +1,16 @@
...
+
 from citrine._rest.collection import Collection
 from citrine._rest.resource import Resource
 from citrine._serialization import properties
 from citrine._session import Session
 from citrine.exceptions import CitrineException, BadRequest
 from citrine.jobs.job import JobSubmissionResponse, JobFailureError, _poll_for_job_completion
 from citrine.resources.api_error import ApiError, ValidationError
-from gemd.enumeration.base_enumeration import BaseEnumeration
+from citrine.resources.file_link import FileLink
```
kroenlein, line 13: Is this introducing a circular reference risk?  I think there's a clear dependency order on these two packages, but just want to highlight the question.

kroenlein (approved):
A few general questions, and I think there's a need for a coherent plan around defining lowest supported version in packages, but nothing blocking.

setup.py
anoto-moniz: Just an instinctive bumping, since I feel like nudging users towards newer releases sooner rather than later is good practice. But please let me know if you'd rather I left them as is, since in these (and most) cases, it's just a minor bump, so should be fine left alone.

src/citrine/resources/ingestion.py
anoto-moniz: Ah, you mean moving them around? I don't know about a risk, but I know it ran fine for me locally. I moved it to adhere to PEP 8 import grouping, but if it causes problems, I'm happy to move it back.

setup.py
```diff
@@ -24,25 +24,24 @@
...
+          "boto3>=1.34.12,<2",
           "deprecation>=2.1.0,<3",
           "urllib3>=1.26.18,<3",
-          "tqdm>=4.62.3,<5"
+          "tqdm>=4.66.1,<5"
```
kroenlein, line 32: Python 3.12 happily installs tqdm 1.0, and the feature we use (`tqdm.auto`) came in with 4.27.0.  I feel like we don't need to have a strong opinion on this one.
```suggestion
          "tqdm>=4.27.0,<5"
```

setup.py
```diff
@@ -24,25 +24,24 @@
       install_requires=[
           "requests>=2.31.0,<3",
           "pyjwt>=2,<3",
-          "arrow>=0.15.4,<2",
-          "gemd>=1.15.0,<2",
-          "boto3>=1.17.93,<2",
-          "botocore>=1.20.95,<2",
+          "arrow>=1.3.0,<2",
```
kroenlein, line 27: I'd grant a little more grace on minimum version here, but requiring >= v1 would certainly make sense.

kroenlein (approved):
I'd like to be a bit more conservative on some of the libraries, and I'd like to see some of these dependencies just gone.

test_requirements.txt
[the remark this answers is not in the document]
anoto-moniz: I don't want new developers to come into using it for the first time to find it broken due to some incompatibility we didn't notice.

test_requirements.txt
```diff
@@ -1,12 +1,17 @@
...
+flake8-docstrings==1.7.0
+mock==5.1.0
+pytest==7.4.4
+pytest-cov==4.1.0
+pytz==2023.3.post1
+requests-mock==1.11.0
+sphinx-rtd-theme==2.0.0
+sphinxcontrib-apidoc==0.4.0
+
+pandas==2.0.3; python_version < "3.9"
```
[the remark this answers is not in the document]
anoto-moniz: Removed it as part of the builders drop at your suggestion, as well as a follow up commit to rewrite that test since I overlooked it initially.

test_requirements.txt
```diff
@@ -1,12 +1,17 @@
...
+sphinx==7.1.2; python_version < "3.9"
```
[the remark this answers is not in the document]
anoto-moniz: Fair enough, I'll just pin it to 7.1.2.

setup.py
anoto-moniz: I don't feel super strongly on these lower bounds, so sure, I'll drop it to 1.0.

setup.py
anoto-moniz: Sure.

PR #916, opened 2024-01-16
[PLA-13814] Switch from Travis to GitHub Actions.

anoto-moniz:
# Citrine Python PR

## Description

We've had some issues with Travis CI, most recently permissions syncing. In the past few years, GitHub Actions has become pretty mature, and it has documentation that's generally far better than Travis. Not to mention its more tightly integrated with GitHub for obvious reasons. And the general benefits of shrinking the number of services we depend on.

This introduces three workflows, which will be kicked off automatically. One runs the tests and linting and such on PR creation against main, one builds and deploys to PyPI when a release is created, and one builds and deploys the docs when a release is created.

### PR Type:


.github/actions/initialize/action.yml
anoto-moniz, line 1: This is a convenience action used by many of our jobs and workflows to initialize the environment.

CONTRIBUTING.md
```diff
@@ -65,32 +65,6 @@ pytest tests/ --cov=src/
 It is not uncommon to have path issues when running pytest from the command line.
 Ensure that your $PATH variable contains the directory with the citrine-python repo.
 
-#### Docker
```
anoto-moniz, line 68: As far as I can tell, no one is using this, given that the corresponding files haven't been changed in 5 years.

requirements.txt
```diff
@@ -2,7 +2,7 @@ gemd==1.16.0
 requests==2.31.0
 pyjwt==2.4.0
 arrow==0.15.4
-sphinx==4.3.0
+sphinx==5.0.0
```
anoto-moniz, line 5: The GitHub Actions build and deploy actions require Sphinx 5.0. The full upgrade will be in the 3.0 release.

.github/actions/initialize/action.yml
```diff
@@ -0,0 +1,19 @@
...
+      shell: bash
+    - name: Install runtime dependencies
+      run: python -m pip install --only-binary ':all:' -r requirements.txt --upgrade
+      shell: bash
+    - name: Install test dependencies
+      run: python -m pip install --only-binary ':all:' -r test_requirements.txt --upgrade
+      shell: bash
+    - name: Install citrine
+      run: python -m pip install --only-binary ':all:' -e . --upgrade
+      shell: bash
```
kroenlein, line 19: What is the `--upgrade` supposed to accomplish?  All the entries in requirements.txt are pinned, so that flag has no impact.

.github/workflows/pr-tests.yml
```diff
@@ -0,0 +1,47 @@
...
+      - name: Initialize the environment
+        uses: ./.github/actions/initialize
+      - name: Deprecated check
+        run: derp src/ src/citrine/__version__.py
+  run-tests:
+    name: Execute unit tests
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.8", "3.9", "3.10", "3.11"]
```
kroenlein, line 41: Is there no way to have a low-library-version / high-library-version dimension on the test matrix?  We're doing that now because it has caught problems with multiple libraries in the past.

.github/workflows/build-docs.yml
```diff
@@ -0,0 +1,39 @@
...
+jobs:
+  deploy:
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Initialize the environment
+        uses: ./.github/actions/initialize
```
kroenlein, line 28: Would it make sense to have a different build for the documentation build?  There's no real purpose to having Sphinx et al. installed during testing if the doc build it is a separate step.

scripts/run_tests.sh
kroenlein, line 1: Running this file is in my git commit hooks.  What's the value is removing it?  Or is this an attempt at removing redundancy?

scripts/validate_version_bump.py
```diff
@@ -18,7 +18,7 @@ def main():
         raise ValueError(f"Couldn't extract version from {version_path}") from e
 
     try:
-        with popen(f"git show main:src/citrine/__version__.py", mode="r") as fh:
+        with popen(f"git show origin/main:src/citrine/__version__.py", mode="r") as fh:
```
kroenlein, line 21: Was this a requirement to get actions to run?  Seems weird you'd refer to a remote branch, especially if you don't precede it with a fetch.

CONTRIBUTING.md
```diff
@@ -196,15 +170,16 @@ PR descriptions should describe the motivation and context of the code changes i
...
-
 As it can be easy to forget to verify these prior to pushing, it's possible to use [git hooks](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks) to enforce compliance during normal workflows.
 Consider editing `.git/hooks/pre-commit` or `.git/hooks/pre-push` (or adding them and marking them as executable: `chmod +x <file>`).
 For example, you could set your local `.git/hooks/pre-commit` to be
 ```shell
-scripts/run_tests.sh
+python $REPO_DIR/scripts/validate_version_bump.py                        || exit 1;
+flake8 $REPO_DIR/src                                                          || exit 1;
+derp $REPO_DIR/src $REPO_DIR/src/citrine/__version__.py                       || exit 1;
+pytest --cov=src --cov-report term-missing:skip-covered --cov-config=tox.ini --no-cov-on-fail --cov-fail-under=100  tests
```
kroenlein, line 180: What is the value is removing this?

kroenlein (commented):
I don't know much about github-actions limitations, but it seems like you've removed some valuable things in addition to the cruft.  Good that we're migrating, and I look forward to copying a bunch of this into gemd-python.

.github/actions/initialize/action.yml
anoto-moniz: The safety of not knowing if some of the more common packages we use (e.g. `requests`, `pytest`) are already installed at different versions.

.github/workflows/pr-tests.yml
anoto-moniz: I presume you're talking about the `UPGRADE` thing?

As we discussed in Slack, I've yet to be shown any value it's given us. The issues you've pointed me to involve us relying on pre-1.0 libraries, and them breaking something on a minor version bump. Our setup.py should probably constrain them tighter to avoid those problems.

If we want to test for stuff like that, then we can set up some other build, but we shouldn't block PRs for unsupported library versions.

.github/workflows/build-docs.yml
anoto-moniz: You mean breaking out some file like `doc_requirements.txt`? That's not a terrible idea, although I don't see the harm of it being installed.

scripts/run_tests.sh
anoto-moniz: By breaking out the things it checks into separate jobs we get their results independent of each other. So we can still see test results if there are PEP 8 issues, for example. So I removed this script in order to avoid redundancy. But if you find it useful despite it not being used by the jobs, I'm happy to leave it as is.

scripts/validate_version_bump.py
anoto-moniz: Yes. I think the fetch would only be needed for the local branch, right? And referring to the remote will check the actual live value.

CONTRIBUTING.md
anoto-moniz: Same as above. If we leave the script, I guess we can revert this doc, although I like people knowing exactly what's being run. For example, it wasn't until I created the `release/3.0` branch that I learned about `derp`.

scripts/validate_version_bump.py
anoto-moniz: I did end up adding `fetch`, because sometimes it worked and sometimes it didn't. 🤷

.github/workflows/build-docs.yml
kroenlein: Minor optimization.  Install 5x, use once is slower than install once, use once.  Non-blocking / future-work.

anoto-moniz: This is only run once, using whatever their default version of Python is.

scripts/run_tests.sh
kroenlein: Ideally, it'd be nice to have it not redundant, but given the two options, I prefer a redundant file that's still there.

.github/actions/initialize/action.yml
anoto-moniz: (For posterity, since we chatting in Slack)

I tried it again locally, and realized my testing was screwy. You are correct, it's not needed except for when we don't specify the package version, such as the pip upgrade line.

scripts/run_tests.sh
anoto-moniz: I've added it back at your request.

kroenlein (approved):
LGTM

PR #919, opened 2024-01-30
Surface docs warnings.

anoto-moniz:
# Citrine Python PR

In order to keep the docs build messages useful, we should run a check on PR to surface all warnings. With the release of 3.0, they should be completely clean, so a failure of this job will be meaningful. Even such, we'll probably keep it non-required for now so as to not add unnecessarily to the development burden.

## Description 

### PR Type:

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -1,0 +1,44 @@
+name: PR Checks
+
+on:
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  check-version:
+    name: Check version bumped
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


kroenlein (commented):
Are you removing version bump, linking, and deprecation check from required?  I agree with the docs check idea, but not with removing those as gates.

anoto-moniz:
> Are you removing version bump, linking, and deprecation check from required? I agree with the docs check idea, but not with removing those as gates.

Ah, my bad, I should be clear about that. Short version is no, I'm not. Those will still be required.

Turns out you don't mark a job as required in the GitHub Action definition. Rather, it's in the GitHub web interface under the branch restrictions. So this adds the definition, and we just won't mark it required.

PR #923, opened 2024-02-27
Remove explicit, disconnected manifests

kroenlein:
# Citrine Python PR

## Description 
This PR removes the `_attr` field from multiple objects, and updates the bits that depended on that concept to use the generic ser/deser capabilities rather than a potentially decoupled introspection.  

Some minor related changes came along for the ride.  This should have no interface implications, assuming no-one has been using the private methods.

### PR Type:


src/citrine/gemtables/columns.py
```diff
@@ -57,16 +56,10 @@ class Column(PolymorphicSerializable['Column']):
...
     def __eq__(self, other):
-        try:
-            return all([
-                self.__getattribute__(key) == other.__getattribute__(key) for key in self._attrs()
-            ])
-        except AttributeError:
+        if isinstance(other, Serializable):
+            return self.dump() == other.dump()
+        else:
             return False
```
kroenlein, line 63: This pattern is generally repeated.  Not fundamentally opposed to dropping the isinstance test for a try instead on the potential `AttributeError`.

src/citrine/informatics/predictor_evaluation_metrics.py
```diff
@@ -25,23 +25,15 @@ class PredictorEvaluationMetric(PolymorphicSerializable["PredictorEvaluationMetr
     Abstract type that returns the proper type given a serialized dict.
     """
 
-    def _attrs(self) -> List[str]:
-        return ["typ"]
-
     def __hash__(self):
-        return hash(self.__getattribute__("typ"))
+        return hash(self.typ)
```
kroenlein, line 29: Identical behavior with no introspection.  Necessary because PredictorEvaluationMetrics get stored in a set.

It seems like this should maybe be `hash(repr(self))` instead, since it seems like you should be able to have non-collision between `CoverageProbability`s with different levels, but that would technically be a breaking change.

src/citrine/informatics/predictor_evaluator.py
```diff
@@ -23,28 +24,24 @@ def get_type(cls, data) -> Type[Serializable]:
...
+        if isinstance(other, Serializable):
+            return self.dump() == other.dump()
+        else:
             return False
 
     @property
+    @abstractmethod
     def responses(self) -> Set[str]:
         """Responses to compute metrics for."""
-        raise NotImplementedError  # pragma: no cover
```
kroenlein, line 40: Functionally equivalent, minus a change in exception type.  If that feels breaking, then we can do a minor bump.

tests/informatics/test_predictor_evaluation_result.py
```diff
@@ -38,21 +38,32 @@ def test_holdout_serde(example_holdout_result, example_holdout_result_dict):
...
 
+def test_holdout_set_evaluator(example_holdout_result, example_holdout_evaluator_dict):
+    args = example_holdout_evaluator_dict
+    del args["type"]
+    expected = HoldoutSetEvaluator(**args)
+    assert expected.responses == set(example_holdout_evaluator_dict['responses'])
+    assert example_holdout_result.evaluator == expected
+    assert example_holdout_result.evaluator != 0  # make sure eq does something for mismatched classes
+
+
```
kroenlein, line 59: This type was never formally tested, just implicitly through the workflow reports.

kroenlein:
Tests failed on the first execution for some reason: https://github.com/CitrineInformatics/citrine-python/actions/runs/8057531733/attempts/1?pr=923
No changes and they passed round 2.  I don't like the idea of a flaky test, but it's not obvious to me what might have happened.

src/citrine/informatics/predictor_evaluator.py
```diff
@@ -23,28 +24,36 @@ def get_type(cls, data) -> Type[Serializable]:
...
-            ])
-        except AttributeError:
+        if isinstance(other, Serializable):
+            self_dict = self.dump()
+            other_dict = other.dump()
+
+            self_dict['responses'] = set(self_dict.get('responses', []))
+            self_dict['metrics'] = frozenset(
+                frozenset((k, v) for k, v in dct.items()) for dct in self_dict.get('metrics', [])
+            )
```
kroenlein, line 35: A little gross and could use a regular set instead of a frozenset on the outside, but structurally sound.  Can't just use the metrics hashing function because CoverageProbability compares the levels even though its hashing function ignores that field.

src/citrine/gemtables/columns.py
anoto-moniz: Meh, this is fine by me.

tests/informatics/test_predictor_evaluation_result.py
```diff
@@ -38,21 +38,32 @@ def test_holdout_serde(example_holdout_result, example_holdout_result_dict):
...
 def test_check_rmse(example_cv_result, example_rmse_metrics):
     assert example_cv_result["saltiness"]["rmse"].mean == example_rmse_metrics["mean"]
     assert example_cv_result["saltiness"][RMSE()].standard_error == example_rmse_metrics["standard_error"]
     # check eq method does something
     assert example_cv_result["saltiness"][RMSE()] != 0
     with pytest.raises(TypeError):
-        foo = example_cv_result["saltiness"][0]
+        _ = example_cv_result["saltiness"][0]
```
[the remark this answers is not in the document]
anoto-moniz: Why bother retrieving the value at all then?

src/citrine/informatics/predictor_evaluator.py
anoto-moniz: According to the docs, `@abstractmethod` doesn't mean anything unless the class's metaclass is `ABCMeta`. I don't think that's the case here, so that decorator means nothing?

If the intent is for the metaclass to be `ABCMeta`, then I disagree they're functionally equivalent, as that would prevent even instantiating an instance of the class, not only impacting the method.

kroenlein: Removed.

anoto-moniz (dismissed):
LGTM!

anoto-moniz (approved):

PR #925, opened 2024-03-12
Augment Factories to be more standard and closer to API responses

kroenlein:
# Citrine Python PR

## Description 
This PR attempts to bring our FactoryBoy usage closer to standard, with an expectation that follow-ups will be necessary.  If we chose to migrate away from Factory Boy, these correction should be helpful in building out whatever mocking system we migrate to.  There is a substantial amount of work remaining if we want to clean up tests and fix some unnecessarily fragile choices.

* Add many fields that were present in API returns but not reflected in factory definitions
* (Almost) remove all non-standard Factory usage
* Added providers for Citrine-relevant data types
* Uncovered bug in data round-trip serialization due to nanosecond-scale floating point issues
* Minor formatting & best practice corrections

### PR Type:


tests/utils/factories.py
```diff
@@ -4,36 +4,142 @@
...
+        if elements is None or len(elements) == 0:
+            elements = list(EmpiricalFormula.all_elements())  # Must be Sequence
+        if count is None:
+            count = self.generator.random.randrange(1, 5)
+        components = sorted(self.generator.random.sample(elements, count))
+        # Use weights to bias toward looking more real-ish
+        amounts = self.generator.random.choices([1, 2, 3, 4, 5], weights=[40, 40, 10, 10, 2], k=count)
+        return "".join(f"({c}){a}" for c, a in zip(components, amounts))
+
+    def random_smiles(self) -> str:
```
kroenlein, line 41: Probably unnecessary, but took a few minutes to do something interesting.

tests/utils/factories.py
```diff
@@ -320,12 +486,48 @@ class Meta:
...
+class MaterialSpecDataFactory(factory.DictFactory):
+    uids = MaterialSpecFactory.uids
+    name = MaterialSpecFactory.name
+    tags = MaterialSpecFactory.tags
+    notes = MaterialSpecFactory.notes
+    process = factory.SubFactory(LinkByUIDDataFactory)
+    file_links = factory.List([factory.SubFactory(FileLinkDataFactory)])
+    template = factory.SubFactory(LinkByUIDDataFactory)
+    properties = []  # TODO make a PropertiesDataFactory
+    type = 'material_spec'
```
kroenlein, line 518: There are likely cleaner patterns here, particularly using [make_factory](https://factoryboy.readthedocs.io/en/stable/reference.html#factory.make_factory), but I didn't want to get into too much magic until we had a plan for moving forward leveraging the API.

tests/utils/factories.py
```diff
@@ -271,47 +415,69 @@ class DatasetDataFactory(factory.DictFactory):
...
+    class Meta:
+        model = FileLink
+
     url = factory.Faker('uri')
     filename = factory.Faker('file_name')
-    type = 'file_link'
 
 
-class MaterialRunDataFactory(factory.DictFactory):
+class ProcessTemplateFactory(factory.Factory):
```
kroenlein, line 446: Still doesn't have `audit_info` attached, which is pretty important since we create citrine-python child class with extra fields from an upstream library.

tests/utils/factories.py
```diff
@@ -17,26 +20,126 @@
...
+            "S": 4,
+            "F": 1,
+            "Cl": 1,
+            "Br": 1,
+            "I": 1
+        }
+        bonds = ['', '=', '#', '$']
+
+        elements = [x for x in element_weights]
+        weights = [element_weights[x] for x in element_weights]
```
anoto-moniz, line 70: `elements, weights = element_weights.items()`

tests/utils/factories.py
```diff
@@ -17,26 +20,126 @@
...
+
+        smiles = self.generator.random.choices(elements, weights=weights)[0]
+        remain = [valence[smiles[0]]]
+        while len(remain) != 0:
```
anoto-moniz, line 74: At the risk of reigniting our eternal disagreement:

"For sequences, (strings, lists, tuples), use the fact that empty sequences are false": https://peps.python.org/pep-0008/

`while not remain`

tests/utils/factories.py
anoto-moniz, line 1: This looks great! Feel free to cut tickets for the TODOs and assign some of them to me, especially the obviously Orion-centric stuff, like predictor data. It's low priority, sure, but could be useful to pick up when I have some time between tickets/projects.

src/citrine/_serialization/properties.py
```diff
@@ -408,7 +408,8 @@ def _deserialize(self, value) -> datetime:
         raise TypeError("{} must be an int or a string".format(value))
 
     def _serialize(self, value: datetime) -> int:
-        return int(arrow.get(value).float_timestamp * 1000)
+        # Add 100 nanoseconds to avoid floating point truncation issues from microseconds
```
anoto-moniz, line 411: What are the truncation issues to which you're referring?

anoto-moniz (approved):
Thanks for diving into this! Just some nitpicks and questions, but nothing blocking, so feel free to do what you will with them.

tests/utils/factories.py
kroenlein: Huh.  I would have thought I took the other side in this one.  Sure.

tests/utils/factories.py
kroenlein: I think that's backwards:
`ValueError: too many values to unpack (expected 2)`

src/citrine/_serialization/properties.py
kroenlein: As per the test I added, 269815509154 was being changed to 269815509153 on round trip de-serialization.  Multiplying by 1000 brought finite precision issues into the least significant spot, and `int` would then just miss when truncating.

tests/resources/test_experiment_datasource.py
```diff
@@ -47,43 +31,31 @@ def erds_base_path(collection):
...
+def test_build(collection):
+    erds_dict = ExperimentDataSourceDataFactory()
+    actual_erds: ExperimentDataSource = collection.build(erds_dict)
 
     assert str(actual_erds.uid) == erds_dict["id"]
     assert str(actual_erds.branch_root_id) == erds_dict["metadata"]["branch_root_id"]
     assert actual_erds.version == erds_dict["metadata"]["version"]
     assert str(actual_erds.created_by) == erds_dict["metadata"]["created"]["user"]
-    assert actual_erds.create_time.replace(tzinfo=None).isoformat() == erds_dict["metadata"]["created"]["time"]
+    # TODO: It'd be better to actually invoke the Datetime._serialize method
```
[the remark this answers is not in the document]
kroenlein: Not something I'd really fight over, just I hate re-implementing logic in a test that isn't actually supposed to test that logic.

tests/utils/factories.py
anoto-moniz: ...ignore me, it's early. XD

anoto-moniz: You could just do `element_weights.keys()` and `element_weights.values()`, although I'm not too adamant about that.

anoto-moniz (approved):

PR #926, opened 2024-03-13
Implement GemdQuery obejcts for table building

kroenlein:
# Citrine Python PR
https://citrine.atlassian.net/browse/PLA-13437

## Description 
This PR implements the GemdQuery object and affiliates it with TableConfig objects, as [documented](https://docs.corp.citrine.io/platform-backend/faraday/index.html#tag/GemTables/operation/upgradeTableConfig).  

@dleblanc @lbryan-citrine are to validate the implemented objects, as apparently not everything implemented on the backend is currently in use and the API docs are both incomplete and do not indicate the preferred object types.  If some objects should have been deprecated, we will eliminate them from the Python implementation to avoid a deprecation cycle.

### PR Type:


src/citrine/gemd_queries/filter.py
```diff
@@ -0,0 +1,223 @@
...
+from citrine._serialization.polymorphic_serializable import PolymorphicSerializable
+from citrine._serialization import properties
+
+__all__ = ['PropertyFilterType',
+           'RealFilter', 'IntegerFilter',
+           'MaterialClassification', 'TextSearchType',
+           ]
+
+
+class RealFilter(Serializable['RealFilter']):
```
dleblanc, line 16: Like I mentioned previously, I don't think this logic is working at present, and I'd be hesitant to expose it now.
I think the only filters we have that we know work today are the All* filters.

src/citrine/gemd_queries/gemd_query.py
```diff
@@ -0,0 +1,64 @@
...
+from citrine._serialization.serializable import Serializable
+from citrine._serialization import properties
+from citrine.gemd_queries.criteria import Criteria
+
+
+class GemdObjectType(BaseEnumeration):
+    """The style of text search to run."""
+
+    # An old defect has some old GemdQuery values stored with invalid enums
+    # The synonyms will allow invalid old values to be read, but not emitted
```
dleblanc, line 13: Nice ... what a hassle that situation is.

tests/gemd_query/test_gemd_query.py
```diff
@@ -0,0 +1,45 @@
...
+
+from citrine.gemd_queries.criteria import PropertiesCriteria
+from citrine.gemd_queries.filter import AllRealFilter
+from citrine.gemd_queries.gemd_query import GemdQuery
+
+from tests.utils.factories import GemdQueryDataFactory
+
+
+def test_gemd_query_version():
+    valid = GemdQueryDataFactory()
```
dleblanc, line 12: This uses the faker stuff below to make arbitrary GemdQueries?

dleblanc (approved):
This looks good to me.

In future I think we can look at using Tapir to generate the models we depend on (in Python). I don't have experience with it, but if it can give us useful models it could help cut down on the client generation time.

tests/gemd_query/test_gemd_query.py
kroenlein: It doesn't do complex configurations, but it does fuzz values, yes.  

I started down the path of what it would take to generate different length lists, random branching, but there was more custom there than I was comfortable with.  Particularly given the future auto-generated question.

dleblanc: I don't know what that would buy you at this level though?

kroenlein: If it were easy, it buys you what fuzzing always buys you.  Instead, I just had to make sure the tested data structure was comprehensive.

src/citrine/gemd_queries/filter.py
kroenlein: I've verified that at least the NominalRealFilter + RealFilter combo works, though the keys are all in camel case instead of snake case and if unit is the empty string, it erroneously gets dropped.  So buggy, but the results get filtered.

I'm comfortable not exposing it here if you have a high degree of confidence that there are no artifacts in the wild that use it.  Given the SDK's role in expert debugging, I don't want a table config fetch to result in a ValueError brick wall.

dleblanc: "no artifacts in the wild that use it" - well, there's no SDK users that use it, I know that!

My understanding is that the UI folks tried it, found issues, requested a simpler approach (the All* filters) and use that exclusively now.

I've asked Lynnae to craft a bug report to get them fixed (or remove them), exposing this currently I think would be a bad thing.

kroenlein: The risk of encountering objects being low and probably associated with Citrine team members, I'm going to not expose the filter classes that we are not planning on using moving forward.

dleblanc (approved):

PR #927, opened 2024-04-11
PNE-241: Supporting team datasets

pacdaemon:
# Citrine Python PR

## Description 
See [Jira Ticket](https://citrine.atlassian.net/browse/PNE-241)
The goal of this work is to support datasets at the team level and lay down the deprecation path.

### Note
This is a very immature PR which I'm creating in a very early stage of my work so I can get feedback.

### TODO
- [ ] Support dataset crud operations at team-level
- [ ] Support operating with files at team-level
- [ ] Mark project-based dataset related methods to be deprecated
- [ ] Deal with publishing and un-publishing resources, we should deprecate those.

### PR Type:


src/citrine/resources/dataset.py
```diff
@@ -380,6 +383,97 @@ def gemd_batch_delete(
         return self.gemd.batch_delete(id_list, timeout=timeout, polling_delay=polling_delay)
 
 
+class TeamDatasetCollection(Collection[Dataset]):
```
pacdaemon, line 386: @kroenlein I was temped to use the `DatasetCollections` for the team-based ones and make it parametrizable, that I can't introduce any breaking change over there, so I'm creating a new collection here.

kroenlein: Can you detail what would break by making the project_id potentially None?  If the route is going away, all traffic should be going through the team-based route.

pacdaemon: If you are thinking of adding the `team_id` as an optional parameter and making the `project_id` to be optional for the constructor DatasetCollection, I'd say that some people might be using positional parameters for constructing it. Or... what do you have in mind?

kroenlein: People aren't using either of those patterns.  The general usage is 
```
citrine = Citrine()
team = citrine.teams.get('xxx')
project = team.projects.get('xxx')
dataset = project.datasets.get('xxx')
```
Because of some weird interface choices, doing this just with constructors is actually remarkably awkward.

pacdaemon: Got it, that helps me a lot

src/citrine/resources/project.py
```diff
@@ -139,9 +141,12 @@ def generative_design_executions(self) -> GenerativeDesignExecutionCollection:
         return GenerativeDesignExecutionCollection(project_id=self.uid, session=self.session)
 
     @property
+    @deprecation.deprecated(deprecated_in="3.2.0", removed_in="4.0.0",
```
pacdaemon, line 144: @kroenlein how does it look? Is the right way of deprecating this?. I gues we'll need to jump to version 4 as removing the dataset collection from the project entity is a breaking change

kroenlein: The deprecation syntax is correct, but given that projects are owned by teams, is there a reason to remove this route?  i.e., is there a failure mode where in the new paradigm, a user with a `project` object wouldn't get the right dataset if they said `project.datasets.list()` and that routed to `teams/{team}/datasets`?

kroenlein (commented):
We should talk about the introduction of a Left / Right paradigm into the library.

src/citrine/resources/dataset.py
```diff
@@ -393,14 +397,18 @@ class DatasetCollection(Collection[Dataset]):
...
-    _path_template = 'projects/{project_id}/datasets'
+    _path_template = None
     _individual_key = None
     _collection_key = None
     _resource = Dataset
 
-    def __init__(self, project_id: UUID, session: Session):
-        self.project_id = project_id
+    def __init__(self, project_or_team: Either[UUID, UUID], session: Session):
+        self.project_or_team = project_or_team
```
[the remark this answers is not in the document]
pacdaemon: I thought you said that was fine here because nobody was using that syntax.

https://github.com/CitrineInformatics/citrine-python/pull/927#discussion_r1561573166

By that moment I was thinking on creating a new collection for the team datasets to not alter the constructor. What would be the best for you? Do you want me to internalize that flag telling me what kind of collection is and only add `team_id` to the constructor? That way I'd leave `project_id` as is. I can then verify that this is called with one and only one of the two options and fail with an exception if it is not. Does that sound better?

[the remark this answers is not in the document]
kroenlein: Right when I said that the common pattern was
```
citrine = Citrine()
team = citrine.teams.get('xxx')
project = team.projects.get('xxx')
dataset = project.datasets.get('xxx')
```
I think I was saying that having the above functionally short circuit to
```
citrine = Citrine()
team = citrine.teams.get('xxx')
dataset = team.datasets.get('xxx')
```
by having `project.datasets.get('xxx')` be functionally equivalent to (pseudocode) `Team(project.team_id).datasets.get('xxx')`, then we support both the existing interface and the new pattern.  This could also be implemented either through explicitly creating a Team object, or by simply changing the route that the existing endpoint hits.  

We should also ask the question of if we want to deprecate the entire `Project.datasets` family of accessors, or if we think they are convenient enough that we want to maintain the redirect for the sake of ergonomics.  As a point in favor, if I were doing AI modeling (operating in a `project` context), I could definitely gain benefit in having a single command access to all potential dataset, as opposed to requiring holding onto a second `team` variable or having to chain methods.

> By that moment I was thinking on creating a new collection for the team datasets to not alter the constructor. What would be the best for you? Do you want me to internalize that flag telling me what kind of collection is and only add team_id to the constructor? That way I'd leave project_id as is. I can then verify that this is called with one and only one of the two options and fail with an exception if it is not. Does that sound better?

Why would we create a new collection for TeamDatasets instead of just adding a `team_id` to the existing Dataset collection?  The way those internal identifiers are handled is frankly a mess right now, but at least that keeps the paradigm consistent until we decide on a general refactor.  In Python, it's perfectly fine to use `None` as an indicator of object state.  Those are also really not user facing -- like I said, it need refactoring.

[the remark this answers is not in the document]
pacdaemon: I was choosing to create a new collection to prevent a lot of branching in the code as the project-based code will be killed (here I'm talking about calling the project-based URLs or any similar behavior). No matter if the datasets collection in the project is a proxy for team one, there is no reason for maintaining a project_id in the dataset collection, or is it? On whether we should have this proxy or not, I think that's more a call for @jspeerless or you, I think it could be beneficial for ergonomics if it doesn't make people confused about ownership. I'll not disagree if we think it is the best. Let me push a new version of this. But I still believe the project_id should be removed from the collection at a given point of time being that a breaking change if someone is not using the patterns you are referring to.

tests/seeding/test_find_or_create.py
```diff
@@ -267,8 +266,8 @@ def test_find_or_create_raise_error_project_exist_multiple(project_collection):
 
 def test_find_or_create_project_no_team(project_collection):
     # test when project collection has no team
-    collection = project_collection(team_id=None)
-    with pytest.raises(NotImplementedError):
```
pacdaemon, line 271: @kroenlein find_or_create relies in first enumerating the datasets on the dataset_collection, as intentionally the team_id wasn't provided, it is now going to fail in a different place.

kroenlein: I think no-team mode doesn't even work with the platform anymore.  If this is the case, this should be throwing errors.

pacdaemon: It will, some errors will be thrown in different places given the heavy dependency of teams in the dataset collection.

src/citrine/resources/project.py
lkubie: My preferred approach here would be to add a property for project.team which returns the parent Team of the Project, and then set an alias within the Project class of `project.datasets = project.team.datasets`. This would be both easy to maintain and read.

src/citrine/resources/project.py
lkubie, line 1: `owned_dataset_ids()` should be deprecated as projects will no longer own any datasets. We should be sure to add this equivalent to the Team class  as this method is how a user can tell if a dataset is shared with or native to a given Team. 

I could think there is a strong argument for depricating `gemd_batch_delete()` as it likely gives the user the impression that the GEMD objects will only be deleted from the project and not for the entire Team.

src/citrine/resources/team.py
lkubie, line 1: We will want to add an `owned_dataset_ids` property or method to Team to deprecate that in Projects today (It's a method today in Projects). This differs from Team.datasets.list() as that method returns the given Team's owned Datasets _and_ any Datasets shared with the Team from another Team.

lkubie:
I have gotten some feedback from the Customer org (the primary consumers and instructors of this SDK). There is a very strong desire to _not_ keep around methods/properties that are counter to the new inheritance structure. i.e., deprecate all endpoints like `project.datasets.list()` and `project.gemd.list()` and force the user to (eventually) explicitly use `team.datasets.list()` and `team.gemd.list()`. The team feels this will enforce understanding.

Some of this feedback ran counter to my initial intuition, so I may be going back and editing/deleting some old comments on this PR.

I also **strongly believe** that we should un-mock the tests for at least one round of testing to ensure we have not broken things once ready.

src/citrine/_session.py
```diff
@@ -303,6 +304,10 @@ def checked_get(self, path: str, **kwargs) -> Response:
         """Execute a GET request to a URL and utilize error filtering on the response."""
         return self.checked_request('GET', path, **kwargs)
 
+    def get_team_id_from_project_id(self, *, project_id: UUID) -> UUID:
```
lkubie, line 307: I am missing why we do not simply make a @property that is project.team that returns a Team object

src/citrine/resources/dataset.py
```diff
@@ -388,19 +392,62 @@ class DatasetCollection(Collection[Dataset]):
...
+                raise TypeError("Only one of project_id or team_id should be provided")
+            else:
+                warn("position parameter project_id is deprecated. Use keyword parameter"
+                     "project_id instead.",
+                     DeprecationWarning)
+                self._project_id = project_id
+
+        self._path_template = f"teams/{team_id}/datasets"
+
+    @property
```
lkubie, line 445: Why not return the team as opposed to the team UUID?

PR #928, opened 2024-04-25
[PNE-236] Add analysis workflow support.

anoto-moniz:
# Citrine Python PR

### PR Type:


test_requirements.txt
```diff
@@ -10,3 +10,9 @@ requests-mock==1.11.0
 sphinx==7.1.2
 sphinx-rtd-theme==2.0.0
 sphinxcontrib-apidoc==0.4.0
+
+# faker is a dependency of factory-boy, but factory-boy sets a very low floor
```
anoto-moniz, line 14: I was getting "Unknown formatter 'enum' with locale 'en_US'" errors. This was fixed in 15.1.0, but factory-boy only requires version >=0.7.0

src/citrine/_rest/engine_resource.py
```diff
@@ -53,13 +45,31 @@ def _post_dump(self, data: dict) -> dict:
...
+    status = properties.Optional(properties.String(), 'metadata.status.name', serializable=False)
+    """:Optional[str]: short description of the resource's status"""
+    status_detail = properties.List(properties.Object(StatusDetail), 'metadata.status.detail',
+                                    default=[], serializable=False)
+    """:List[StatusDetail]: a list of structured status info, containing the message and level"""
+
+    @classmethod
+    def build(cls, data: dict):
+        """Build an instance of this object from given data."""
+        return super().build_with_parent(data, __class__)
```
kroenlein, line 70: I don't know how we feel about making significant improvements to the properties module, but this seems like something that would make sense to make core.

tests/resources/test_analysis_workflow.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+def _assert_user_timestamp_equals_dict(user, time, ut_dict):
+    assert str(user) == ut_dict['user']
+    assert int(time.timestamp()) == int(ut_dict['time'] / 1000)
+
+
+def _assert_aw_plot_equals_dict(plot, plot_dict):
+    # Presently, plots are just raw JSON, so we can compare them directly.
+    assert plot == plot_dict
```
kroenlein, line 25: Not just having this be an in-line assert because we expect this to normally be checking multiple things?

kroenlein (commented):
Can you provide me with a link to the API docs for the routes you are wrapping?

src/citrine/_rest/engine_resource.py
anoto-moniz: Our intent to replace it is part of why I never bothered. If I end up stalling out on Pydantic, then maybe it becomes worthwhile.

tests/resources/test_analysis_workflow.py
anoto-moniz: I initially just set it up this way as part of a pattern before realizing there was nothing I needed to do for now. It could be useful if we ever add plot operations to the SDK. But for now, it doesn't feel important to collapse it.

src/citrine/informatics/workflows/analysis_workflow.py
```diff
@@ -0,0 +1,75 @@
...
+    """
+
+    uid = properties.UUID('id', serializable=False)
+    name = properties.String('data.name')
+    description = properties.String('data.description')
+    snapshot_id = properties.Optional(properties.UUID, 'data.snapshot_id')
+    plots = properties.List(properties.Raw, 'data.plots', default=[])
+
+    latest_build = properties.Optional(properties.Object(LatestBuild), 'metadata.latest_build',
+                                       serializable=False)
```
kroenlein, line 33: Is there a reason for not implementing `metadata.created`, `metadata.updated`, and `metadata.status`?

src/citrine/informatics/workflows/analysis_workflow.py
```diff
@@ -0,0 +1,75 @@
...
+
+    Note that plots are not fully supported. They're captured as raw JSON in order to facilitate
+    cloning an existing workflow, but no facilities are provided to validate them in the client.
+    """
+
+    uid = properties.UUID('id', serializable=False)
+    name = properties.String('data.name')
+    description = properties.String('data.description')
+    snapshot_id = properties.Optional(properties.UUID, 'data.snapshot_id')
+    plots = properties.List(properties.Raw, 'data.plots', default=[])
```
kroenlein, line 30: Since we are explicitly not supporting plot objects, wouldn't it make more sense to make this a private variable?  Otherwise, if we ever did add Plot objects, we'd have a breaking change going from a dict to a class.

kroenlein (commented):
I'd prefer to see plots as private, but if you have a good reason why they wouldn't be, I'll approve.

src/citrine/informatics/workflows/analysis_workflow.py
anoto-moniz: We are. That's part of what the inheritance from `EngineResourceWithoutStatus` does.

src/citrine/informatics/workflows/analysis_workflow.py
```diff
@@ -0,0 +1,75 @@
...
+        self.plots = plots
+
+    @property
+    def status(self) -> str:
+        """An alias for 'latest_build.status', to allow its usage with 'wait_while_validating."""
+        return self.latest_build.status
+
+    def _post_dump(self, data: dict) -> dict:
+        data["data"]["plots"] = [plot["data"] for plot in data["data"].get("plots") or []]
+        return super()._post_dump(data)
```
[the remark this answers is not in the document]
anoto-moniz: Same as above.

src/citrine/_rest/engine_resource.py
kroenlein: It's amazing how code collapse can make you blind...

kroenlein (approved):

PR #935, opened 2024-05-27
PNE-504: remove jobid table

andyczerwonka:
# Citrine Python PR

The backend allows you to submit a job id when submitting a table build. There is no reason to do that, and [we want to change the backend to take that feature away](https://citrine.atlassian.net/browse/PNE-503) to enable an optimization.

### PR Type:


src/citrine/resources/gemtables.py
```diff
@@ -210,21 +209,17 @@ def initiate_build(self, config: Union[TableConfig, str, UUID], *,
             if version is None:
                 raise ValueError('Version must be specified when building by config uid.')
             uid = config
```
kroenlein, line 211:
```suggestion
            uid = config
        logger.info(f'Submitting table build for config {uid} version {version}...')
```
Removing non-existent job_id from logging statement.  Including log statement so client knows which request went into the ether.

src/citrine/resources/gemtables.py
```diff
@@ -210,21 +209,17 @@ def initiate_build(self, config: Union[TableConfig, str, UUID], *,
...
             json={},
-            params={
-                'job_id': job_id
-            }
+            params={}
         )
         submission = JobSubmissionResponse.build(response)
-        logger.info('Build job submitted with job ID {}.'.format(submission.job_id))
+        logger.info('Table build job submitted from config {} version {} with job ID {}'
+                    .format(uid, version, submission.job_id))
```
kroenlein, line 222:
```suggestion
        logger.info(
            f'Table build job submitted from config {uid} '
            f'version {version} with job ID {submission.job_id}'
        )
```
Use f-string, per current best practice.

src/citrine/resources/gemtables.py
```diff
@@ -210,21 +209,17 @@ def initiate_build(self, config: Union[TableConfig, str, UUID], *,
...
         path = format_escaped_url('projects/{}/ara-definitions/{}/versions/{}/build',
                                   self.project_id, uid, version
                                   )
         response = self.session.post_resource(
             path=path,
             json={},
-            params={
-                'job_id': job_id
-            }
+            params={}
```
kroenlein, line 218:
```suggestion
```
Not a required argument

kroenlein (changes requested):
The goal of adding something to the logs before a call in case the call goes south was not satisfied by including the logging statement after return.

andyczerwonka:
> The goal of adding something to the logs before a call in case the call goes south...

I put it back. If that's a standard idiom in the SDK, we could adjust the session method to take a logging function for the before and after case, logging the input before the call is made and output after it returns. Not in this PR of course, but as a better way to enforce that standard.

kroenlein:
> > The goal of adding something to the logs before a call in case the call goes south...
> 
> I put it back. If that's a standard idiom in the SDK, we could adjust the session method to take a logging function for the before and after case, logging the input before the call is made and output after it returns. Not in this PR of course, but as a better way to enforce that standard.

It's not that the standard idiom is before + after.  It's always before, plus maybe after.  In this case, as long as we were logging a job-id, figured we should keep logging it.

PR #936, opened 2024-05-28
Revert "PNE-504: remove jobid table"

andyczerwonka:
Reverts CitrineInformatics/citrine-python#935

See https://citrineinformatics.slack.com/archives/C05MSDQGPDK/p1716845981281859


src/citrine/__version__.py
```diff
@@ -1 +1 @@
-__version__ = "3.2.6"
+__version__ = "3.2.5"
```
kroenlein, line 1:
```suggestion
__version__ = "3.2.7"
```

PR #947, opened 2024-06-20
[PNE-416] Use v4 listing endpoints.

anoto-moniz:
Predictors and design spaces have exposed v4 listing endpoints. The key difference is in the filtering. By default, they return all assets of the given type, and filters are specified as direct query arguments.

### PR Type:


src/citrine/_rest/pageable.py
```diff
@@ -69,14 +70,14 @@ def _fetch_page(self,
...
-        fetch_func = self.session.get_resource if fetch_func is None else fetch_func
+        fetch_func = fetch_func or partial(self.session.get_resource, version=self._api_version)
         json_body = {} if json_body is None else json_body
 
         module_type = getattr(self, '_module_type', None)
         params = self._page_params(page, per_page, module_type)
         params.update(additional_params or {})
 
-        data = fetch_func(path, params=params, version=self._api_version, **json_body)
+        data = fetch_func(path, params=params, **json_body)
```
StevenKauwe, line 80: So if a fetch_func is provided (not the one you define with partial args) it will always have the api version defined?

src/citrine/resources/design_space.py
```diff
@@ -127,6 +128,30 @@ def restore(self, uid: Union[UUID, str]) -> DesignSpace:
         entity = self.session.put_resource(url, {}, version=self._api_version)
         return self.build(entity)
 
+    def _list_base(self, *, per_page: int = 100, archived: Optional[bool] = None):
+        filters = {}
+        if archived is not None:
```
StevenKauwe, line 133: nit: This is use of `True`, `False`, and `None` is a bit unintuitive for me. In that it's unclear that the True and False are filters and that `None` is no filter. Not a huge deal, but maybe a quick enum for the three different archive_status' might be nice?

tests/resources/test_predictor.py
```diff
@@ -278,30 +278,70 @@ def test_train(valid_graph_predictor_data):
...
+    expected_call = FakeCall(method='GET',
+                             path='/projects/{}/predictors'.format(collection.project_id),
+                             params={'per_page': 25, 'page': 1})
+    assert 1 == session.num_calls, session.calls
+    assert expected_call == session.calls[0]
+    assert len(predictors) == 2
+
+
+def test_list_archived(valid_graph_predictor_data):
+    # Given
```
StevenKauwe, line 330: this structure is nice

src/citrine/resources/design_space.py
anoto-moniz: `None` always signifies the lack of specification: https://docs.python.org/3/c-api/none.html#c.Py_None

The right way to think of this is that omitting `archived` is no filter, which we indicate in Python with `None`.

tests/resources/test_predictor.py
anoto-moniz: Agreed. I try to follow something of the sort, even if I don't put the comments in there to clearly denote it.

src/citrine/_rest/pageable.py
anoto-moniz: Also, on a related note, this highlights that our units tests aren't checking the called API version matches what's expected. I vaguely recall knowing that, even purposely maintaining it to avoid needing to overhaul a bunch of tests. But I should look into fixing that...

StevenKauwe (dismissed):
Lgtm

src/citrine/_rest/pageable.py
anoto-moniz: FYI, I created the aforementioned ticket: https://citrine.atlassian.net/browse/PNE-662

StevenKauwe (approved):
lgtm

PR #949, opened 2024-07-10
[PNE-241] Data manager release

anoto-moniz:
# Citrine Python PR

## Description 

I wanted to preserve Lenore's old PR for posterity, but clean up the release we'll be working with. Hence this separate branch and PR.

### PR Type:


src/citrine/resources/gemtables.py
```diff
@@ -244,7 +246,12 @@ def get_by_build_job(self, job: Union[JobSubmissionResponse, UUID], *,
             The table built by the specified job.
 
         """
-        status = _poll_for_job_completion(self.session, self.project_id, job, timeout=timeout)
+        # TODO: Should this use the project or team version?
```
anoto-moniz, line 249: Calling this out here in case I forget to ask Pablo before the PR.

src/citrine/resources/data_concepts.py
```diff
@@ -482,7 +513,7 @@ def async_update(self, model: ResourceType, *,
         job_id = response_json["job_id"]
 
         if wait_for_response:
-            self.poll_async_update_job(job_id, timeout=timeout,
+            self.poll_async_update_job(job_id=job_id, timeout=timeout,
                                        polling_delay=polling_delay)
```
kroenlein, line 517: Are we intending to obsolete polling with just a job_id?  What's the point of this change?

src/citrine/resources/dataset.py
```diff
@@ -268,7 +290,7 @@ def delete(self, uid: Union[UUID, str, LinkByUID, DataConcepts], *, dry_run=Fals
             collection = self.gemd._collection_for(uid)
         else:
             collection = self.gemd
-        return collection.delete(uid, dry_run=dry_run)
+        return collection.delete(uid=uid, dry_run=dry_run)
```
kroenlein, line 293: Are we intending to obsolete deletion with just an object ID? What's the point of this change?

src/citrine/resources/dataset.py
```diff
@@ -386,21 +412,43 @@ class DatasetCollection(Collection[Dataset]):
...
+            obj_type="Datasets")
+
+    # After the Data Manager deprecation
+    # this can be a Class Variable using the `teams/...` endpoint
+    @property
+    def _path_template(self):
+        if self.project_id is None:
+            return f'teams/{self.team_id}/datasets'
+        else:
+            return f'projects/{self.project_id}/datasets'
```
kroenlein, line 451: If we're migrating  `_path_template` to a property, why aren't we doing that everywhere in its own PR?

src/citrine/resources/delete.py
```diff
@@ -14,9 +15,10 @@
 
 def _async_gemd_batch_delete(
         id_list: List[Union[LinkByUID, UUID, str, BaseEntity]],
-        project_id: UUID,
         session: Session,
-        dataset_id: Optional[UUID],
+        project_id: Optional[UUID] = None,
+        team_id: Optional[UUID] = None,
+        dataset_id: Optional[UUID] = None,
```
kroenlein, line 21: Why are these arguments getting re-ordered?  This is a breaking change, albeit to a private interface.  Why isn't there a `*, ` to block positional arguments?  `session` as positional argument doesn't really make sense.

src/citrine/resources/delete.py
```diff
@@ -71,31 +73,45 @@ def _async_gemd_batch_delete(
...
-    path = format_escaped_url('/projects/{project_id}/gemd/async-batch-delete',
-                              project_id=project_id
-                              )
+    if team_id is not None:
+        path = format_escaped_url('/teams/{team_id}/gemd/async-batch-delete',
+                                  team_id=team_id)
+    elif project_id is not None:
+        warn("project_id is deprecated as of 3.4.0, and will be removed in 4.0.0. Please provide "
+             "team_id instead",
+             DeprecationWarning)
```
kroenlein, line 82: Why are we issuing a Deprecation warning inside a private method?

src/citrine/resources/delete.py
```diff
@@ -71,31 +73,45 @@ def _async_gemd_batch_delete(
...
 
 def _poll_for_async_batch_delete_result(
-        project_id: UUID,
+        team_id: UUID,
         session: Session,
         job_id: str,
         timeout: float,
-        polling_delay: float
+        polling_delay: float,
+        project_id: UUID = None
```
kroenlein, line 106: Similar for `_async_gemd_batch_delete`

src/citrine/resources/file_link.py
```diff
@@ -192,14 +192,32 @@ def __str__(self):
...
+        project_id: Optional[UUID] = None
+    ):
+        args = _pad_positional_args(args, 3)
+        self.project_id = project_id or args[0]
+        self.dataset_id = dataset_id or args[1]
+        self.session = session or args[2]
+        if self.session is None:
+            raise TypeError("Missing one required argument: session.")
+        if self.dataset_id is None:
+            raise TypeError("Missing one required argument: dataset_id.")
```
kroenlein, line 214: Similar to DataConceptsCollection

src/citrine/resources/gemd_resource.py
```diff
@@ -1,38 +1,49 @@
...
+        super().__init__(*args,
+                         team_id=team_id,
+                         dataset_id=dataset_id,
+                         session=session,
+                         project_id=project_id)
+        args = _pad_positional_args(args, 3)
+        self.project_id = project_id or args[0]
+        self.dataset_id = dataset_id or args[1]
+        self.session = session or args[2]
+        self.team_id = team_id
```
kroenlein, line 46: Similar to DataConceptsCollection

src/citrine/resources/table_config.py
```diff
@@ -388,9 +416,11 @@ class TableConfigCollection(Collection[TableConfig]):
...
     _individual_key = None
 
-    def __init__(self, project_id: UUID, session: Session):
-        self.project_id = project_id
-        self.session: Session = session
+    def __init__(self, *args, team_id: UUID, project_id: UUID = None, session: Session = None):
+        args = _pad_positional_args(args, 2)
+        self.project_id = project_id or args[0]
+        self.session: Session = session or args[1]
+        self.team_id = team_id
```
kroenlein, line 423: Why would a Table Config need a Project ID?  Are Tables still affiliated with particular Projects and not affiliated with datasets anymore?

src/citrine/resources/project.py
```diff
@@ -249,6 +353,9 @@ def publish(self, *, resource: Resource):
         """
         resource_access = resource.access_control_dict()
         resource_type = resource_access["type"]
+        if resource_type == ResourceTypeEnum.DATASET:
+            warn("Datasets are no longer owned by Projects, so cannot be published by a Project.",
```
kroenlein, line 357: Shouldn't this warning explicitly call out the Team-based route?

src/citrine/resources/table_config.py
```diff
@@ -23,6 +24,11 @@
     IngredientIdentifierInOutput, IngredientQuantityInOutput, \
     IngredientLabelsSetByProcessAndName, IngredientLabelsSetInOutput
 
+from typing import TYPE_CHECKING
+if TYPE_CHECKING:   # pragma: no cover
+    from citrine.resources.project import Project
+    from citrine.resources.team import Team
+
```
kroenlein, line 31: Are we really benefiting from this?

https://docs.python.org/3.10/library/typing.html?highlight=type_checking#constant

kroenlein (commented):
That's a lot of changes....

src/citrine/resources/gemtables.py
pacdaemon: Not strictly required, but if it happens that you have the teamId it would be nice, I'm trying to move the jobs API to the team level.

src/citrine/resources/data_concepts.py
```diff
@@ -202,28 +202,56 @@ class DataConceptsCollection(Collection[ResourceType], ABC):
...
+                 *args,
+                 session: Session = None,
+                 dataset_id: Optional[UUID] = None,
+                 team_id: Optional[UUID] = None,
+                 project_id: Optional[UUID] = None):
+        # Handle positional arguments for backward compatibility
+        args = _pad_positional_args(args, 3)
+        self.project_id = project_id or args[0]
+        self.dataset_id = dataset_id or args[1]
+        self.session = session or args[2]
```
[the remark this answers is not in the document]
anoto-moniz: Best practice is to beware. We are using `typing` to indicate the caller should either be passing `None` or a type for which this holds true. If the caller violates the stated contract, they take responsibility for unexpected behavior.

src/citrine/resources/data_concepts.py
```diff
@@ -1,6 +1,6 @@
 """Top-level class for all data concepts objects and collections thereof."""
-from abc import abstractmethod, ABC
 import re
+from abc import abstractmethod, ABC
```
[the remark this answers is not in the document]
anoto-moniz: Because most resources also recommend grouping `import ...` and `from ... import ...` together (e.g. [Reddit's style guide](https://github.com/reddit-archive/reddit/wiki/PythonImportGuidelines)), and it's how do it in most places in CP already.

src/citrine/resources/data_concepts.py
anoto-moniz: What do you mean? It's just switching to use a keyword argument invocation, as we advocate almost everywhere.

src/citrine/resources/dataset.py
anoto-moniz: Same as above.

src/citrine/resources/delete.py
anoto-moniz: That looks like a mistake which should be reverted. Good catch.

src/citrine/resources/delete.py
anoto-moniz: Looking more closely, I don't think Lenore knew about `@deprecated`, and was actually trying to deprecate the place we call it from. I'll make that change.

src/citrine/resources/delete.py
anoto-moniz: This might actually highlight the reason more clearly: despite it being preceded with an underscore, we actually import it (and `_async_gemd_batch_delete`) outside of their modules. Which we should definitely correct.

That being said, you are correct, as a result they don't require deprecation warnings.

src/citrine/resources/table_config.py
anoto-moniz: It's debatable. This is what allowed me to add the proper type hints on lines 189-190 and 296-297 while still allowing it to compile and without `flake8` complaining. Exactly how valuable that is depends on how much people rely on the type hinting. 🤷

src/citrine/resources/data_concepts.py
```diff
@@ -202,28 +202,56 @@ class DataConceptsCollection(Collection[ResourceType], ABC):
...
-        self.dataset_id = dataset_id
-        self.session = session
+    def __init__(self,
+                 *args,
+                 session: Session = None,
+                 dataset_id: Optional[UUID] = None,
+                 team_id: Optional[UUID] = None,
+                 project_id: Optional[UUID] = None):
+        # Handle positional arguments for backward compatibility
+        args = _pad_positional_args(args, 3)
```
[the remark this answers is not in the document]
anoto-moniz: Given that we had to apply this to a bunch of classes, I was more worried about excessive verbosity as well as copying the same code in a bunch of places, making any changes to be applied broadly a huge pain. Given the deprecation warnings it triggers, I'm not too worried about people providing both.

src/citrine/_utils/functions.py
```diff
@@ -319,3 +319,25 @@ def resource_path(*,
     new_url = base._replace(path='/'.join(path), query=query).geturl()
 
     return format_escaped_url(new_url, *action, **kwargs, uid=uid)
+
+
+def _data_manager_deprecation_checks(session, project_id: UUID, team_id: UUID, obj_type: str):
+    if project_id is None and team_id is None:
+        raise TypeError("Missing one required argument: team_id.")
```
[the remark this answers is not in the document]
anoto-moniz: Definite gray zone. But I like using `TypeError` here because it will mirror the error they'd get in the future once the deprecated code is removed.

src/citrine/jobs/job.py
```diff
@@ -96,7 +97,12 @@ def _poll_for_job_completion(session: Session,
         job_id = job.job_id
     else:
         job_id = job  # pragma: no cover
-    path = format_escaped_url('projects/{}/execution/job-status', project_id)
+    if project_id is None and team_id is None:
+        raise TypeError('Either a project_id or team_id must be provided to poll a job.')
+    if team_id is not None:
+        path = format_escaped_url('teams/{}/execution/job-status', team_id)
+    else:
+        path = format_escaped_url('projects/{}/execution/job-status', project_id)
```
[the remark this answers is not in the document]
anoto-moniz: I need to check with Pablo. Based on his comment above, it sounds like the intent is for the project-level jobs endpoint to be deprecated.

src/citrine/resources/data_concepts.py
```diff
@@ -202,28 +202,56 @@ class DataConceptsCollection(Collection[ResourceType], ABC):
 
     Parameters
     ----------
-    project_id: UUID
-        The uid of the project that this collection belongs to.
```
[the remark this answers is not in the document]
anoto-moniz: Good callout, I should update that message.

My thought had been that users are mostly getting to this through `ProjectCollection.<concept>`. But A) that's a bad approach to the message content, and B) adding `@deprecated` to those methods means this one will be hidden in the default `warnings` configuration.

src/citrine/resources/data_concepts.py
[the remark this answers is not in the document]
kroenlein: Python has no contracts.

src/citrine/resources/dataset.py
anoto-moniz: I agree that in an ideal world, this mega-PR would have been chunked up into smaller, more focused PRs. Moving everything (at least data-related) to keyword arguments is another example that comes to mind. But given it changed hands multiple times, most recently due to the layoffs, and has come under time pressure, it didn't seem like splitting it up after the fact would be worthwhile when it got to me.

Also, just to be clear, the intention isn't to migrate `_path_template` to a property everywhere. It's just a temporary approach for the endpoints relevant to data manager until the `/projects` versions are removed.

src/citrine/resources/table_config.py
anoto-moniz: My understanding is that it's still needed to retrieve it based on the table it built: https://github.com/CitrineInformatics/citrine-python/blob/c7117a6640bd5be6bae97b09e29eb527bd02e734/src/citrine/resources/table_config.py#L460

I easily could be wrong though. @pacdaemon did I misinterpret what you said?

src/citrine/resources/project.py
anoto-moniz: "Publishing" no longer means anything with regards to datasets in the world of Data Manager. Previously, it's what made them visible to the team so another project could pull them in. Now, since everything is at a team level, it's a no-op.

Sounds like the message should probably be changed to be a bit clearer.

src/citrine/resources/dataset.py
kroenlein: To be clear, `_path_template` would be better written as an abstract method so that we could require every Collection to implement it, but that's definitely outside scope here.

Just giving appropriate stylistic push back.  I chose `comment` because I saw nothing blocking.

src/citrine/resources/data_concepts.py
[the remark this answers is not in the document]
kroenlein: Upon further reflection, I think team_id is not Optional.  Scenarios where no team_id is passed result in either errors or deprecation warnings.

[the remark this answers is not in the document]
anoto-moniz: > I think team_id is not Optional

Agreed! I intended to drop the `Optional` typing, but seems I missed some spots (or maybe completely forgot). Good catch!

src/citrine/resources/data_concepts.py
```diff
@@ -202,28 +202,56 @@ class DataConceptsCollection(Collection[ResourceType], ABC):
...
+        return f'teams/{self.team_id}/datasets/{self.dataset_id}/{collection_key}'
+
+    # After Data Manager deprecation, both can use the `teams/...` path.
+    @property
+    def _dataset_agnostic_path_template(self):
+        if self.project_id is None:
+            return f'teams/{self.team_id}/{self._collection_key.replace("_","-")}'
+        else:
+            return f'projects/{self.project_id}/{self._collection_key.replace("_","-")}'
+
```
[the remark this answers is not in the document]
anoto-moniz: Yeah, agreed on both counts. We decided to accept it as a necessary evil to make the transition easier.

src/citrine/jobs/job.py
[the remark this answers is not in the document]
pacdaemon: If the team_id is known we should call the team API. I'm unsure if we have guarantees of the team_id to be available here. But yes, ideally we should be moving to the new pattern. As a side note, the project-based job ID is not going to be deprecated because of the data manager, it will be for another reason (having a homogenous API for dealing with jobs).

src/citrine/resources/project.py
pacdaemon: Yes, publishing doesn't mean anything now for team-based datasets, however, the accounts service doesn't implement an NOP for this type of dataset. Instead, it returns a 403, because the project doesn't have admin rights on the new type of dataset. I'd say that on top of the warning we should not call the API at all.

src/citrine/_utils/functions.py
```diff
@@ -319,3 +319,25 @@ def resource_path(*,
     new_url = base._replace(path='/'.join(path), query=query).geturl()
 
     return format_escaped_url(new_url, *action, **kwargs, uid=uid)
+
+
+def _data_manager_deprecation_checks(session, project_id: UUID, team_id: UUID, obj_type: str):
```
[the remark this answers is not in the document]
anoto-moniz: Inherited code. =) Yeah, I agree that your structure is better.

src/citrine/resources/material_run.py
```diff
@@ -139,8 +139,14 @@ def get_history(self, id: Union[str, UUID, LinkByUID, MaterialRun]) -> MaterialR
...
         }
         data = self.session.post_resource(path, json=query)
-
-        return MaterialRun.build(data)
+        if data and data[0]["roots"]:
+            # Since the above query presents a single dataset to the endpoint, the response will be
+            # a list of length one, with a single route.
+            history_data = data[0]
+            history_data["roots"] = history_data["roots"][0]
+            return MaterialRun.build(history_data)
```
anoto-moniz, line 147: This was a change that came from Pablo running the e2es.

The result of the endpoint changed slightly, such that it can now encode multiple material histories. He confirmed to me that because this call is querying on a single dataset ID, it will always return a single element, and its `roots` field will be an array of length 1.

I could have put this stuff in `_pre_build`, but my impression is we very well might want to expose the whole thing at some point in the future, so I'd make it clear these changes apply to this specific usage.

kroenlein (approved):
Cosmetic fixes.  Nothing blocking.

* I thought I submitted this review, sorry.

src/citrine/resources/dataset.py
anoto-moniz: Same as above: the `delete` signature hasn't changed, just the invocation here.

src/citrine/resources/material_run.py
```diff
@@ -123,14 +121,32 @@ def get_history(self, id: Union[str, UUID, LinkByUID, MaterialRun]) -> MaterialR
...
+                        {
+                            "scope": link.scope,
+                            "id": link.id
+                        }
+                    ]
+                }
+            ]
+        }
+        data = self.session.post_resource(path, json=query)
+        if data and data[0]["roots"]:
```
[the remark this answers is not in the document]
anoto-moniz: 😬 You are correct!

src/citrine/resources/material_run.py
anoto-moniz: Agreed, it's not great. And yeah, I have no problem making that tweak. Looking at the code in `GEMDResource`, it should work fine.

kroenlein (approved):

PR #950, opened 2024-07-18
[PNE-203] Fix reading file links on Windows.

anoto-moniz:
If FileCollection.read is called with a FileLink which encodes a local path (i.e. starts with file:///), then it's a local file. Due to the differences in path handling on Windows and Unix, we need to explicitly convert the path portion of that URL into a filepath.

### PR Type:


anoto-moniz:
I tested this by spinning up a Windows VM in VirtualBox and running the snippet [here](https://github.com/CitrineInformatics/citrine-python/blob/07e76172c05c3036404aec158df54f24ea71cbf8/src/citrine/resources/file_link.py#L651) on a variety of Windows-style paths (and a few Linux), and ensuring the output looked right.

kroenlein (commented):
How thoroughly was this tested?  It sounds reasonable, but I'd like to maybe see additional tests added, since there were no additions to the test suite in this PR.  

Also, I note that the documentation to [url2pathname](https://docs.python.org/3.10/library/urllib.request.html?highlight=url2pathname#urllib.request.url2pathname) says

> This function uses [unquote()](https://docs.python.org/3.10/library/urllib.parse.html#urllib.parse.unquote) to decode path.

which sounds like we might have a double-decode bug in this workflow.

anoto-moniz:
> How thoroughly was this tested? It sounds reasonable, but I'd like to maybe see additional tests added, since there were no additions to the test suite in this PR.
> 
> Also, I note that the documentation to [url2pathname](https://docs.python.org/3.10/library/urllib.request.html?highlight=url2pathname#urllib.request.url2pathname) says
> 
> > This function uses [unquote()](https://docs.python.org/3.10/library/urllib.parse.html#urllib.parse.unquote) to decode path.
> 
> which sounds like we might have a double-decode bug in this workflow.

While I'm sure there's a way to test Windows filepaths in a Linux environment, no obvious way revealed itself on Googling, which means it will be a lot more effort than it's worth for this bug, given it was reported months ago but seemingly not caused many of our customers any problems.

Also, we use `urlunquote_plus`, hence why I left it as is. But yeah, you're right about the double decode, so that would need to be fixed.

I was just looking for a quick win in some downtime, but given this has a bit more to it, I'm gonna put it aside and come back to it later.

src/citrine/resources/file_link.py
```diff
@@ -648,7 +648,11 @@ def read(self, *, file_link: Union[str, UUID, FileLink]) -> bytes:
         file_link = self._resolve_file_link(file_link)
 
         if self._is_local_url(file_link.url):  # Read the local file
-            path = Path(unquote_plus(url2pathname(urlparse(file_link.url).path)))
+            parsed_url = urlparse(file_link.url)
+            if parsed_url.netloc not in {'', '.', 'localhost'}:
+                raise ValueError("Non-local UNCs (e.g., Windows network paths) are not supported.")
```
anoto-moniz, line 653: This is a new restriction, is it not?

kroenlein: Formally it's a new restriction, but we would have broken on this case before because we were not considering the netloc value.

kroenlein (approved):
With changes, I this is ready.

PR #955, opened 2024-07-19
Building GEMD tables should always be team scope.

anoto-moniz:
There is a slight behavior change here when omitting the dataset context, in the scope of datasets considered. However, customers aren't using it right now, so accounting for it is unnecessary.

Also take this chance to update ara-definitions endpoints to the modern table-configs segment.

### PR Type:


src/citrine/resources/table_config.py
```diff
@@ -421,6 +421,10 @@ def __init__(self, *args, team_id: UUID, project_id: UUID = None, session: Sessi
         self.project_id = project_id or args[0]
         self.session: Session = session or args[1]
         self.team_id = team_id
+        if self.project_id is None:
+            raise TypeError("Missing one required argument: project_id.")
+        if self.session is None:
+            raise TypeError("Missing one required argument: session.")
```
anoto-moniz, line 427: I noticed we weren't enforcing that these were provided, despite previously being required, and still assumed to be non-`None` in the class.

anoto-moniz:
Turns out the `table-configs` endpoint doesn't return some data which the `TableConfig.build` call relies on. As such, switching over to it will require more effort. And given we want to get data manager out ASAP, we'll ticket it up and take care of it later.

PR #958, opened 2024-07-25
Restrict enumerated design space data value types.

anoto-moniz:
The backend has always restricted them to strings, but the SDK never enforced it. And since the backend used a lax serde library, it would automatically coerce ints and floats to strings. To preserve backwards compatability while moving in the right direction, `data` is restricted to int, float, and str, and when assigning a data list, if any int or float values trigger a deprecation warning.

### PR Type:


kroenlein (approved):
LGTM

PR #964, opened 2024-09-18
Adjust type of ChemicalFormulaFeaturizer.powers.

anoto-moniz:
The backend is opening up the type of powers to be floats. As such, this commit will deserialize powers as a list of floats. However, to preserve backwards compatability, the type of powers will be preserved, and a new field (powers_float) with the list as floats is introduced. In v4.0.0, the type of 'powers' will be changed, and 'powers_float' will be deprecated before its removal in v5.0.0.

### PR Type:


tests/informatics/test_predictors.py
```diff
@@ -259,7 +259,10 @@ def test_chemical_featurizer(chemical_featurizer):
     assert chemical_featurizer.input_descriptor == ChemicalFormulaDescriptor("formula")
     assert chemical_featurizer.features == ["standard"]
     assert chemical_featurizer.excludes == []
-    assert chemical_featurizer.powers == [1, 2]
+    with pytest.warns(UserWarning):
+        assert chemical_featurizer.powers == [1, 2]
+    with pytest.warns(PendingDeprecationWarning):
+        assert chemical_featurizer.powers_float == [1.0, 2.0]
```
kroenlein, line 265:
```suggestion
        assert chemical_featurizer.powers_float == [1.0, 2.0]
    chemical_featurizer.powers == [0.5, -1.5]
    with pytest.warns(PendingDeprecationWarning):
        assert chemical_featurizer.powers_float == [0.5, -1.5]
    with pytest.raises(ValueError)"
        _ = chemical_featurizer.powers
```

src/citrine/informatics/predictors/chemical_formula_featurizer.py
```diff
@@ -152,5 +153,23 @@ def __init__(self,
...
+    @property
+    def powers(self) -> List[int]:
+        """The list of powers when computing generalized weighted means of element properties."""
+        warn("The type of 'powers' will change to a list of floats in v4.0.0. To retrieve them as "
+             "floats now, use 'powers_float'.")
+        return [int(p) for p in self._powers]
+
+    @powers.setter
+    def powers(self, value: List[Union[int, float]]):
+        self._powers = value
```
kroenlein, line 165:
```suggestion
        self._powers = [float(p) for p in value]
```
If we're being particular about int v. float, then the above setter would cause powers_float to potentially return a list of ints, breaking the API.

kroenlein (changes requested):
I think the migration strategy proposed is overkill given the actual impacts in Python of changing an int to a float.  But I will disagree and commit.

Returning incorrect values when the object contains non-integer powers feels like mishandling data.  That definitely violates minimum-surprise for me.

src/citrine/informatics/predictors/chemical_formula_featurizer.py
```diff
@@ -152,5 +153,23 @@ def __init__(self,
...
+
+    @property
+    def powers_float(self) -> List[float]:
```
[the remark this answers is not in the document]
anoto-moniz: Sure.

src/citrine/informatics/predictors/chemical_formula_featurizer.py
```diff
@@ -152,5 +153,23 @@ def __init__(self,
...
+        """Powers when computing generalized weighted means of element properties."""
+        warn("'powers_float' will be deprecated in v4.0.0 for 'powers', and removed in v5.0.0",
+             PendingDeprecationWarning)
```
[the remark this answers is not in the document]
anoto-moniz: That's why `power_float` exists. The field is changing. The alternative is to mark this field deprecated and reintroduce it in v4.0.0, which feels unnecessary.

src/citrine/informatics/predictors/chemical_formula_featurizer.py
anoto-moniz: Nope. The serde library converts them automatically.

tests/informatics/test_predictors.py
anoto-moniz: Sure.

src/citrine/informatics/predictors/chemical_formula_featurizer.py
```diff
@@ -152,5 +153,23 @@ def __init__(self,
         self.excludes = excludes if excludes is not None else []
         self.powers = powers if powers is not None else [1]
 
+    @property
+    def powers(self) -> List[int]:
+        """The list of powers when computing generalized weighted means of element properties."""
+        warn("The type of 'powers' will change to a list of floats in v4.0.0. To retrieve them as "
+             "floats now, use 'powers_float'.")
+        return [int(p) for p in self._powers]
```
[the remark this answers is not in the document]
anoto-moniz: All of this is about preserving the current behavior, as an interface must. Rounding is a distinct behavior change, and not what the server does, so it's doubly wrong.

tests/informatics/test_predictors.py
```diff
@@ -259,7 +259,10 @@ def test_chemical_featurizer(chemical_featurizer):
     assert chemical_featurizer.input_descriptor == ChemicalFormulaDescriptor("formula")
     assert chemical_featurizer.features == ["standard"]
     assert chemical_featurizer.excludes == []
-    assert chemical_featurizer.powers == [1, 2]
+    with pytest.warns(UserWarning):
```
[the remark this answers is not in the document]
anoto-moniz: A `UserWarning` is not a deprecated call.

src/citrine/informatics/predictors/chemical_formula_featurizer.py
kroenlein: So then you don't actually accept int to the setter, and your TypeHint is incorrect.

src/citrine/informatics/predictors/chemical_formula_featurizer.py
```diff
@@ -152,5 +153,23 @@ def __init__(self,
         self.excludes = excludes if excludes is not None else []
         self.powers = powers if powers is not None else [1]
 
+    @property
+    def powers(self) -> List[int]:
+        """The list of powers when computing generalized weighted means of element properties."""
+        warn("The type of 'powers' will change to a list of floats in v4.0.0. To retrieve them as "
+             "floats now, use 'powers_as_float'.")
+        return [int(p) for p in self._powers]
```
kroenlein, line 161:
```suggestion
        converted = [round(p) for p in self._powers]
        if converted == self._powers:
            warn("The type of 'powers' will change to a list of floats in v4.0.0. To retrieve them as "
                 "floats now, use 'powers_as_float'.", DeprecationWarning)
        else:
            diffs = [f"{x} => {y}" for x, y in zip(self._powers, converted) if x != y]]
            warn(f"The following powers were cast to ints: {'; '.join(diffs)}. To retrieve them as "
                 "floats, use 'powers_as_float'.")
        return converted
```

kroenlein (commented):
A proposed workflow that alerts users when we cast their values without adding additional exceptions.

src/citrine/informatics/predictors/chemical_formula_featurizer.py
kroenlein: Concern withdrawn; the `properties` module handles casting the result.

StevenKauwe (dismissed):
Catching up from OOO. These changes look good to me! 👍

They seem to preserve current client behavior while also adding the ability to work with floats if the user opts into it. Beyond that, looks like a lot already got discussed. This was helpful for catching up!

src/citrine/informatics/predictors/chemical_formula_featurizer.py
anoto-moniz: I don't love adding an extra warning, but if it will move this PR forward, then I'm willing to do so.

Adding `round` does not preserve the old behavior, so it does not belong here.

kroenlein (approved):
Given the small number of people who I think will be affected by this and the addition of a mutation warning, I'll approve.

PR #968, opened 2024-10-03
Add a datasource field to the DesignWorkflow object

kroenlein:
# Citrine Python PR

## Description 
* Add SnapshotDataSource to the implemented classes of DataSource
* Augment DesignWorkflow to include a datasource field
* Consolidate DesignWorkflow test objects into one factory

When the user of the UI creates a branch with a data source but not yet any predictor, the API DesignWorkflow object has a direct reference to the data source.  When a Predictor was referenced by the DesignWorkflow, that Predictor contained a reference to the DataSource, thus an SDK user could ultimately recover the value, but without a Predictor, there is no such workflow.

Unfortunately, the API returns a DataSourceId, which is a specially formatted string, not a DataSource object, as with other API references to DataSources. To avoid creating a new class or requiring users to manually construct a specially formatted string, this PR also adds basic mapping tools between DataSourceId strings and DataSource objects, as well as a convenience method for generating a TableDataSource from a GemTable object.

I think this approach is ready, but happy to work through concerns or potential improvements.

Prerequisite PR for https://citrine.atlassian.net/browse/PNE-731

### PR Type:


src/citrine/informatics/data_sources.py
```diff
@@ -74,6 +104,20 @@ def __init__(self,
         self.column_definitions = column_definitions
         self.identifiers = identifiers
 
+    @classmethod
+    def _data_source_id_builder(cls, *args) -> DataSource:
+        # TODO Figure out how to populate the column definitions
+        return CSVDataSource(
+            file_link=FileLink(url=args[0], filename=args[1]),
+            column_definitions={}
+        )
```
kroenlein, line 113: There is no obvious method to identify what the column definitions and identifiers would be from just the DataSourceId.

src/citrine/informatics/data_sources.py
```diff
@@ -113,5 +181,42 @@ class ExperimentDataSourceRef(Serializable['ExperimentDataSourceRef'], DataSourc
...
+    @classmethod
+    def _data_source_id_builder(cls, *args) -> DataSource:
+        return ExperimentDataSourceRef(datasource_id=UUID(args[0]))
+
+    def to_data_source_id(self) -> str:
+        """Generate the data_source_id for this DataSource."""
+        return "::".join(str(x) for x in [self._data_source_type, self.datasource_id])
+
+
+class SnapshotDataSource(Serializable['SnapshotDataSource'], DataSource):
```
kroenlein, line 198: This will presumably be useful soon, since it's the next step after Multistep Materials tables.

src/citrine/informatics/workflows/design_workflow.py
```diff
@@ -50,20 +52,54 @@ def __init__(self,
...
+
+    def _post_dump(self, data: dict) -> dict:
+        """Run data modification after dumping."""
+        data_source = data.pop("data_source", None)
+        if data_source is not None:
+            data["data_source_id"] = DataSource.build(data_source).to_data_source_id()
+        else:
+            data["data_source_id"] = None
+        return data
+
```
kroenlein, line 83: This uses the existing serde hooks to filter out `data_source_id` fields and insert DataSource objects.

tests/conftest.py
```diff
@@ -949,16 +949,3 @@ def predictor_evaluation_workflow_dict(generic_entity, example_cv_evaluator_dict
...
-    ret = generic_entity.copy()
-    ret.update({
-        "name": "Example Design Workflow",
-        "description": "A description! Of the Design Workflow! So you know what it's for!",
-        "design_space_id": str(uuid.uuid4()),
-        "predictor_id": str(uuid.uuid4()),
-        "predictor_version": random.randint(1, 10),
-        "branch_id": str(uuid.uuid4()),
-    })
-    return ret
```
kroenlein, line 964: Having two different DesignWorkflow test objects (a fixture & a Factory) made testing on the updated object complicated.

tests/resources/test_design_workflows.py
```diff
@@ -42,24 +48,9 @@ def collection(branch_data, collection_without_branch) -> DesignWorkflowCollecti
...
-    workflow.uid = uuid.uuid4()
-    return workflow
-
-
-@pytest.fixture
-def workflow_minimal(collection, workflow) -> DesignWorkflow:
-    workflow.predictor_id = None
-    workflow.predictor_version = None
-    workflow.design_space_id = None
-    return workflow
```
kroenlein, line 62: This became obsolete because it was only used twice, and so was better served by a Factory with arguments.

src/citrine/informatics/data_sources.py
```diff
@@ -34,13 +36,39 @@ def get_type(cls, data) -> Type[Serializable]:
...
+    def _data_source_type(self) -> str:
+        """The data source type string, which is the leading term of the data_source_id."""
+
+    @classmethod
+    def from_data_source_id(cls, data_source_id: str) -> "DataSource":
+        """Build a DataSource from a datasource_id."""
+        terms = data_source_id.split("::")
+        types: List[Type[Serializable]] = [
+            CSVDataSource, GemTableDataSource, ExperimentDataSourceRef, SnapshotDataSource
+        ]
```
anoto-moniz, line 57: Nitpick: maybe lift this list to the class level, as it's duplicated above.

src/citrine/informatics/data_sources.py
```diff
@@ -34,13 +36,39 @@ def get_type(cls, data) -> Type[Serializable]:
...
+        res = next((x for x in types if x._data_source_type == terms[0]), None)
+        if res is None:
+            raise ValueError("Unrecognized type: {}".format(terms[0]))
```
anoto-moniz, line 60: Nitpick: f-string.

src/citrine/informatics/data_sources.py
```diff
@@ -92,13 +136,37 @@ class GemTableDataSource(Serializable['GemTableDataSource'], DataSource):
...
 
+    @classmethod
+    def _data_source_id_builder(cls, *args) -> DataSource:
+        return GemTableDataSource(table_id=UUID(args[0]), table_version=args[1])
+
+    def to_data_source_id(self) -> str:
+        """Generate the data_source_id for this DataSource."""
+        return "::".join(
+            str(x) for x in [self._data_source_type, self.table_id, self.table_version]
+        )
```
anoto-moniz, line 156: Nitpick: what's the benefit of this over an f-string?

f`{self._data_source_type}::{self.table_id}::{self.table_version}`

src/citrine/informatics/data_sources.py
```diff
@@ -113,5 +181,42 @@ class ExperimentDataSourceRef(Serializable['ExperimentDataSourceRef'], DataSourc
...
     def __init__(self, *, datasource_id: UUID):
         self.datasource_id: UUID = datasource_id
+
+    @classmethod
+    def _data_source_id_builder(cls, *args) -> DataSource:
+        return ExperimentDataSourceRef(datasource_id=UUID(args[0]))
+
+    def to_data_source_id(self) -> str:
+        """Generate the data_source_id for this DataSource."""
+        return "::".join(str(x) for x in [self._data_source_type, self.datasource_id])
```
anoto-moniz, line 195: Same nitpick as above.

src/citrine/informatics/data_sources.py
anoto-moniz: Yeah, Bill and I were just talking about what's needed to properly support training, etc on snapshots.

src/citrine/informatics/data_sources.py
```diff
@@ -113,5 +181,42 @@ class ExperimentDataSourceRef(Serializable['ExperimentDataSourceRef'], DataSourc
...
+    def __init__(self, *, snapshot_id: UUID):
+        self.snapshot_id = snapshot_id
+
+    @classmethod
+    def _data_source_id_builder(cls, *args) -> DataSource:
+        return SnapshotDataSource(snapshot_id=UUID(args[0]))
+
+    def to_data_source_id(self) -> str:
+        """Generate the data_source_id for this DataSource."""
+        return "::".join(str(x) for x in [self._data_source_type, self.snapshot_id])
```
anoto-moniz, line 222: Same as above.

src/citrine/informatics/data_sources.py
anoto-moniz: Would it be too disruptive to issue a warning here, so the user knows that the data source they're getting is incomplete? Actually, maybe the warning should come when they access the `DesignWorkflow.data_source` field, if it's a `CSVDataSource` with no `column_definition`.

Either way, I'd like if we can alert them this isn't a completely accurate representation.

anoto-moniz (dismissed):
Mostly looks good, just some nitpick comments which aren't worth blocking on.

My biggest concern is the same you call out in TODOs regarding the `CSVDataSource`. I'd like the sort of warning I mention, but I'm approving anyways since I don't know it's the most crucial thing. And we can always add it later if it becomes a problem.

src/citrine/informatics/data_sources.py
kroenlein: My thought when writing was that we have a series of identifiers joined by `::` in all cases, but yes, it ended up not great.  I thought of creating a `get_identifiers` method that got joined at the parent level, but then that seemed a bit absurd.  I'll swap to f-string.

src/citrine/informatics/data_sources.py
kroenlein: Yeah, using a warning makes sense.  I don't think people are using CSVDataSources anymore, but...

anoto-moniz (approved):

PR #970, opened 2024-10-09
Modify compound ingest operation to wait for table build completion

kroenlein:
# Citrine Python PR

## Description 
This PR updates the poll_for_job_completion method in ingestion to continue blocking until a table is built if the build_table option was true.

This PR also 
* updates the status attribute of Job objects to be an enumeration (as per the API spec), maintaining support for arbitrary strings with issued deprecation warnings.
* Adds a deprecation warning for any use of project_id on an Ingestion object, whereas before only changing the value issued this warning

### PR Type:


src/citrine/jobs/job.py
```diff
@@ -50,13 +77,37 @@ class JobStatusResponse(Resource['JobStatusResponse']):
...
-    status = properties.String("status")
+    _status = properties.String("status")
     """:str: The status of the job. One of "Running", "Success", or "Failure"."""
     tasks = properties.List(Object(TaskNode), "tasks")
     """:List[TaskNode]: all of the constituent task required to complete this job"""
     output = properties.Optional(properties.Mapping(String, String), 'output')
     """:Optional[dict[str, str]]: job output properties and results"""
 
+    @property
+    def status(self) -> Union[JobStatus, str]:
```
anoto-moniz, line 88: Since `BaseEnumeration` has `str` as a superclass, returning a `JobStatus` would not lead to a change in behavior, so this is not an API break.

src/citrine/jobs/job.py
```diff
@@ -50,13 +77,37 @@ class JobStatusResponse(Resource['JobStatusResponse']):
...
+        """The last reported status of this particular task."""
+        if resolved := JobStatus.from_str(self._status, exception=False):
+            return resolved
+        else:
+            return self._status
+
+    @status.setter
+    def status(self, value):
```
anoto-moniz, line 96: Nitpick: type hint

src/citrine/jobs/job.py
```diff
@@ -50,13 +77,37 @@ class JobStatusResponse(Resource['JobStatusResponse']):
...
+        if resolved := JobStatus.from_str(value, exception=False):
+            if resolved not in [JobStatus.RUNNING, JobStatus.SUCCESS, JobStatus.FAILURE]:
```
anoto-moniz, line 98: Shouldn't this support any valid `JobStatus` value? If so, you can use `list(JobStatus)` to ensure it's resilient against typos and future code changes.

anoto-moniz: Oh, I see. That's strange, that the `Response` can't take the other statuses. Not even `Pending`?

anoto-moniz (dismissed):

PR #974, opened 2024-11-12
PNE-816: Make the project search endpoint to use the teams version one

pacdaemon:
# Citrine Python PR

## Description 

Jira ticket: https://citrine.atlassian.net/browse/PNE-816

For searching for projects in the SDK we were using a team-agnostic endpoint, causing the search to be executed in a broader scope (all of the teams visible to the user). This work switches the endpoint so the search will have the right scope.

### PR Type:


src/citrine/resources/project.py
```diff
@@ -647,7 +647,11 @@ def search_all(self, search_params: Optional[Dict]) -> Iterable[Dict]:
 
         """
         collections = []
-        path = self._get_path(action="search")
+        if self.team_id is None:
+            path = "/projects/search"
+        else:
+            path = format_escaped_url("/teams/{team_id}/projects/search", team_id=self.team_id)
+
```
kroenlein, line 654: I'd prefer to follow the existing patterns if possible, as opposed the the more local definition here.  Are there any routes that still should always be using a team-agnostic project access?  It seems like the better strategy here is to define `_path_template` as a `@property` that conditionally includes the team.  

I can code that up if it doesn't make sense.

kroenlein (commented):
I have a preferred solution approach, but I'm not fundamentally opposed to this solution.

src/citrine/resources/project.py
pacdaemon: It totally makes sense. For this, we might need to do some research to verify that every endpoint called has a counterpart. If you agree, I would ticket that up, as I imagine I will find a few interesting things while doing it.
I can handle that myself after some high-priority work I need to address on another topic. What do you think?

kroenlein (approved):
With the merged changes, this looks good to go.  Verified functionality with current e2es, but those should be updated to use team routes.

pacdaemon:
I reproduced the flow locally, and I have a green light on the execution:

### SDK code and response
Screenshot 2024-11-15 at 15 33 08

### Account service logs
```
2024-11-15 15:31:27 POST /api/v3/projects/search?userId=6f664ee7-aae8-45f9-8afc-24747379f4c1 200 17.739 ms - 15
```

PR #976, opened 2024-11-15
Revert "PNE-816: Make the project search endpoint to use the teams version one"

pacdaemon:
Reverts CitrineInformatics/citrine-python#974


src/citrine/__version__.py
```diff
@@ -1 +1 @@
-__version__ = "3.11.1"
+__version__ = "3.11.2"
```
robbtraister, line 1: you have to go to `.3`, since `.2` is the current

robbtraister (dismissed):

src/citrine/__version__.py
```diff
@@ -1 +1 @@
-__version__ = "3.11.2"
+__version__ = "3.11.3"
```
robbtraister, line 1: this now needs to be `.4` because of #977

robbtraister (dismissed):

andyczerwonka (approved):

jamie-heller (approved):

PR #982, opened 2024-11-22
[PNE-6367] Add support for feature effects.

anoto-moniz:
The payload comes in as a condensed format, so it's expanded in order to contructed nested lists of objects for clarity and ease of use.

Additionally, the hierarchy of data is flipped to more closely match how it will be used by our customers. To that end, 'as_dict' is provided to ease importing it into a pandas DataFrame for whatever processing and analysis the customer desires.

### PR Type:


src/citrine/resources/table_config.py
```diff
@@ -88,6 +88,7 @@ class TableConfig(Resource["TableConfig"]):
         The query used to define the materials underpinning this table
     generation_algorithm: TableFromGemdQueryAlgorithm
         Which algorithm was used to generate the config based on the GemdQuery results
+
```
anoto-moniz, line 91: My local `flake8` was complaining about this file, despite me not touching it. 🤷‍♂️

src/citrine/informatics/feature_effects.py
```diff
@@ -0,0 +1,76 @@
...
+
+            outputs.append({"output": output, "features": features})
+
+        data["resultobj"] = outputs
+        return data
+
+    @property
+    def as_dict(self) -> Dict[str, Dict[str, Dict[UUID, float]]]:
+        """Presents the feature effects as a dictionary by output."""
+        return {output.output: output.feature_dict for output in self.outputs}
```
kroenlein, line 76: resource.py supports an `as_dict` method that is not a property.  Does it need to be a property?  I'm concerned about changing the signature on a derived class.

kroenlein (commented):
Mostly good.  I have a concern about a sig change.

src/citrine/informatics/feature_effects.py
kroenlein: Yes, you're right.  Not a sig collision.

Choice of property or method doesn't seem to have a consistent flavor in the library.  This feels more like a method to me than a property, but acceptable as is if you feel strongly to the contrary.

PR #983, opened 2025-01-08
[PNE-6511] Expose feature effects generation.

anoto-moniz:
At present, the backend will automatically generate feature effects upon successful training. However, that means old models lack them. Additionally, there may be circumstances where regenerating the values may be desired.

The exposed endpoint will begin an asynchronous job which will generate new Shapley values, and replace the old ones.

### PR Type:


src/citrine/resources/predictor.py
```diff
@@ -196,6 +196,14 @@ def rename(self,
         entity = self.session.put_resource(path, json, version=self._api_version)
         return self.build(entity)
 
+    def generate_feature_effects(self,
+                                 uid: Union[UUID, str],
+                                 *,
+                                 version: Union[int, str] = MOST_RECENT_VER) -> GraphPredictor:
+        path = self._construct_path(uid, version, "shapley/generate")
+        self.session.put_resource(path, {}, version=self._api_version)
+        return self.get(uid, version=version)
```
anoto-moniz, line 205: The `FeatureEffects` object returned by the `generate` call will always be in progress and have no results yet. Additionally, as currently constructed, it doesn't provide an easy way to get the updated status: you'd have to retrieve the corresponding predictor, then call `.feature_effects` on it. So instead, this does the get automatically.

kroenlein (approved):
Something seems aesthetically weird here, but all lines up and presumably passes tests.  Will you be adding a follow-on PR to devkit for e2es?

PR #986, opened 2025-01-17
Speed up DS unit test.

anoto-moniz:
Have requests in `test_design_space_limits` return smaller responses, as most of the time was spent building them into objects.

### PR Type:


kroenlein (approved):
This dramatically reduces the entropy in this test, but I was always skeptical that we needed that much.  

Pulling in numpy would also speed this up (https://numpy.org/doc/stable/reference/random/generated/numpy.random.rand.html), but probably doesn't matter.

PR #989, opened 2025-03-25
PNE-6482 Support `default_labels` in MaterialRun constructor

jamie-heller:
# Citrine Python PR
[PNE-6482](https://citrine.atlassian.net/browse/PNE-6482)

## Support `default_labels` in MaterialRun constructor 
This PR adds support for a new optional `default_labels: List[str]` parameter in the MaterialRun constructor.

Default labels are used to:
  - Populate labels on the ingredient run, if none are explicitly specified, when the material run is later used as an ingredient
  - Marking the material run as a potential replacement ingredient for a particular label when generating new candidates using a design space. Note that during design, default labels are only applicable if the material run has no associated ingredient run within the training set in question.

### PR Type:

[PNE-6482]: https://citrine.atlassian.net/browse/PNE-6482?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kroenlein:
Why is this being update here and not in https://github.com/CitrineInformatics/gemd-python/blob/main/gemd/entity/template/material_template.py?  Nominally, that's where the data structure is inherited from.

jamie-heller:
> Why is this being update here and not in https://github.com/CitrineInformatics/gemd-python/blob/main/gemd/entity/template/material_template.py? Nominally, that's where the data structure is inherited from.

This seemed to be the most directly equivalent to the logic that was added in https://github.com/CitrineInformatics/platform-backend/pull/5021 to support the design-over-untested-ingredients functionality via UI ingest -- adding the generated tags to the material run.

Since we don't add tags to the material template in that workflow I think it probably makes sense to limit the scope of this change as well, but I'm open to reconsidering.

kroenlein (dismissed):
This is fit for purpose.  Some stylistic feedback to be provided off-line.

kroenlein (approved):
LGTM. Thanks for the additional tests.

PR #995, opened 2025-06-17
[PNE-7018] Deprecate PEWs.

anoto-moniz:
[TODO: Primarily, update the PEW and PEW exec tests to account for deprecation, and add tests for predictor evaluators, which will largely be copies of the PEW exec tests. Also should confirm the deprecated messages appear as expected.]

PEWs no longer need to be a customer facing asset type. Their only function is to allow reusing the same set of evaluators with different predictors, which A) doesn't often work due to mismatched responses, and B) isn't particularly helpful. As such, we can move users on to a cleaner, easier to use workflow, where predictor evaluation is kicked off directly from a predictor, and you can start one using the default evaluators with a single call.

In addition to exposing additional endpoints and adding evaluators to the PEW executions, this also deprecates the
predictor_evaluation_workflow property, and other Predictor Evaluation Workflow related assets. It also deprecates properties and classes named something akin to "predictor evaluation execution", in favor of the simpler and clearer "predictor evaluation".

### PR Type:


anoto-moniz:
Changing branch name.

PR #997, opened 2025-07-16
[PNE-7018] Add new evaluations API and deprecate PEWs.

anoto-moniz:
PEWs no longer need to be a customer facing asset type. Their only function is to allow reusing the same set of evaluators with different predictors, which A) doesn't often work due to mismatched responses, and B) isn't particularly helpful. As such, we can move users on to a cleaner, easier to use workflow, where predictor evaluation is kicked off directly against predictor, and you can start one using the default evaluators with a single call.

The PredictorEvaluation API and associated objects are similar to the existing PEW and PEW execution APIs/objects, but are stand-alone. This ensures we can easily make a clean break when the deprecation cycle completes, and also provides an opportunity to ensure the interfaces are  as clean and simple as possible.

This also attaches a deprecation warning to the PEW and PEW Execution collections and their associated operations. The associated response objects (`PredictorEvaluationWorkflow` and `PredictorEvaluationExecution`) are also deprecated, but as they're only produced by the collections, their docstring notes their deprecation rather than an emitted message, to avoid inundating the user.

### PR Type:

### DIFF
--- a/src/citrine/informatics/executions/predictor_evaluation_execution.py
+++ b/src/citrine/informatics/executions/predictor_evaluation_execution.py
@@ -8,7 +8,7 @@ from citrine._rest.resource import Resource
 
 
 class PredictorEvaluationExecution(Resource['PredictorEvaluationExecution'], Execution):
-    """The execution of a PredictorEvaluationWorkflow.
+    """[DEPRECATED] The execution of a PredictorEvaluationWorkflow.
 
     Possible statuses are INPROGRESS, SUCCEEDED, and FAILED.
     Predictor evaluation executions also have a ``status_description`` field with more information.
--- a/src/citrine/informatics/workflows/predictor_evaluation_workflow.py
+++ b/src/citrine/informatics/workflows/predictor_evaluation_workflow.py
@@ -12,7 +12,7 @@ __all__ = ['PredictorEvaluationWorkflow']
 
 class PredictorEvaluationWorkflow(Resource['PredictorEvaluationWorkflow'],
                                   Workflow, AIResourceMetadata):
-    """A workflow that evaluations a predictor.
+    """[DEPRECATED] A workflow that evaluates a predictor.
 
     Parameters
     ----------
--- a/src/citrine/resources/predictor_evaluation_execution.py
+++ b/src/citrine/resources/predictor_evaluation_execution.py
@@ -1,4 +1,5 @@
 """Resources that represent both individual and collections of predictor evaluation executions."""
+from deprecation import deprecated
 from functools import partial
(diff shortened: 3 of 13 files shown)


docs/source/workflows/predictor_evaluation_workflows.rst
anoto-moniz, line 1: I debated whether to change this filename to `predictor_evaluations.rst`, but left it for now to avoid any broken links coming into it.

src/citrine/informatics/executions/predictor_evaluation_execution.py
```diff
@@ -8,7 +8,7 @@
 
 
 class PredictorEvaluationExecution(Resource['PredictorEvaluationExecution'], Execution):
-    """The execution of a PredictorEvaluationWorkflow.
+    """[DEPRECATED] The execution of a PredictorEvaluationWorkflow.
```
anoto-moniz, line 11: Since the `PredictorEvaluationExecutionCollection` is already going to emit a deprecation warning on any operation, and it's the only way a `PredictorEvaluationExecution` is created, it feels like overkill to decorate this class with `@deprecated` as well. Hence the docstring update without the decorator.

src/citrine/resources/predictor_evaluation.py
```diff
@@ -0,0 +1,219 @@
...
+        """
+        path = self._get_path(action="default-from-config")
+        payload = config.dump()["instance"]
+        result = self.session.post_resource(path, json=payload, version=self._api_version)
+        return PredictorEvaluatorsResponse.build(result).evaluators
+
+    def default(self,
+                *,
+                predictor_id: Union[UUID, str],
+                predictor_version: Union[int, str] = LATEST_PRED_VER
```
anoto-moniz, line 121: The predictor version defaults to `latest` for consistency with `trigger`, which requires a trained predictor.

src/citrine/informatics/executions/predictor_evaluation.py
```diff
@@ -0,0 +1,98 @@
...
+    """:List[StatusDetail]: a list of structured status info, containing the message and level"""
+
+    def _path(self):
+        return format_escaped_url(
+            '/projects/{project_id}/predictor-evaluations/{evaluation_id}',
+            project_id=str(self.project_id),
+            evaluation_id=str(self.uid)
+        )
+
+    @lru_cache()
```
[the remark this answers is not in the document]
anoto-moniz: Fair question, but I'm not sure. This is just copied from the `PredictorEvaluationExecution` class, so I decided to keep it as is.

PR #1000, opened 2025-09-25
[PNE-7426] Expose Project.archived.

anoto-moniz:
Allow users to view the archive status of a project.

# Citrine Python PR

## Description 

### PR Type:


src/citrine/resources/project.py
```diff
@@ -599,9 +602,57 @@ def register(self, name: str, *, description: Optional[str] = None) -> Project:
         project = Project(name, description=description)
         return super().register(project)
 
+    def _list_base(self, *, per_page: int = 1000, archived: Optional[bool] = None):
+        filters = {}
+        if archived is not None:
+            filters["archived"] = str(archived).lower()
```
anoto-moniz, line 608: Both `requests` and our testing library just call `str` on a `bool`, leaving it capitalized. And it seems the accounts service doesn't recognize that as a valid boolean.

For the record, I confirmed that at least Orion paths can handle either. So I'm slightly inclined to add a snippet in `Session` that converts any `bool` query parameters to lower-cased strings automatically. But I've left that for the future, when we can more thoroughly ensure there are no ill effects elsewhere in the system.

src/citrine/resources/project.py
```diff
@@ -599,9 +602,57 @@ def register(self, name: str, *, description: Optional[str] = None) -> Project:
...
+
+        fetcher = partial(self._fetch_page, additional_params=filters, version=self._api_version)
+        return self._paginator.paginate(page_fetcher=fetcher,
+                                        collection_builder=self._build_collection_elements,
+                                        per_page=per_page)
+
     def list(self, *, per_page: int = 1000) -> Iterator[Project]:
```
anoto-moniz, line 615: Note that `ProjectCollection.list` defaults to listing every project. This differs from many other collections, such as predictors an design spaces, where `list` only lists the active assets. This will be addressed in 4.0 by renaming `PredictorCollection.list` to `list_active` and `PredictorCollection.list_all` to `list`.

kroenlein (approved):
Minor optional quibble.

PR #1033, opened 2026-03-25
Add exception chaining and fix silent pagination error

gjmulhol:
## Summary
- Add `from e` to exception re-raises in `collection.py` and `ingestion.py` so the original cause is preserved in tracebacks
- Replace `raise e` with bare `raise` to preserve traceback context
- Add `logger.warning` in `pageable.py` when response data is not a dict (was silently swallowed)

## Test plan
- [x] All 1342 existing tests pass
- [x] Exception chaining only affects traceback display, not catch behavior

🤖 Generated with [Claude Code](https://claude.com/claude-code)

*PR 7 of 11. Independent — no dependencies on other PRs.*


gjmulhol:
Consolidated into docs/improve-sdk-documentation branch.

